PR #3, opened 2020-05-18
Fix GitHub Actions

mattwthompson:
Fixing GitHub Actions since the checks weren't running in the other PR #1


mattwthompson:
In summary,

* Dropped Windows
* Added Python 3.8
* Other fixes (that in the future may exist upstream in the cookiecutter)

PR #7, opened 2020-05-29
Prototype A

mattwthompson:
## Description
A prototype of building internal topology and forcefield objects loosely off of the toolkit's. At some point this should diverge into either clearly using new objects or not. The user-level API is loosely structured in `scratch/example.py`; although clunky the general structure is there.

## Todos
Notable points that this PR has either accomplished or will accomplish.
  - [x] Provide a minimal proof-of-concept prototype of some API points and features
  - [ ] Add tests

## Status
- [ ] Ready to go


mattwthompson:
This approach involves distinct `ForceField` and `Topology` objects that largely duplicate existing functionality in the toolkit. With this approach, the primary call will be a constructor from toolkit objects that looks something like this

```
System(topology=openff_topology, forcefield=openff_forcefield)

PR #18, opened 2020-06-19
Update GitHub Actions script with miniconda and optional JAX install

mattwthompson:


mattwthompson:
Checked the logs and JAX is properly not installed and the relevant test is being skipped with Python 3.6.

PR #21, opened 2020-06-29
Namespace migration

mattwthompson:
`system` -> `openff/system`

`from system import ...` -> `from openff.system import ...`

Copying https://github.com/openforcefield/openff-evaluator/pull/241


mattwthompson:
It ... worked? On the first try!?

PR #30, opened 2020-10-15
Component refactor

mattwthompson:
## Description
Starting to move towards a simpler structure with less duplicated data and more clear separation of responsibilities

## Todos
Notable points that this PR has either accomplished or will accomplish.
  - [ ] Refactor base classes to adhered to the updated spec (draft)
  - [ ] Define SMIRNOFF subclasses

## Questions
- [ ] Should "slots" be explicitly stored? They can be accessed by calling something like `[*parsley['ImproperTorsions'].find_matches(top)]`
- [ ] Should we come up with a better term than "slots?" It may conflict with the magic `__slots__` that some Pydantic features interact with.

## Status
- [ ] Ready to go


mattwthompson:
Here is a minimal, yet hopefully instructive, example of an implementation of the handler class that uses these changes and roughly adheres to the current version of the spec:

https://gist.github.com/mattwthompson/c603d4895d9eafcc37fe7ab8afcb64d8

mattwthompson:
Not perfect or complete, but I'm merging anyway to avoid this getting out of hand (already 20 commits)


It's unclear why codecov is unhappy. Maybe that's what happens when you gut a project in one PR.

PR #45, opened 2020-11-11
Update ParmEd converter for new components

mattwthompson:
### Description
This PR updates the old ParmEd converter for new components, including building out more core attributes of the `System` class.

### DIFF
--- a/.github/workflows/integration_tests.yaml
+++ b/.github/workflows/integration_tests.yaml
@@ -25,7 +25,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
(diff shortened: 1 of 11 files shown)


mattwthompson:
Working on a system with bonds, angles, and dihedrals, currently running into https://github.com/ParmEd/ParmEd/issues/1123, which is either a bug or a bad implementation on my part. De-duplicating the bond types might be easiest solution. I'd expect angles to work similarly. Dihedrals likely require properly dealing with the 1-4 exceptions, which ParmEd treats a little differently.

PR #54, opened 2020-11-20
Include integration tests in codecov

mattwthompson:
Not sure if this will work


mattwthompson:
That definitely didn't work. Maybe I shouldn't have minimal interoperability tests separated out, and only have a separate integration testing suite for really heavy stuff.

PR #59, opened 2020-11-23
Add basic ParmEd tests

mattwthompson:
### Description
This PR adds in some non-exhaustive tests  of the ParmEd conversion module. It should at least increase the test coverage.


mattwthompson:
Torsions aren't hit, which is expected.

PR #69, opened 2020-12-10
Add to_openmm

mattwthompson:
### Description
This PR adds `System.to_openmm` for exporting to `openmm.System`.

I am starting with an approach that does not try to treat each handler similarly and instead hardcodes the addition of each force. This is because, while our view of an internal representation of a parametrized system may be elegantly siloed into separate parameters, external libraries like OpenMM will never structure data in precisely the same way. The hope is that separating out the processing at export from the internal representation will allow for the complexity of this interface to exist, but only exist in one place, a place that is safely controlled and mostly quarantined from the rest of the codebase.

## Todos
 - [x] Add proper torsion forces
 - [ ] Add impproper torsion forces
 - [x] Add nonbonded forces
- [ ] Tests
  - [x] Compare system energies, in sum and in parts, against `ForceField.to_openmm_system`
  - [ ] Various cases of vacuum and periodic systems

### DIFF
--- a/openff/system/components/smirnoff.py
+++ b/openff/system/components/smirnoff.py
@@ -196,6 +196,42 @@ class SMIRNOFFProperTorsionHandler(PotentialHandler):
                 self.potentials[identifier] = potential
 
 
+class SMIRNOFFImproperTorsionHandler(PotentialHandler):
+
+    name: str = "ImproperTorsions"
+    expression: str = "k*(1+cos(periodicity*theta-phase))"
+    independent_variables: Set[str] = {"theta"}
+    idivf: float = 1.0
+    slot_map: Dict[str, str] = dict()
+    potentials: Dict[str, Potential] = dict()
+
+    @validator("idivf")
+    def validate_idivf(cls, val):
+        if val != 1.0:
+            return UnsupportedParameterError
+
+    def store_matches(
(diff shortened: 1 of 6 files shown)


mattwthompson:
Some notes on the SMIRNOFF support for different types of nonbonded methods, and how they're implemented in OpenMM: https://github.com/openforcefield/openforcefield/issues/191#issuecomment-476842873

mattwthompson:
I have the energies _close_ for the simplest molecules that Parsley will type, but I have to admit that the energies differ by more than I expected and more than I am comfortable with. With more reasonable tolerances:

```
...
        np.testing.assert_allclose(
            toolkit_energy / unit.kilojoule_per_mole,
            system_energy / unit.kilojoule_per_mole,
            rtol=1e-6,
>           atol=1e-4,
        )
E       AssertionError:
E       Not equal to tolerance rtol=1e-06, atol=0.0001
E
E       Mismatched elements: 1 / 1 (100%)
E       Max absolute difference: 0.00048165
E       Max relative difference: 4.89489862e-05
E        x: array(9.839369)
E        y: array(9.83985)

openff/system/tests/test_interop/test_openmm.py:40: AssertionError
```

mattwthompson:
I'll handle other cases (like different combinations of periodicity, cutoffs, cutoff methods, etc.) in a separate PR

PR #70, opened 2020-12-11
Test on Python 3.8 and 3.9

mattwthompson:
### Description
This PR attempts to add Python 3.8 and 3.9 to CI, now that OpenMM's migration to `conda-forge` has  unblocked much of our dependencies.


mattwthompson:
Did not expect that to work - let's hope I didn't miss any false positives

PR #72, opened 2020-12-17
Update OpenMM tests and nonbonded methods

mattwthompson:
### Description
This PR updates the OpenMM export tests and also fixes setting the nonbonded methods in the `NonbondedForce` when the vdW and Electrostatics cutoff methods in the SMIRNOFF force field are set to `Cutoff` and `PME`, respectively (the default/common case).

When working on #69 I was frustrated by how difficult OpenMM made comparing energies - turns out I was simply disregarding existing functionality in the OpenFF Toolkit. Thanks @j-wags

### DIFF
--- a/openff/system/tests/test_interop/test_openmm.py
+++ b/openff/system/tests/test_interop/test_openmm.py
@@ -23,6 +23,8 @@ def test_from_openmm_single_mols(periodic, mol, n_mols):
     mol = Molecule.from_smiles(mol)
     mol.generate_conformers(n_conformers=1)
     top = Topology.from_molecules(n_mols * [mol])
+    mol.conformers[0] -= np.min(mol.conformers) * unit.angstrom
+
     if periodic:
         top.box_vectors = np.eye(3) * np.asarray([10, 10, 10]) * unit.nanometer
     else:
@@ -34,6 +36,7 @@ def test_from_openmm_single_mols(periodic, mol, n_mols):
         positions = np.vstack(
             [mol.conformers[0], mol.conformers[0] + 3 * unit.nanometer]
         )
+        positions = positions * unit.nanometer
 
     toolkit_system = parsley.create_openmm_system(top)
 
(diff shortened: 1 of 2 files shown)


mattwthompson:
ref https://github.com/openforcefield/openforcefield/issues/219 for more details on SMIRNOFF + OpenMM

mattwthompson:
Some of the tests are producing NaN energies, but I can't get to the bottom of it. They're the tests with two molecules and a periodic box. They pass on my machine, which is using the most recent version of everything and seems to match the versions pulled down by CI:

```
$ conda list | grep -e "^open[f,m]"
openforcefield            0.8.2              pyh39e3cac_0    omnia
openforcefields           1.3.0                      py_0    omnia
openmm                    7.5.0           py38h6cc8c4d_4_apple    conda-forge

File: openff/system/tests/test_interop/test_openmm.py [hunk 1]
```diff
@@ -21,32 +23,30 @@ def test_from_openmm_single_mols(mol, n_mols):
...
+        top.box_vectors = None
 
     if n_mols == 1:
         positions = mol.conformers[0]
     elif n_mols == 2:
         positions = np.vstack(
-            [mol.conformers[0], mol.conformers[0] + 2 * unit.nanometer]
+            [mol.conformers[0], mol.conformers[0] + 3 * unit.nanometer]
         )
+        positions = positions * unit.nanometer
```
ahy3nz, line 39:
```suggestion
        positions = positions * unit.angstrom
```

I was also getting the nan energies with `periodic=True` and `n_mols=2`. I think there's an artificial inflation of positions when multiplying by `unit.nanometer`. When adding 3 nm to the conformer positions, this correctly adds 30 angstroms to the positions, but everything is still implicitly in angstroms, even when you convert everything to a numpy array. Then, multiplying by `unit.nanometer` to re-add units will incidentally increase the positions by an order of magnitude. I think it might be more appropriate to multiply by `unit.angstrom` (and this got my tests to pass locally)

packages:
```
openforcefield            0.8.2              pyh39e3cac_0    omnia
openforcefields           1.3.0                      py_0    omnia
openmm                    7.5.0            py37hd9ee695_4    conda-forge
```

mattwthompson:
Thanks @ahy3nz for checking -  that fixed it. I thought that was in a previous commit but I must have had too many moving parts at once. Kinda spooked that it was passing locally all this time ...

PR #75, opened 2020-12-18
Add improper torsions, fix multi-term torsions in OpenMM export

mattwthompson:
### Description
This PR adds support for improper torsions in SMIRNOFF and the export to OpenMM. Some of this is copied from the corresponding code in the toolkit, but energies are not yet equal.


mattwthompson:
Tests are still failing, although it's been difficult to isolate the cause. The proper torsions seem to be incorrectly implemented in some cases (only simple cases were tested before).

mattwthompson:
Thanks @j-wags for the invaluable contributions of  debugging and isolating useful test cases!

PR #77, opened 2020-12-23
Add unit models

mattwthompson:
### Description
This PR beings to implement some models for `pint.Quantity` wrappers that play well with pydantic models.

### Todos
  - [x] `ArrayQuantity`
  - [x] Deserialization
  - [x] Remove old code
  - [ ] Enforce explicitness of units in existing models? (a.k.a. revert most of #65)

### DIFF
--- a/openff/system/tests/test_types.py
+++ b/openff/system/tests/test_types.py
@@ -43,13 +43,13 @@ class TestQuantityTypes:
 
     @pytest.mark.parametrize("val", [True, [1]])
     def test_bad_float_quantity_type(self, val):
-        class Atom(DefaultModel):
-            mass: FloatQuantity["atomic_mass_constant"]
+        class Model(DefaultModel):
+            a: FloatQuantity["atomic_mass_constant"]
 
         with pytest.raises(
             ValueError, match=r"Could not validate data of type .*[bool|list].*"
         ):
-            Atom(mass=True)
+            Model(a=val)
 
     def test_array_quantity_model(self):
         class Molecule(DefaultModel):
@@ -87,15 +87,15 @@ class TestQuantityTypes:
             except ValueError:
                 assert all(getattr(m, key) == getattr(parsed, key))
 
-    @pytest.mark.parametrize("val", [True, [1]])
+    @pytest.mark.parametrize("val", [True, 1])
     def test_bad_array_quantity_type(self, val):
-        class Atom(DefaultModel):
-            mass: ArrayQuantity["atomic_mass_constant"]
+        class Model(DefaultModel):
+            a: ArrayQuantity["atomic_mass_constant"]
 
         with pytest.raises(
-            ValueError, match=r"Could not validate data of type .*[bool|list].*"
+            ValueError, match=r"Could not validate data of type .*[bool|int].*"
         ):
-            Atom(mass=True)
+            Model(a=val)
 
     def test_float_and_quantity_type(self):
         class MixedModel(DefaultModel):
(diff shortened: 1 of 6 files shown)


openff/system/types.py
```diff
@@ -1,6 +1,146 @@
...
+                # return through converting to some intended default units (taken from the class)
+                return val.to(unit_)
+                # could return here, without converting
+                # (could be inconsistent with data model - heteregenous but compatible units)
+                # return val
+            elif isinstance(val, (float, int)):
+                return val * unit_
+            elif isinstance(val, str):
+                # could do custom deserialization here?
+                return unit.Quantity(val).to(unit_)
```
mattwthompson, line 47: Add an `elif` clause for handling some common failures, i.e. list-like

openff/system/tests/test_types.py
```diff
@@ -33,3 +36,68 @@ def test_default_units(self, default_unit):
...
+        class Molecule(DefaultModel):
+            masses: ArrayQuantity["atomic_mass_constant"]
+            charges: ArrayQuantity["elementary_charge"]
+            foo: ArrayQuantity
+            bar: ArrayQuantity["degree"]
+            baz: ArrayQuantity["second"]
+
+        m = Molecule(
+            masses=[16, 1, 1],
+            charges=[-1, 0.5, 0.5],
```
mattwthompson, line 79:
```suggestion
            charges=np.asarray([-1, 0.5, 0.5]),
```

openff/system/types.py
```diff
@@ -1,6 +1,146 @@
...
+                )
+        else:
+            unit_ = unit(unit_)
+            if isinstance(val, unit.Quantity):
+                assert unit_.dimensionality == val.dimensionality
+                return val.to(unit_)
+            elif isinstance(val, np.ndarray):
+                return val * unit_
+            elif isinstance(val, list):
+                return np.array(val) * unit_
```
mattwthompson, line 124:
```suggestion
            elif isinstance(val, (list, np.ndarray)):
                return val * unit_
```

It's possible that `pint` handles this for us, as it casts `list * Unit` into an array (before it is passed to any other methods)

openff/system/types.py
```diff
@@ -1,6 +1,146 @@
...
+                )
+            return {
+                "val": data,
+                "unit": str(obj.units),
+            }
+        return json.JSONEncoder.default(self, obj)
+
+
+def custom_quantity_encoder(v):
+    return json.dumps(v, cls=QuantityEncoder)
```
mattwthompson, line 74: Might be worth doing something to hide this so as to ensure it's not used outside of its intended scope (which is inside the model, not the entire class) - is this function necessary, or can the returned function be defined inside of the model config?

mattwthompson:
Some feedback
   * Need to set `validate_assignment = True` in order for validators to be called when mutating a model
   * Units of fields are mutable (within the constraint that changes to units match dimensionality) - unclear if this is a feature or a downside
   * Does `baz=np.array([3, 2, 1]).tobytes(),` work on all platforms? (Something about endianness of the NumPy array)
   * Test for `Atom.parse_raw(a.json()) == a`?
   * Test for `Atom(**a.dict()) == a`?
   * Main limitation of this implementation is reading primitives back in, i.e. in `Atom(**json.loads(a.json()))`/`Molecule(**json.loads(m.json()))` because those representations of data are ambiguous (without a custom deserializer)
   * Add a test case for mixing `FloatQuantity`, `ArrayQuantity`, and primitive(s)

openff/system/types.py
mattwthompson: 8fc17a1eb5c24f1b70bad85d95de878ba122128b

mattwthompson:
Merging this now, and in a separate PR will roll these classes out to the existing models

PR #89, opened 2021-01-20
See if I can run exisiting actions from a fork

mattwthompson:


mattwthompson:
Looks successful, based on how the only checks that ran in the CI workflow were Python 3.6.

PR #90, opened 2021-01-21
Add GROMACS .top file writer

mattwthompson:
### Description
This PR adds support for writing GROMACS `.top` files without the use of ParmEd or other external converters/writers. It is, however, heavily lifted from InterMol.


mattwthompson:
Running into an issue about how to test this. The ParmEd writer is not fully functional, so using it as a direct point of reference is blocked. The other energy comparisons were easy to do because the toolkit and this project can both generate reliable OpenMM objects, which can be passed through existing interoperability to GROMACS files, which InterMol can evaluate energies of. I supposed the route could still go through OpenMM, but that's not ideal. So this would be comparing OpenFF System -> GROMACS files to a route that is OpenFF System -> OpenMM System -> ParmEd -> GROMACS files.  Yuck.

mattwthompson:
Working for simple systems, merging now to prevent an unwieldy PR from forming

PR #94, opened 2021-01-28
Topology models

mattwthompson:
### Description
This is a draft implementation of an MM-focused topology, which is intended to serve as a stripped down-version of other representations.


mattwthompson:
After some more digging and though, I think I had previously under-estimated how well the existing `Topology` classes in the OpenFF Toolkit can support some of the short-term needs (granting that support for things like biopolymers and solid does not exist yet). I'll try it out in a separate PR and may let this one stagnate for a few days.

mattwthompson:
Closing in support of re-using toolkit models, which we've been doing for a few months now.

Issue #97, opened 2021-01-29
Design: De-duplicate parameterized data based on molecule isomorphism


mattwthompson:
This is already done in some topology models. It would require some overhead but potentially be a huge win for data compression while allowing for de-compression as needed (i.e. in an OpenMM conversion).

mattwthompson:
Related #12 ?

PR #98, opened 2021-01-29
Improvements to GROMACS writers

mattwthompson:
### Description
This PR is the result of experimenting in #94 and extends the existing GROMACS writers to support topologies with multiple molecules. This potentially enables #97 to be done with less refactoring.


mattwthompson:
Going to merge this now and deal with tests in a separate PR, since that is not a small task.

PR #101, opened 2021-02-04
Expand energy tests to more systems

mattwthompson:


mattwthompson:
One test is failing by a small value. Going to restart and see what happens, but this may be a hint that depending on engines to reproduce energies within a tolerance of a 0.001 k/mol may not be viable.

mattwthompson:
Another thought: the tolerances should really be something like 1e-6, closer to the reproducibility of the actual engines. Some small differences to look into are how the GRO file rounding happens and if the tail corrections match up exactly

Issue #103, opened 2021-02-08
SMIRNOFF: Resolving nonbonded methods


mattwthompson:
Tag https://github.com/openforcefield/openff-toolkit/issues/380

mattwthompson:
Another useful piece of data: https://github.com/openforcefield/openff-toolkit/issues/219#issuecomment-478161240

mattwthompson:
I don't think that GROMACS has a way to do no-cutoff electrostatics like OpenMM's `NoCutoff`, and I think that setting a large cutoff would not work without making `rlist` large, which is bad

https://manual.gromacs.org/documentation/2020/user-guide/mdp-options.html#mdp-value-coulombtype=Cut-off

mattwthompson:
Closing in favor of more targeted issues which already exist elsewhere

PR #105, opened 2021-02-08
Add nonbonded method compatibility checks, fix energy tests

mattwthompson:
### Description
This PR adds a small compatibility layer that can be built out once there is some progress on i.e. #103

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/system/components/system.py
+++ b/openff/system/components/system.py
@@ -69,3 +69,20 @@ class System(DefaultModel):
     def to_parmed(self):
         """Export this sytem to a ParmEd Structure"""
         return to_parmed(self)
+
+    def _get_nonbonded_methods(self):
+        nonbonded_ = {
+            "electrostatics_method": self.handlers["Electrostatics"].method,
+            "vdw_method": self.handlers["vdW"].method,
+            "periodic_topology": self.box is not None,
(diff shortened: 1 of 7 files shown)


mattwthompson:
This works as a first pass, but with a little bit more hardcoding that I'd prefer. I'm not sure how feasible it would be to create an elegant way to do this generally; by necessity, a lot of exports are going to require one-off solutions that conflict with the idea of an extremely general compatibility layer.

mattwthompson:
I've spent several hours trying to get to the bottom of the differences in non-bonded energies. I've been able to verify that the partial charges and LJ parameters are identical, but still not able to get energies remotely close. This is with or without long-range electrostatics, using the same cutoff for both systems, and producing errors that are much more than i.e. known PME errors. This is with either methanes or ethanols (i.e. can rule out 1-4 being the lone issue).

mattwthompson:
Thanks largely to @j-wags being diligent enough to check for atom re-ordering when I was casually using SMILES as an interchange format (🥶 ), nonbonded energies are now matching to roughly the fourth significant figure (instead of being totally garbage, as before)

PR #108, opened 2021-02-11
Add support for library charges

mattwthompson:
### Description
This started as an attempt to use a water dimer as a constrained test cases for energy comparisons, but doing that requires (at least) adding support for library charges. The constrained angle might also produce some complexity

### Checklist
- [x] Update writers
- [ ] Compare energies between OpenMM and GROMACS
- [x] Add tests
- [x] Lint
- [x] Update docstrings


mattwthompson:
It might be nice to have https://github.com/openforcefield/openff-toolkit/pull/833 resolved before merging this

mattwthompson:
Everything I set out to implement at the start of this PR is done. The only missing functionality is handling constraints in GROMACS, but that's more complex than I'm likely to tackle here. Most constraints are specified in the `.mdp` file, but some (notable water) are or can be specified in the `.top` file, so handling the different cases is tricky.

PR #114, opened 2021-02-17
Foyer Handlers

umesh-timalsina:
### Description
In #71, @mattwthompson provided an outline processing data from `foyer`. This PR presents a draft implementation for creating PotentialHandlers for data from `foyer`.

Co Authors:
1. @daico007
2. @justinGilmer

Here are couple of issues:
1. Currently, we are taking in a `parmed` structure, `atomtyping` it with foyer, and creating a system object. However, one question is on the creation of a `openff.toolkit.Topology` object out of the structure. This seems trivial but I am not sure if this is in the scope of this PR.
2. For a `FoyerBondHandler` we need the `slots` that were created using a `FoyerAtomTypes`, which seems logical but I am not sure if it's intended to be used in such fashion
### Checklist
- [ ] Add tests
- [x] Lint
- [ ] Update docstrings


openff/system/components/foyer.py
```diff
@@ -0,0 +1,152 @@
...
+        structure: pmd.Structure,
+    ) -> None:
+        """
+        Populate self.slot_map with key-val pairs of slots
+        and unique potential identifiers
+
+        """
+        type_map = forcefield.run_atomtyping(structure)
+        for key, val in type_map.items():
+            self.slot_map[key] = val["atomtype"]
```
mattwthompson, line 66: I like the simplicity here.

openff/system/components/foyer.py
```diff
@@ -0,0 +1,152 @@
...
+    expression: str = "4*epsilon*((sigma/r)**12-(sigma/r)**6)"
+    independent_variables: Set[str] = {"r"}
+    slot_map: Dict[int, str] = dict()  # type: ignore
+    potentials: Dict[str, Potential] = dict()
+
+    def store_matches(
+        self,
+        forcefield: Forcefield,
+        structure: pmd.Structure,
+    ) -> None:
```
mattwthompson, line 58: Would it be possible to modify this function in a way that bypasses a reliance on ParmEd? I know it is currently used as an intermediary data store, but I think generating matches directly from an `mb.Compound` would be more consistent with the scope of this function. It may also be worth exploring a larger refactor that allows Foyer to operate directly on compounds with less reliance on ParmEd and OpenMM.

```suggestion
    def store_matches(
        self,
        forcefield: Forcefield,
        compound: mb.Compound,
    ) -> None:
```

openff/system/components/foyer.py
```diff
@@ -0,0 +1,152 @@
...
+            self.slot_map[key] = val["atomtype"]
+
+    def store_potentials(self, forcefield) -> None:
+        from simtk.openmm.app.forcefield import NonbondedGenerator  # type: ignore
+
+        non_bonded_forces_gen = _get_openmm_force_gen(forcefield, NonbondedGenerator)
+
+        if non_bonded_forces_gen:
+            nonbonded_params = non_bonded_forces_gen.params.paramsForType
+
```
mattwthompson, line 75: I think the amount of needed here points to how significantly Foyer would benefit from more de-coupling from OpenMM. Going through OpenMM force objects works but should not be necessary - they are hugely complex data structures targeted at performance in simulations. Ideally it should be able to do stuff like

```python3
import foyer

oplsaa = foyer.Forcefield(name="oplsaa")
oplsaa.get_parameters(group="Atoms", key="opls_151")  # arguments should be named differently
# prints something containing values of sigma/espilon/q
oplsaa.get_parameters(group="Bonds", key=["CT", "CT"])
# prints something containing k, length
...
```

openff/system/components/foyer.py
```diff
@@ -0,0 +1,152 @@
...
+    name: str = "Bonds"
+    expression: str = "1/2 * k * (r - length) ** 2"
+    independent_variables: Set[str] = {"r"}
+    slot_map: Dict[Tuple[int, int], Tuple[str, str]] = dict()  # type: ignore
+    potentials: Dict[Tuple[int, int], Potential] = dict()  # type: ignore
+
+    def store_matches(
+        self, structure: pmd.Structure, atom_slots: Dict[str, str]
+    ) -> None:
```
mattwthompson, line 106: Since Foyer types bonds (and other sections) using the same typing algorithm as the non-bonded section, it would be better to use those results here and **not** look into a ParmEd structure 

```suggestion
    def store_matches(
        self, atom_types: FoyerAtomTypes,
    ) -> None:
```

mattwthompson:
Apologies for some of the confusion with making the linters happy. I need to add some developer docs that explain how to automate that.

Quick thoughts on your questions:
1. **We should avoid using ParmEd as an intermediary.** As long as Foyer is intended to take in mBuild compounds, they should be used as the source of topological data. Converting among these objects is not trivial; maybe simple for simple cases, but carefully outlining conversions between rich objects, and the unsupported behavior, is tricky.
2. "Slots," which deserve a more descriptive name, should be generated from the topology itself. So, bond "slots" are tuples of atom indices generated by inspecting the graph of the topology.

openff/system/components/foyer.py
umesh-timalsina: This can be achieved with https://github.com/mosdef-hub/foyer/pull/358 in the long run, All in all I agree, this feature is not to be implemented here. However, I will be happy to create an issue in foyer and resolve it there.

openff/system/components/foyer.py
umesh-timalsina: I was under the impression that the `run_atomtyping` step doesn't parameterize the structure. The tricky thing here is to get the types for underlying bonds in the structure/ compound. Am I missing something?

openff/system/tests/test_foyer.py
```diff
@@ -0,0 +1,23 @@
...
+
+from openff.system.components.foyer import from_foyer
+from openff.system.tests.base_test import BaseTest
+
+
+class TestFoyer(BaseTest):
+    @pytest.fixture(scope="session")
+    def oplsaa_system_ethane(self):
+        import foyer  # type: ignore
+        from foyer.tests.utils import get_fn  # type: ignore
```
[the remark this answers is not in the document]
umesh-timalsina: Will do

openff/system/components/foyer.py
umesh-timalsina: With https://github.com/mosdef-hub/foyer/pull/358, I think we can either take in an untyped gmso `Topology` if so desired. This step needs some extension to support a `mBuild` compound IMO. https://github.com/mosdef-hub/foyer/blob/82ebfca79044af9f2fb449b91941fdc8071e6c5d/foyer/forcefield.py#L583

openff/system/components/foyer.py
mattwthompson: I had to look this up - I thought that `foyer.Forcefield.run_atomtyping` took in an un-parametrized ParmEd structure  and returned the same object with atom types stored as string on each atom. But I was wrong - iit does give something more compact and useful as an output:


```
foyer.Forcefield(name='oplsaa').run_atomtyping(compound.to_parmed())
/Users/mwt/miniconda3/envs/openff-system/lib/python3.8/site-packages/foyer/forcefield.py:448: UserWarning: No force field version number found in force field XML file.
  warnings.warn(
/Users/mwt/miniconda3/envs/openff-system/lib/python3.8/site-packages/foyer/forcefield.py:460: UserWarning: No force field name found in force field XML file.
  warnings.warn(
/Users/mwt/miniconda3/envs/openff-system/lib/python3.8/site-packages/foyer/validator.py:132: ValidationWarning: You have empty smart definition(s)
  warn("You have empty smart definition(s)", ValidationWarning)
Out[13]:
{0: {'whitelist': set(), 'blacklist': set(), 'atomtype': 'opls_135'},
 1: {'whitelist': set(), 'blacklist': set(), 'atomtype': 'opls_157'},
 2: {'whitelist': set(), 'blacklist': set(), 'atomtype': 'opls_154'},
 3: {'whitelist': set(), 'blacklist': set(), 'atomtype': 'opls_140'},
 4: {'whitelist': set(), 'blacklist': set(), 'atomtype': 'opls_140'},
 5: {'whitelist': set(), 'blacklist': set(), 'atomtype': 'opls_140'},
 6: {'whitelist': set(), 'blacklist': set(), 'atomtype': 'opls_140'},
 7: {'whitelist': set(), 'blacklist': set(), 'atomtype': 'opls_140'},
 8: {'whitelist': set(), 'blacklist': set(), 'atomtype': 'opls_155'}}
```

It would still, in my opinion, be useful to have that accept a compound as an argument, more closely matching one of the intended pathways for data to flow through foyer (`Forcefield.apply(compound)`). The data that Foyer actually needs from the un-parametrized ParmEd structure is fairly minimal (elements, bonds, residues, maybe formal charge) compared to all of the data that can be shoved into a structure. An OpenFF Topology object could serve this purpose well - probably out of the box but possibly with a thin subclass.

umesh-timalsina:
Tests will fail because of https://github.com/mosdef-hub/foyer/pull/382, which should be merged soon. But, the handlers implementated here should be a rough draft of what we want.

mattwthompson:
Feel free to `pip` install against a non-release version of Foyer (i.e. [here](https://github.com/openforcefield/openff-system/blob/adfb86ae42d1884476acc3aae89798e86c3c99aa/.github/workflows/ci.yaml#L40-L44)). You can also use the `git+` syntax to install against a non-default branch (i.e. feature branch) if that makes development easier.

mattwthompson:
Partially documenting the expected keys in `Potential.parameters` for various handlers:

Bonds: `k`, `length`
Angles: `k`, `angle`
Torsions: `k`, `phase`, `periodicity`
vdW: `sigma`, `epsilon`
Electrostatics: not stored as a `Potential`, just stored as a unit-tagged float

umesh-timalsina:
There are couple of points to be addressed here:
1. Two types of torisions are currently extracted from foyer. RB and Periodic. Which means, there will be two handlers for them. For Periodic Torsions, `k`, `phase` and `periodicity` have been used but for `RBTorsions`. `c0, .... c5` have been used for their parameters.  
2. For angle params, I have used `theta` as angle. Which should be changed.

umesh-timalsina:
Also, after https://github.com/mosdef-hub/foyer/pull/389. The system creation should consider taking an `OpenFF` Topology instead of parmed structure.

mattwthompson:
> `RBTorsions`

The best way to handle this is probably to have a separate `RBTorsionsHandler` that is separate from the `ProperTorsionsHandler` objects that already exist, similar to how OpenMM has separate forces for each. I can handle this in a separate PR as it's a useful feature outside the scope of Foyer interoperability.

umesh-timalsina:
https://github.com/openforcefield/openff-system/blob/b5ecee428ef55402f5ab3c37ab3189c5b2a8d1b8/openff/system/components/foyer.py#L189-L301.

I have added support for them but instead of two, I created four handlers:
1. `FoyerPeriodicProperHandler`
2. `FoyerPeriodicImproperHandler`
3. `FoyerRBProperHandler`
4. `FoyerRBImproperHandler`
This is done specifically because the topology keys differ in their order for dihedral/impropers, which I figured should be handled separately. The duplicate interactions are handled in foyer itself.

mattwthompson:
Excellent!

What's missing at the moment is that exporters currently don't support RB-style torsions at all, so there's nothing for `FoyerRBProperHandler` to hook into, so we can't run energy tests. This is a bigger task so it may not be appropriate to handle it here.

devtools/conda-envs/test_env.yaml
```diff
@@ -24,4 +24,5 @@ dependencies:
   - gromacs =2019
   - lammps
   - mbuild
+  - parmed
   - foyer
```
mattwthompson, line 28:
```suggestion
  - foyer
  - foyer >=0.8.0
```

So as to avoid pulling down an older version without the new API points

openff/system/components/foyer.py
```diff
@@ -0,0 +1,248 @@
...
+from foyer.topology_graph import TopologyGraph
+from openff.toolkit.topology import Topology
+
+from openff.system import unit as u
+from openff.system.components.potentials import Potential, PotentialHandler
+from openff.system.components.system import System
+from openff.system.models import PotentialKey, TopologyKey
+
+# Is this the safest way to achieve PotentialKey id separation?
+POTENTIAL_KEY_SEPARATOR = "-"
```
mattwthompson, line 17: Perhaps `PotentialKey.id` could be `Union[str, Tuple[str]]` in order to cover this use case without smooshing strings together

devtools/conda-envs/test_env.yaml
umesh-timalsina: Its available [here](https://anaconda.org/conda-forge/foyer). But, I cannot pull it locally, neither is CI pulling it.

mattwthompson: Yeah, it's unavailable to `conda` for an hour or two after it shows up on the website. Something about CDN caching that I don't understand - I just know to wait for a bit

umesh-timalsina: Okay. Will try again in a bit.

umesh-timalsina:
@mattwthompson can we document expected `Parameter` Keys, the expression for the `Potential` from each of the `OpenMM` forces and their units, I having a hard time to find default units, specially for  derived quantities which some of the parameters seem to be.

mattwthompson:
I'm looking for ways to get the following information out of `foyer.Forcefield`
* 1-4 scaling factor for LJ terms
* 1-4 scaling factor for electrostatics
* Mixing rule/combination rule

mattwthompson:
@umesh-timalsina @daico007 @justingilmer how do you feel about merging this in its current state and then expanding the tests in a follow-up?

umesh-timalsina:
@mattwthompson. Apart from the `@classmethod` that converts a OFF-Topology into a foyer topology-graph, I don't see any issue.

mattwthompson:
Happy to remove that code from here if it's in the next Foyer release

umesh-timalsina:
> I'm looking for ways to get the following information out of `foyer.Forcefield`
> 
> * 1-4 scaling factor for LJ terms
> * 1-4 scaling factor for electrostatics
> * Mixing rule/combination rule

@mattwthompson see https://github.com/mosdef-hub/foyer/issues/397. First class support doesn't exist currently but will be added to foyer.

mattwthompson:
Perfect! Thanks @daico007

mattwthompson (approved):
Looks great! Thanks @umesh-timalsina @justinGilmer @daico007!

Issue #107, opened 2021-02-10
Interop: Documenting non-bonded methods


mattwthompson:
Work in progress

| English | Internal | SMIRNOFF | OpenMM | GROMACS
|--|--|--|--|--|
| Electrostatics
| PME | `off_sys.handlers["Electrostatics"].method == "PME"` | `<Electrostatics ... method="PME"></Electrostatics> ` | `openmm.NonbondedForce.PME` | `coulombtype              = PME`
| Cut-off | `off_sys.handlers["Electrostatics"].method == "PME"` | `<Electrostatics ... method="Coulomb"></Electrostatics> ` | `openmm.NonbondedForce.CutoffPeriodic`, `setReactionFieldDielectric(1.0)` | `coulombtype = cut-off; epsilon-r                = 1`
| vdW

Tag #103

PR #116, opened 2021-02-18
Improve support for multi-molecule systems

mattwthompson:
### Description
This PR is intended to expand support and testing of multi-molecule systems, i.e. mixtures of organic liquids, with both OpenMM and GROAMCS.

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
Lots of strange things are happening with nonbonded energies right now ... I'm confident it's a mixture of

* Exceptions aren't being passed through GROMACS tests
* Non-bonded methods don't quite match up GROMACS <--> OpenMM  #107 
* Settings in the `openmm.NonbondedForce` are getting muddied during tests #103

mattwthompson:
Jotting down some ideas

* The float formatting in the GROMACS writers is copied from InterMol, which does 8 digits past the decimal with scientific notation. I think this is actually causing some issues with larger systems in which the rounding error is compounded several times over. Parsley has many significant digits and we should attempt to respect them. This is something to keep in mind for other writers - I think OpenMM will just take whatever number you give it.
* For some tests, it may be useful to have references energies from (slow) hand-coded evaluators. I've already done this for the case of bonds, but most other handlers will be difficult and possibly out of scope. 
* It would be useful to refactor the`get_engine_energy` helper functions to produce discrete "reports," which could then be fed into the tests. This would enable these to be more concrete object models, instead of dictionaries that are thrown around, and potentially manage several engines' data with a single model. This would also make reporting more generally useful, i.e. track and store reports instead of just seeing if they pass a test and discarding them.

mattwthompson:
Hey, valence terms are working pretty well:

image

this is for [this](https://github.com/openforcefield/openff-toolkit/blob/adde241a052b31a2ffa6356b2c2ded34694270cc/openff/toolkit/data/systems/packmol_boxes/cyclohexane_ethanol_0.4_0.6.pdb) file in the toolkit, which is 96 cyclohexanes and 143 ethanols for a total of around 3,000 atoms.

Nonbonded forces are still bad (18 kJ/mol error) but other factors are at play there

mattwthompson:
This has grown out of its original scope; if tests pass I'm going to merge even though there are some missing details.

PR #122, opened 2021-02-25
Add constraints to exporters

mattwthompson:
### Description
This PR adds constraints to exporters. They are already stored in the internal representation, I just forgot to include them in the exporters when doing the first implementation.

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
Will figure out a way to better handle constraints in GROMACS later.

PR #124, opened 2021-03-01
Partial re-parametrization

mattwthompson:
### Description
This PR add a `System.reparametrize` method that allows for partial re-parametrization, i.e. reparametrizing an existing object handler-by-handler.

### Checklist
- [ ] Case of typing not changing, just parameters?
- [x] Add tests
- [x] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/system/components/system.py
+++ b/openff/system/components/system.py
@@ -6,9 +6,14 @@ from openff.toolkit.topology.topology import Topology
 from pydantic import validator
 
 from openff.system.components.potentials import PotentialHandler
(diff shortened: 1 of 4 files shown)


mattwthompson:
It might be useful to have a similar feature that only replaces parameter values, bypassing the typing step. This would need to be off by default, since it would be hard to determine if the typing results would change without doing so.

mattwthompson:
This has gone stale and is now out of date with current development; if this is a desired feature in the future I should think it through a little more beforehand

PR #131, opened 2021-03-10
Add mBuild conversions

mattwthompson:
### Description
This PR adds experimental conversions to (and possibly later _from_) mBuild Compound objects.

Note that this code will likely live somewhere else in the future; this submodule is intended to be a temporary home. I'd rather develop here than passing around scripts and notebooks.

cc: @j-wags @umesh-timalsina @daico007 @justinGilmer please point out relevant test cases not covered by this implementation (or any other general comments)

### Checklist
- [ ] Roundtrips?
- [ ] Test that converted compounds can be passed through Foyer
- [x] Add tests
- [x] Lint
- [ ] Update docstrings


openff/system/misc.py
```diff
@@ -0,0 +1,38 @@
...
+
+    return comp
+
+
+def offtop_to_compound(off_top: Topology) -> mb.Compound:
+
+    sub_comps = []
+
+    for top_mol in off_top.topology_molecules:
+        this_comp = offmol_to_compound(top_mol.reference_molecule)
```
j-wags, line 34: This is a tiny bit risky because the "reference molecule" may have a different atom ordering than the "topology molecule" (this allows an OFF Topology to have multiple copies of the same molecule, with different atom orderings). This isn't a common situation, so I'm fine to move forward with the code in its current form for now. 
```suggestion
        # TODO: This could have unintended consequences if the TopologyMolecule has atoms in a different order than the reference Molecule
        this_comp = offmol_to_compound(top_mol.reference_molecule)
```

mattwthompson: Ah, I was unaware of that behavior. Thanks!

mattwthompson:
The failure here is probably stochastic

PR #135, opened 2021-03-15
Add Buckingham-6 support

mattwthompson:
### Description
This PR adds support for Buckingham-6 potentials in the internal representation and some exporters.

### Checklist
- [x] Internal representation
- [x] GROMACS export
- [x] OpenMM export
- [ ] LAMMPS export
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -20,7 +20,7 @@ dependencies:
   - mypy
   - unyt
   - intermol
-  - gromacs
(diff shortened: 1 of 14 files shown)


mattwthompson:
Tests are funky because of packaging issues (I can only pull down 2019 locally on macOS, and the Linux builds grab totally different versions) and some changes to GROMACS between 2019, 2020, and 2021. Will tidy those up in #136

PR #136, opened 2021-03-17
Pin GROMACS to >=2021

mattwthompson:
### Description
Some builds have been pulling old versions of GROMACS (2020 and 2019.1) on different operating systems. This PR attempts to pin to a newer version.


mattwthompson:
Interesting, bioconda requires Python 3.9 for recent GROMACS builids ...

https://anaconda.org/bioconda/gromacs/files

mattwthompson:
I asked in their gitter and got a quick response and patch:

> Whoever made the recipes put the build-time python and perl dependencies in the wrong place. I'll make a PR to fix that

https://github.com/bioconda/bioconda-recipes/pull/27409

Once that gets merged, we should be able to kick the tires on this and ensure this pin. When I was testing locally, the conflict came from their GROMACS builds and nothing else.

mattwthompson:
The PR linked above only made changes to the GROMACS 2020 builds, so I submitted another PR to apply a similar patch to 2021 builds: https://github.com/bioconda/bioconda-recipes/pull/27438

PR #138, opened 2021-03-20
Update mypy configuration

mattwthompson:
### Description
There are several instances in which heavy classes are imported just for type hints; this is one way to reduce those hints bloating import times.

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
It turns out this isn't such a huge win; it might help _running_ `mypy`, but my understanding is that it does not typically impact runtime performance.

PR #146, opened 2021-04-02
Document energy tests

mattwthompson:
### Description
This PR adds some rough initial documentation for the energy tests, including some docstrings.

The important points relating to GROMACS are
1. Call `get_gromacs_energies` directly on a `System` object to get single-point energies
2. Call `_run_gmx_energy` on a collection of GROMACS files (prepared here or externally, i.e. via ParmEd) to get single-point energies. `get_gromacs_energies` calls this function, so the behavior should be similar (or able to be made similar).
3. I've hacked together a few `.mdp` files to cover some types of calculations, but these are not exhaustive and a better solution (programmatically generating them?) should replace these in the future. For now, see [this](https://github.com/openforcefield/openff-system/tree/9bca382ba64aede0be9e9138159db51a58f7a7c8/openff/system/tests/files/mdp) path.

cc: @justingilmer, @umesh-timalsina @daico007 - feedback welcome

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/docs/energy_tests.md
+++ b/docs/energy_tests.md
@@ -8,7 +8,7 @@ There are also some internal functions that can be used to compute energies from
 
 This example shows many ways of computing single-point energies from a constructed `System` object.
 
-```
+```python3
 import numpy as np
 
 # Create an OpenFF System object from an OpenFF Molecule, Topology, and ForceField
--- a/docs/energy_tests.md
+++ b/docs/energy_tests.md
@@ -1,0 +1,60 @@
+## Energy Tests
+
+This project includes infrastructure for comparing representations of parametrized systems via computing single-point energies using molecular simulation engines.
+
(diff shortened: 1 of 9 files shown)


docs/energy_tests.md
```diff
@@ -0,0 +1,60 @@
...
+
+This project includes infrastructure for comparing representations of parametrized systems via computing single-point energies using molecular simulation engines.
+
+Currently, OpenMM, GROMACS, and LAMMPS are supported engines for running energy tests.
+There are high-level functions for computing energies from constructed `System` objects.
+There are also some internal functions that can be used to compute energies from other inputs.
+
+This example shows many ways of computing single-point energies from a constructed `System` object.
+
+```
```
justinGilmer, line 11:
```suggestion
```python3
```

Small change that will provide syntax highlighting

PR #150, opened 2021-04-08
Test against ParmEd files

mattwthompson:
### Description
This PR begins the process of porting some of ParmEd's test suite.


mattwthompson:
Leaving one detail left undone because I don't want this to snowball even more #152 

This PR includes #96

PR #157, opened 2021-04-21
Improve non-bonded tests and exporters

mattwthompson:
### Description
- [x] Split `Nonbonded` section of report into vdW and electrostatics contributions
- [ ] Split `NonbondedForce` in OpenMM export into separate `CustomNonbondedForce` objects
- [x] Generate GROMACS `.mdp` files programmatically and on-the-fly
- [x] Generate `[ pairs ]` section in GROMACS `.top` files
- [ ] Attempt to tail corrections across methods
- [x] Somewhat reduce non-bonded energy differences across exports and tests

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
Documenting some discrepancies in switching functions across engines:

1. The [SMIRNOFF](https://open-forcefield-toolkit.readthedocs.io/en/0.9.1/smirnoff.html#vdw) spec allows the use of `switch_width`, which implicitly dictates that a switching function should be used. Related https://github.com/openforcefield/openff-toolkit/issues/882 https://github.com/openforcefield/openff-toolkit/issues/896
1. OpenMM allows the user to set a switching function with [`.setUseSwitchingFunction()`](http://docs.openmm.org/7.5.0/api-python/generated/simtk.openmm.openmm.CustomNonbondedForce.html), but I'm not sure the functional form of what's used. Maybe this is something I should know by heart, but the documentation does not detail it.
2. GROMACS has several options, of which I believe [`vdw-modifier = Potential-switch`](https://manual.gromacs.org/documentation/current/user-guide/mdp-options.html#mdp-value-vdw-modifier=Potential-switch) is the most accurate.
3. LAMMPS does not use the same terminology, but I've found [`pair_modify shift yes`](https://lammps.sandia.gov/doc/pair_modify.html) [`pair_style lj/switch3/coulgauss`](https://lammps.sandia.gov/doc/pair_lj_switch3_coulgauss_long.html), [`pair_style lj/smooth`], none of which seems to be exactly what I'm after.

mattwthompson:
Whoops, found the relevant bits of the OpenMM docs: http://docs.openmm.org/7.5.0/userguide/theory.html#lennard-jones-interaction

mattwthompson:
This is kinda getting out of hand and it's only going to get more unbearably large if I don't split out the `CustomNonbondedForce` changes into a separate PR.

PR #158, opened 2021-04-30
More GROMACS non-bonded fixes

mattwthompson:
I've managed to get a small system of liquid argon to be more consistently evaluated by OpenMM and GROMACS:
```
ipdb> gmx_energies['vdW'] - omm_energies['Nonbonded']
<Quantity(-0.0079359824, 'kilojoule / mole')>
```

Will dig in a little bit more to LAMMPS settings. Also need to figure out how to best make each of these settings exposed somewhere instead of forcing default behavior that I happen to prefer.


mattwthompson:
Okay, I've dug into the LAMMPS errors, and I think it's mostly because LAMMPS does not offer a simple way of applying a switching function. I've made a toy system of a pair of argon atoms interacting over distances `r` with a cutoff of 0.9 nm and a switching function applied at 0.8 nm. These are my interpretations of the current [SMIRNOFF spec](https://open-forcefield-toolkit.readthedocs.io/en/0.9.1/smirnoff.html#vdw) using Parsley values.

The OpenMM and GROMACS curves fall right on top of each other (this is what I used to do 1063ecb) but LAMMPS is obviously wrong without a switching function set:

image


One can just use `pair_style shift yes` and get a smooth cut-off by shifting everything to the cut-off value, but that only moves the errors around and is clearly more of an error, since it's changing the potential everywhere, not just near the cutoff. This probably works fine for an individual simulation but it not good for ensuring results between engines.
image

Two options I've found are 
* [`pair_style lj/gromacs`](https://lammps.sandia.gov/doc/pair_gromacs.html), which might need non-default packages to be installed
* [`pair_style lj/charmm`](https://lammps.sandia.gov/doc/pair_charmm.html), which unfortunately uses an incorrect Coulomb constant for CHARMM compatibility.

PR #160, opened 2021-05-03
Update API with outer/inner data

mattwthompson:
### Description
This PR updates the `System` class to follow an "outer/inner" data model in which the internals are hidden away in a pseudo-private inner class and an outer model has some API points for getting/setting pieces of the inner model. This is inspired by the [`Molecule`](https://github.com/SimonBoothroyd/strawman/blob/41cbee89b41b33afa49cbaec9ba28ca4ee5d729b/strawman/molecule.py#L145) class in the "strawman."

### Checklist
- [ ] Add tests
- [x] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/system/components/system.py
+++ b/openff/system/components/system.py
@@ -27,22 +27,60 @@ class System(DefaultModel):
     .. warning :: This API is experimental and subject to change.
     """
 
-    handlers: Dict[str, PotentialHandler] = dict()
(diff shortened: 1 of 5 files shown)


mattwthompson:
I expected this to be much more disruptive ... we'll see if there are unexpected side effects in the future.

PR #169, opened 2021-05-10
Make `parmed` an optional import

### LINKED ISSUE #162: Make `parmed` an optional import
## Description

Currently `parmed` is a required dependency as it is a top-level import. It would be good if it could be made a local only import and an optional dependency.

SimonBoothroyd:
### Description

This PR makes `parmed` an optional dependency / import. Fixes #162

### Checklist
- [ ] Add tests
- [X] Lint
- [ ] Update docstrings


mattwthompson (approved):
LGTM - I thought this was going to be much more work with many more files changed

PR #173, opened 2021-05-11
Update models

mattwthompson:
### Description
Refactoring/cleanup some code based on technical feedback from @SimonBoothroyd 

- [x] Use `Literal`s in Pydantic models
- [x] `PotentialHandler.name` -> `PotentialHandler.type`
- [ ] Explicitly use `Field` where possible
  - [ ] Add descriptions
  - [ ] Other things that are optional but useful?
 - [ ] Try out `sphinx-pydantic`
 - [x] Make `PotentialHandler.expression` a Literal
 - [x] Make `PotentialHandler.independent_variables` a property, not a field
 - [ ] Remove duplicated code in `openff/system/components/smirnoff.py`
   - [ ] Try out `GenericModel`?
- [ ] Write more validators
  - [ ] `PotentialHandler.parameters` (just to make sure the names of the parameters are actually in the expressions)
  - [ ] Make sure that slot_map and potentials are self-consistent
 - [ ] Fix L323 in `smirnoff.py`
 - [x] `SMIRNOFFElectrostaticsMetadataMixin.method` validator could be rewritten as a Literal
    - [ ]  Look into if the `Literal` class can do some string comparison
  - [x] Split out contents of `misc.py` into more explicitly targeted modules
- [ ] Make `SMIRNOFFPotentialHandler.from_toolkit` class methods
- [ ] Use `SMIRNOFFPotentialHandler.from_toolkit` class methods in `to_openff_system`

### DIFF
--- a/openff/system/components/potentials.py
+++ b/openff/system/components/potentials.py
@@ -9,7 +9,7 @@ from openff.system.models import DefaultModel, PotentialKey, TopologyKey
 from openff.system.types import ArrayQuantity, FloatQuantity
 
 if TYPE_CHECKING:
-    from openff.system.components.misc import OFFBioTop
+    from openff.system.components.mdtraj import OFFBioTop
 
 
 class Potential(DefaultModel):
--- a/openff/system/components/potentials.py
+++ b/openff/system/components/potentials.py
@@ -1,10 +1,10 @@
-from typing import TYPE_CHECKING, Dict, List, Set, Union
+import ast
+from typing import TYPE_CHECKING, Dict, List, Set
 
 from openff.toolkit.typing.engines.smirnoff.parameters import ParameterHandler
 from openff.utilities.utils import requires_package
-from pydantic import validator
+from pydantic import Field, validator
 
-from openff.system.exceptions import InvalidExpressionError
 from openff.system.models import DefaultModel, PotentialKey, TopologyKey
 from openff.system.types import ArrayQuantity, FloatQuantity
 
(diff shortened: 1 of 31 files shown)


openff/system/components/potentials.py
```diff
@@ -31,20 +30,29 @@ def validate_parameters(cls, v):
...
+
+    @property
+    def independent_variables(self) -> Set[str]:
+        vars_in_potentials = set([*self.potentials.values()][0].parameters.keys())
+        vars_in_expression = {
+            node.id
+            for node in ast.walk(ast.parse(self.expression))
+            if isinstance(node, ast.Name)
+        }
+        return vars_in_expression - vars_in_potentials
```
mattwthompson, line 55: This block of code (self-contained) might be of interest to GMSO developers (@justinGilmer @umesh-timalsina @daico007)

```
In [24]: import ast
    ...:
    ...:
    ...: def get_vars_in_expr(expr):
    ...:     names = {
    ...:         node.id for node in ast.walk(ast.parse(expr))
    ...:         if isinstance(node, ast.Name)
    ...:     }
    ...:     return names
    ...:

In [25]: get_vars_in_expr('m*x+b')
Out[25]: {'b', 'm', 'x'}

In [26]: get_vars_in_expr("4*epsilon*((sigma/r)**12-(sigma/r)**6)")
Out[26]: {'epsilon', 'r', 'sigma'}
```

Taken from https://stackoverflow.com/a/41038615/4248961

mattwthompson:
Not done with everything I aimed to do originally, but this is already a large set of changes. I'll do the next bits in follow-up PRs.

openff/system/components/potentials.py
umesh-timalsina: @mattwthompson Thank You for pointing this out. We currently have a problem with loading large [forcefield](https://github.com/mosdef-hub/gmso/issues/521) files. Validating potential expressions takes the longest amount of time. We have discussed some memoizing ideas. Together with the approach mentioned here and memoization, it looks like we can get performance benefits.

PR #189, opened 2021-05-18
Update for openff-utilities and openff-units releases

mattwthompson:

PR #190, opened 2021-05-18
Add parameter splitting example

mattwthompson:
### Description
This PR adds an example notebook that resulted from chatting with @wutobias yesterday. It shows how, with a little work, a parameter can be split out into two parameters and "re-applied" to the system, but without re-running atom-typing.

Depends on #189 

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
This needs a little smithing, but I'm merging so it does not get forgotten

PR #191, opened 2021-05-19
Add system from SMIRNOFF class method

SimonBoothroyd:
### Description
This PR replaces the `create_openff_system` monkeypatch stub method with a cleaner (and handy) `System.from_smirnoff` class method. 

Exposing `from_smirnoff` will make integrating with the `openff-toolkit` easier, and will ensure that all of the main 'business' logic stays in this repository.

### Checklist
- [X] Add tests
- [X] Lint
- [X] Update docstrings


mattwthompson:
What do you think the long-term API should be? (Here and in the toolkit)
1. The "toolkit readme" with some `mm` replaced with `ff`(https://github.com/openforcefield/openff-toolkit/pull/938)
2. Replace `openff_sys = force_field.create_openff_system(...)` with `openff_sys = System.from_smirnoff(...)` 
3. Choice 1, but with `ForceField.create_openff_system` internally calling `System.from_smirnoff`

It seems like this is heading toward choice 3 - similar user experience but most of the code sits here?

I don't have a strong opinion at the moment, so I'd be happy to hear if you or others have reason to prefer one over any others!

mattwthompson:
That all sounds reasonable to me

mattwthompson (approved):
LGTM!

We'll want to wait for some other moving pieces to solidify before merging, mostly #189 and updating examples/other places that might call `.create_openff_system`. I can be responsible for that work.

PR #195, opened 2021-05-20
Fix from SMIRNOFF used with base OFF topology

SimonBoothroyd:
### Description

This PR fixes creating an OpenFF system object using the `from_smirnoff` class method with a base OpenFF `Topology` object and not an `OFFBioTop`. Previously the existing topology was passed as the `mdtraj` arg, rather than the `other` kwarg.

The PR further adds a workaround to https://github.com/openforcefield/openff-toolkit/issues/946 as OpenFF topologies cannot currently be instantiated from `other`.

### Checklist
- [X] Add tests
- [X] Lint
- [X] Update docstrings

### DIFF
--- a/openff/system/components/mdtraj.py
+++ b/openff/system/components/mdtraj.py
@@ -1,3 +1,5 @@
+import copy
+
 import mdtraj as md
 from openff.toolkit.topology import Topology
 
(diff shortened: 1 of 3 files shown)


mattwthompson (approved):
Thanks for fixing this, it was a known pain point of the current `OFFBioTop` hack that I always worked around manually. Sorry it was not documented!

PR #200, opened 2021-05-24
Group electrostatic potential handlers

SimonBoothroyd:
### Description

This PR aims to group all of the available SMIRNOFF electrostatic parameter handlers under a single potential handler.

This proposal is driven by:

1) The systems' potential handlers have a 'slot' based view of the world, whereby each potential handler maps to a specific valence type or pairwise interaction. The bond potential handler handles terms applied to valence bond terms, the angle handler to valence angles and so on. 

   The current implementation however introduces handlers which don't necessarily map to corresponding pairwise interaction types, namely library charge slots, charge increment slots and (in a sense) am1bcc slots. What these different handlers really do is provide parameters which occupy specific **electrostatic** / **charge** slots.

   Hence it would seem to make sense to have a single **potential** handler that describes which parameter have been applied to which electrostatic slots, even if those parameters came from multiple **parameter** handlers.

2) In the current implementation validating which slots have been filled and by what is challenging and leads to more complex design patterns being needed. What happens, for example, if a system that has both a `SMIRNOFFLibraryChargeHandler` and a `SMIRNOFFChargeIncrementHandler` have assigned a parameter to the same `TopologyKey(atom_indices=(X,)` slot?
Should they be added? Or one override the other? and if so what is responsible for ensuring that this logic is enforced?

   Moving all of the matching logic under a single handler unifies such validation and choices, and does not require a parent object to orchestrate and enforce such logic.

3) Having a single slot map for electrostatics makes it easier to track which original parameter resulted in which partial charge. Currently one would need to look into the AM1, charge increment and library charge handlers to track this information down. This makes bookkeeping and aggregation during parameter optimisation significantly cleaner and easier. This is also true when trying to compute the energy of a topology based on its system object due to all of the related terms being cleanly grouped together.

4) Having a potential handler be able to store parameters from multiple handlers should make adding things like v-site support down the road easier. I.e. rather than have a `VSiteVdWHandler` and a `VSiteElectrostaticsHandler` one have the existing `vdW` and `Electrostatics` potential handlers consume the v-site parameter handler and themselves handle the required charge distribution and parameter matching logic.

@mattwthompson it would be good to get your input here.

**Note:** the 'tests' added here are to show the proof of concept and should be replaced with proper unit tests if this proposal moves ahead.

#### Questions

* Currently a potential key does not track the handler that assigned a particular parameter, only the 'id' (i.e. SMIRKS pattern). Maybe `PotentialKey` should allow a tuple as an id such that the id could be `PotentialKey(id=(handler_type, smirks_pattern))`?
* How should a potential handler communicate via its API which parameter handlers it can consume?

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
Thanks for putting all of this together - I'll need to think about it more deeply, but the motivations seems sound. In particular, pointing out that electrostatics potentially involve a "many parameters -> one slot" mapping makes a single handler sound like a good approach.

>  Maybe PotentialKey should allow a tuple as an id such that the id could be `PotentialKey(id=(handler_type, smirks_pattern))`?

Would it be better or worse to group these together into a single tuple vs. a separate field? I.e. what about `PotentialKey(id=smirks_pattern, handler_type=handler_type)`? Currently this question is avoided altogether because potential keys are only allowed to exist _within_ a handler, bypassing the need for a key to know where it should go. `handler_type` could maybe an optional field in order to cover electrostatics without carrying around redundant information for keys in valence handlers.

> How should a potential handler communicate via its API which parameter handlers it can consume?

We could store a sequence of accepted handler types in the class and expose them with a method `ElectrostaticsHandler.digestable_parameter_handlers` (named something better). I can see this being handled with a one-off solution to make an electrostatics handler work, but maybe it's worth considering how it looks for the other handlers, like the valence parameters that we implicitly assume will only ever digest one type of parameter handler?

One thing that comes to mind is if/how parameters should be applied after initial construction. This is probably a moot point if the construction delegates to other calls intelligently. Splicing some of this test code together, the idea is that these two snippets should produce the same result (I think), both the same charges and similar provenance


```python3
        parameter_handlers = [
            ElectrostaticsHandler(version=0.3),
            charge_increment_handler,
            library_charge_handler,
        ]

        potential_handler = SMIRNOFFElectrostaticsHandler.from_toolkit(
            parameter_handlers, top
        )
        potential_handler.partial_charges
```

```python3

        parameter_handlers = [
            ElectrostaticsHandler(version=0.3),
            charge_increment_handler,
        ]

        potential_handler = SMIRNOFFElectrostaticsHandler.from_toolkit(
            parameter_handlers, top
        )
        potential_handler.somehow_update_with_another_handler(
            library_charge_handler,
        )
        potential_handler.partial_charges

**Comment by SimonBoothroyd** [comment]
> Would it be better or worse to group these together into a single tuple vs. a separate field? I.e. what about PotentialKey(id=smirks_pattern, handler_type=handler_type)? Currently this question is avoided altogether because potential keys are only allowed to exist within a handler, bypassing the need for a key to know where it should go. handler_type could maybe an optional field in order to cover electrostatics without carrying around redundant information for keys in valence handlers.

This is along the lines that I was thinking, but I wasn't sure if it would make sense for the other potential handlers that you support (e.g. Foyer handlers).

> We could store a sequence of accepted handler types in the class and expose them with a method ElectrostaticsHandler.digestable_parameter_handlers (named something better). I can see this being handled with a one-off solution to make an electrostatics handler work, but maybe it's worth considering how it looks for the other handlers, like the valence parameters that we implicitly assume will only ever digest one type of parameter handler?

This is also what I was thinking (I think I went with `associated_handlers` but wasn't a huge fan of that...) but didn't want to make the initial PR too convoluted 🙂  It would ultimately work almost like dependency injection.

> One thing that comes to mind is if/how parameters should be applied after initial construction. This is probably a moot point if the construction delegates to other calls intelligently.

I think one could just have a function which internally calls `from_toolkit`, and then updates the slot / potential dicts as you suggest.

**Comment by mattwthompson** [comment]
Notes from a chat with @SimonBoothroyd 

* If one allows a `PotentialHandler` to store multiple `ParameterHandler` objects, a `PotentialKey` would need to know _which_ handler it's actually associated with
  * This is a feature we need 
* Some use cases in which different handlers need different metadata (like an `AM1BCCHandler` knowing the AM1 method, number of conformers, etc. - how would an `ElectrostaticsHandler` access this information?)
* The simple handlers, like bonds, are "one-to-one" in the sense that one parameter from a force field applies to one "slot." In general, we need to be concerned with "many-to-one," "one-to-many," and "many-to-many," each of which can be demonstrated with partial charge methods and/or virtual sites.
  * For potential keys in charge models, `mult` carries a meaning more like "the atom index of the original attribute (minus 1)," where the charge model could be charge increments, library charges, or something similar. This should be renamed, but it's hard to come up with a suitable name.
  * For topology keys in charge models, `mult` carries a meaning more like "this is the nth layering of a charge model into this atom index," which is fairly consistent with its meaning in topology keys associated with torsions. This should also be renamed to something.


We think that prerequisite PR needs to be done before this. It needs to
* Allow `PotentialHandler`s to be created from multiple `ParameterHandler`. The missing behaviors are
  * A `PotentialHandler` needs to know which types of `ParameterHandler`s they're allowed to digest
  * `PotentialHandler.from_toolkit` needs to allow the input of multiple parameter handlers, mostly as a bookkeeping check. This will probably only be used by electrostatics handlers at first.

Then, this can be fleshed out with unit tests - I will handle this and ask @SimonBoothroyd for review. Integration tests will not be a part of this PR, but may come later.

**Comment by mattwthompson** [comment]
General status update: 

> If one allows a PotentialHandler to store multiple ParameterHandler objects, a PotentialKey would need to know which handler it's actually associated with

Implemented in #207 

> Some use cases in which different handlers need different metadata

Unresolved

> A PotentialHandler needs to know which types of ParameterHandlers it's allowed to digest

Implemented in #204 

> PotentialHandler.from_toolkit needs to allow the input of multiple parameter handlers, mostly as a bookkeeping check. This will probably only be used by electrostatics handlers at first.

Implemented in 9a9bd31, but might be iterated on futher.

The simplest tests cases are implemented, but there's plenty more to do before this will feel useful:


* Handling charge method precedence
* In general missing a few dozen tests for non-trivially-simple behavior
* Updating the different uses of `mult`
* Would `TopologyKey.associated_handler` be useful? Given that it's a key, and cannot perform one-to-many mapping while only storing the atom indices.

**Comment by mattwthompson** [comment]
@SimonBoothroyd I could some feedback on the general direction I've taken this, if that's possible to do with a brief look at the changes. I have the three simple unit tests passing locally (won't pass on CI until the API breaks are fixed elsewhere, and I don't want to do that until this is working in its own scope).

**Comment by SimonBoothroyd** [comment]
@mattwthompson thanks for your work on this, things are coming along nicely!

> Implemented in 9a9bd31, but might be iterated on futher.

I think the `_from_toolkit` method added in 9a9bd31 duplicates most of the logic already implemented in [`SMIRNOFFElectrostaticsHandler.store_matches`](https://github.com/openforcefield/openff-system/blob/9a9bd31bf267a7994c184af0d72ccb571b88fac3/openff/system/components/smirnoff.py#L838). It looks like L581-L606 can be replaced with just `handler.store_matches(parameter_handlers, topology)` (note the use of `parameter_handlers` and not `parameter_handler`).

> Handling charge method precedence

This should [already be handled](https://github.com/openforcefield/openff-system/blob/9a9bd31bf267a7994c184af0d72ccb571b88fac3/openff/system/components/smirnoff.py#L758) I think if the above change is made.

> Would TopologyKey.associated_handler be useful? Given that it's a key, and cannot perform one-to-many mapping while only storing the atom indices.

Hmm I'm not sure I understand this point - could you perhaps give an example of where it would come in handy?

-----

I think a good next step (separate PR which should be merged before this) would be to try and replace the `System.from_smirnoff` [handler mapping code](https://github.com/openforcefield/openff-system/blob/9a9bd31bf267a7994c184af0d72ccb571b88fac3/openff/system/components/system.py#L175-L264) with a more dependency injection type approach now that `allowed_parameter_handlers` is exposed. In particular, it would seem that whole block could be replaced with something more like:

```
parameter_handlers_by_type = {
    force_field[parameter_handler_name].__class__
    for parameter_handler_name in force_field.registered_parameter_handlers
}

for potential_handler_type in [
    SMIRNOFFAngleHandler,    
    SMIRNOFFBondHandler,    
    SMIRNOFFConstraintHandler,    
    SMIRNOFFImproperTorsionHandler,    
    SMIRNOFFProperTorsionHandler,    
    SMIRNOFFvdWHandler,
    SMIRNOFFElectrostaticsHandler,
]:

    parameter_handlers = [
        parameter_handlers_by_type[allowed_type] 
        for allowed_type in potential_handler_type.allowed_parameter_handlers()
    ]

    potential_handler = potential_handler_type._from_toolkit(parameter_handlers, topology)

    sys_out.handlers[potential_handler.type] = potential_handler
```

SimonBoothroyd:
@mattwthompson I think 37a125b exposed a bug in `from_openmm` - namely the vdW parameters are all imported as 0.0 as the conversion only considers standard `NonbondedForce` and not the new split `CustomNonbondedForce` introduced in #201.

For now I've xfailed the test as it is unrelated to this PR but this should probably be fixed once #200 is in!

SimonBoothroyd:
Ok @mattwthompson this is now coming back all green and seems to be working well and efficiently. You may want to consider merging this soon and then handling the remaining issues (`mult`, individual charge handler provenance etc) in more targeted PRs as this one is now getting quite large!

mattwthompson:
Hmm, yeah that behavior was probably not implemented in #201. That's a lower priority feature so I'm fine breaking it between alpha releases.

All tests passing and 93% diff coverage is plenty good for a huge change like this one - hard to really review this so I do think the path forward is to merge this now/soon and build out the other fixes and tests in smaller PRs. Sound good?

SimonBoothroyd:
Sure that sounds good to me!

mattwthompson:
Thanks for all your help on this PR 🙂

PR #204, opened 2021-05-26
Let potential handlers specify allowed parameter handlers

mattwthompson:
### Description
This PR is a prerequisite for finishing #200.

I bet there's a more elegant way to to this - I had a difficult time figuring out how to use Pydantic features to validate something that's not stored in the model.

### Checklist
- [ ] Enforce this check in `ParameterHandler.store_mathces` and `ParameterHandler.store_potentials`? Maybe with a decorator?
- [ ] Let `ParameterHandler.from_smirnoff` accept a list of parameter handlers
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/system/components/smirnoff.py
+++ b/openff/system/components/smirnoff.py
@@ -1,4 +1,5 @@
-from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Type, TypeVar, Union
+import abc
+from typing import TYPE_CHECKING, Dict, Optional, Tuple, Type, TypeVar, Union
 
 from openff.toolkit.typing.engines.smirnoff.forcefield import ForceField
 from openff.toolkit.typing.engines.smirnoff.parameters import (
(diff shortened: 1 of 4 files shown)


mattwthompson:
@SimonBoothroyd I'm not sure if this is ready to go, but I could use help on two points:

1. Is there a better way to implement this, using some Pydantic/typing/etc. tricks? This may work but I didn't find it so satisfactory
2. On the topic of taking in multiple parameter handlers, what would this look like besides changing the function signature? I'm having a hard time thinking about test cases without implementing the electrostatics handler hierarchy in #200. Maybe save those tests for there?

SimonBoothroyd:
Thanks @mattwthompson I think this is a great start!

> Is there a better way to implement this, using some Pydantic/typing/etc. tricks? This may work but I didn't find it so satisfactory

My preference in these types of situations is usually to add a new classmethod which returns the allowed values, something like:

```
class SmirnoffPotentialHandler(..., abc.ABC):

    @classmethod
    @abc.abstractmethod
    def storable_parameter_handlers(cls):
        raise NotImplementedError ()

class MyHandler(...):
    @classmethod
    def storable_parameter_handlers(cls):
        return ["Bonds"]
```
In this way

* The return values are immutable and cannot accidentally be modified, and can be safely made public
* The ABC base class makes it clear that subclasses need to implement it
* It doesn't interfere with pydantic / get created for each instance of the class.

> On the topic of taking in multiple parameter handlers, what would this look like besides changing the function signature? I'm having a hard time thinking about test cases without implementing the electrostatics handler hierarchy in #200. Maybe save those tests for there?

I think the main other logic needed is in the `System.from_smirnoff` method, where we will need to add the 'dependency injection' logic which coordinates passing the right par handlers to the right pot handlers! Handling that here would probably make #200 a lot smaller!

mattwthompson:
Thanks for outlining things so clearly - it took me some time but that seems like a great solution! This allowed for some removal of unnecessary code in valence handlers that only had `from_smirnoff` redefined in order to enforce the type of the parameter handler argument.

PR #206, opened 2021-05-27
Add single-molecule energy comparisons against toolkit

### LINKED ISSUE #205: Interop: Torsion energy mismatches versus toolkit
I grabbed a convenient and small [set of molecules](https://github.com/openforcefield/open-forcefield-data/blob/master/Utilize-All-Parameters/selected/chosen.smi) and ran them through energy comparisons with very tight tolerances (1e-6 kJ/mol) and have found that some torsion energies differ by a range of values, **typically less than 1 J/mol** but sometimes as much as 0.2 kJ/mol.

mattwthompson:
### Description
Closes #205

Likely will look to replace this molecule set with something else that provides more coverage.

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
Some molecules are reporting differences in non-bonded energies, but only with RDKit. I'm not really sure why that would be the case and am not likely to dig into it in this scope.

PR #208, opened 2021-06-03
Update System.from_smirnoff with dependency injection logic

mattwthompson:
### Description
Implements the idea from https://github.com/openforcefield/openff-system/pull/200#issuecomment-853919970

Two major weak points are the convoluted constraints logic and messy electrostatics treatments; I'll fix the constraint logic later, and the electrostatics methods will be completely swapped out when this gets merged into #200. This method should be even slimmer once those changes are in place.

Everything passed locally, which is encouraging for what is a small refactor in a linchpin of this project. It might also point to corner cases that aren't yet covered in tests.

@SimonBoothroyd let me know if this is what you had in mind - the number of lines of code removed is a huge win in my book, especially given that the removed code was the messy result of several ideas I've tried over time.

### Checklist
- [ ] Add tests
- [x] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/system/components/system.py
+++ b/openff/system/components/system.py
@@ -196,7 +196,7 @@ class System(DefaultModel):
                 for allowed_type in potential_handler_type.allowed_parameter_handlers()
                 if allowed_type in parameter_handlers_by_type
             ]
-            
+
             if len(parameter_handlers) == 0:
                 continue
 
--- a/openff/system/components/system.py
+++ b/openff/system/components/system.py
@@ -190,14 +190,14 @@ class System(DefaultModel):
             SMIRNOFFProperTorsionHandler,
             SMIRNOFFvdWHandler,
         ]:
-            try:
-                parameter_handlers = [
-                    parameter_handlers_by_type[allowed_type]
-                    for allowed_type in potential_handler_type.allowed_parameter_handlers()
-                ]
-            except KeyError:
-                # TODO: Better handle case of this potential_handler_type not being associated
-                # with any registered parameter handlers
+
+            parameter_handlers = [
+                parameter_handlers_by_type[allowed_type]
+                for allowed_type in potential_handler_type.allowed_parameter_handlers()
+                if allowed_type in parameter_handlers_by_type
+            ]
+            
+            if len(parameter_handlers) == 0:
                 continue
 
             if len(parameter_handlers) == 0:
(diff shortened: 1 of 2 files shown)


openff/system/components/system.py
```diff
@@ -175,16 +175,37 @@ def from_smirnoff(
...
+        ]:
+            try:
+                parameter_handlers = [
+                    parameter_handlers_by_type[allowed_type]
+                    for allowed_type in potential_handler_type.allowed_parameter_handlers()
+                ]
+            except KeyError:
+                # TODO: Better handle case of this potential_handler_type not being associated
+                # with any registered parameter handlers
                 continue
```
SimonBoothroyd, line 201:
```suggestion

            parameter_handlers = [
                parameter_handlers_by_type[allowed_type]
                for allowed_type in potential_handler_type.allowed_parameter_handlers()
                if allowed_type in parameter_handlers_by_type
            ]
            
            if len(parameter_handlers) == 0:
                continue
```

openff/system/components/system.py
```diff
@@ -175,16 +175,37 @@ def from_smirnoff(
...
-                "Constraints",
-            }:
+        parameter_handlers_by_type = {
+            force_field[parameter_handler_name].__class__: force_field[
+                parameter_handler_name
+            ]
+            for parameter_handler_name in force_field.registered_parameter_handlers
+        }
+
+        # TODO: Wrap electrosotatics logic into this loop (#200)
```
SimonBoothroyd, line 185: 👍

openff/system/components/system.py
```diff
@@ -175,16 +175,37 @@ def from_smirnoff(
...
+                # TODO: Better handle case of this potential_handler_type not being associated
+                # with any registered parameter handlers
                 continue
-            elif parameter_handler_name == "Bonds":
+
+            if len(parameter_handlers) == 0:
+                continue
+
+            # TODO: Might be simpler to rework the bond handler to be self-contained and move back
+            # to the constraint handler dealing with the logic (and depending on the bond handler)
```
SimonBoothroyd, line 207: 👍

SimonBoothroyd (changes requested):
Thanks @mattwthompson for getting this in! I noticed just one minor thing that could be changed but otherwise LGTM.

In the future some more robust logic will probably need to be added to ensure that all parameter handlers in a FF were at least attempted to be consumed by at least one potential handler, and raise an exception otherwise. But this is beyond the scope of this PR I think!

mattwthompson:
Thanks again for the feedback!

PR #214, opened 2021-06-08
Refactor from_foyer and use new foyer API

### LINKED ISSUE #192: Foyer: Refactor from_foyer to a classmethod of System
It might be useful to move the free-floating function [`from_foyer`](https://github.com/openforcefield/openff-system/blob/e1a329d2cfcbfa3b64d5d8919ac8dcc69eed8715/openff/system/components/foyer.py#L67) into a classmethod (`System.from_foyer`). This should only necessitate moving that function; the other classes can stay put. See  #191 for some reasoning in the parallel case of SMIRNOFF typing.

cc: @umesh-timalsina @justinGilmer

umesh-timalsina:
### Description
Closes #192

### Checklist
- [x] Add tests : N/A
- [x] Lint
- [x] Update docstrings

### DIFF
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -32,6 +32,6 @@ dependencies:
   - gromacs >=2021
   - lammps
   - mbuild
-  - foyer>=0.8.0
+  - foyer>=0.8.1
   - panedr
   - pandas-stubs
--- a/openff/system/components/foyer.py
+++ b/openff/system/components/foyer.py
@@ -2,12 +2,10 @@ from abc import abstractmethod
 from copy import copy
 from typing import TYPE_CHECKING, Dict, Type
 
-from ele import element_from_atomic_number
 from openff.units import unit
-from openff.utilities.utilities import has_package, requires_package
+from openff.utilities.utilities import has_package
 
 from openff.system.components.potentials import Potential, PotentialHandler
-from openff.system.components.system import System
 from openff.system.models import PotentialKey, TopologyKey
 from openff.system.types import FloatQuantity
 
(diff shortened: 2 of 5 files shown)


mattwthompson:
@umesh-timalsina I see tests are passing - let me know if you would like this to be reviewed!

umesh-timalsina:
@mattwthompson Yes. I have removed the patch to create `ToopologyGraph` from `OFFTopology` as its available in latest release of foyer. Also, use of `NonBondedGenerator` from `OpenMM` is to get scaling factors is no longer necessary as its exposed by the Forcefield in foyer. 

However, I don't understand the motivation behind the use of `**kwargs` [here](https://github.com/openforcefield/openff-system/blob/47da0ad622b180d4b6e411a33e7a0367f86a2dd7/openff/system/components/system.py#L315).

mattwthompson:
I haven't thought about it much, but I assume that's the only way args, keyword args, or even named args could be passed to [`foyer.forcefield.Forcefield.apply`](https://github.com/mosdef-hub/foyer/blob/17576a2c74c2c643c7d6bf47a84cc0977ea35cdb/foyer/forcefield.py#L684-L697)?

devtools/conda-envs/test_env.yaml
```diff
@@ -32,6 +32,6 @@ dependencies:
   - gromacs >=2021
   - lammps
   - mbuild
-  - foyer>=0.8.0
+  - foyer>=0.8.1
```
mattwthompson, line 35: @umesh-timalsina @justinGilmer I looked at the [changelog](https://github.com/mosdef-hub/foyer/releases) and deduced this was the minimum version for this patch - let me know if that's not the case!

mattwthompson (approved):
One minor question but otherwise LGTM. Thanks!

devtools/conda-envs/test_env.yaml
umesh-timalsina: It looks like the graph conversion support was added in [`0.9.0`](https://github.com/mosdef-hub/foyer/releases/tag/0.9.0) for foyer and there's was a hot bugfix patch release as `0.9.1`.

mattwthompson: Okay, I'll bump it up to 0.9.1

PR #218, opened 2021-06-11
Initial virtual site support

mattwthompson:
### Description
I named this PR when I thought I'd want to create new models for each type of virtual site (there are four in the SMIRNOFF spec), but I think a better approach may be to treat them like `Potential` objects alongside the vanilla non-bonded parameters.

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
Going to point this to a staging branch instead

PR #222, opened 2021-06-16
Catch more cases of unsupported SMIRNOFF force fields

mattwthompson:
### Description
This beefs up error handling in some cases of unsupported SMIRNOFF force field contents, namely bond order interpolation (bonds and torsions) and virtual sites.

### Checklist
- [x] Add tests
- [x] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/components/interchange.py
+++ b/openff/interchange/components/interchange.py
@@ -114,13 +114,6 @@ class Interchange(DefaultModel):
                 continue
             if handler_name not in _SUPPORTED_SMIRNOFF_HANDLERS:
                 unsupported.append(handler_name)
-            if handler_name in {"Bonds", "ProperTorsions"}:
-                handler = force_field[handler_name]
-                if any([p.k_bondorder for p in handler.parameters]):
-                    raise SMIRNOFFHandlersNotImplementedError(
-                        "Bond order interpolation not supported. Found parameters with attribute"
-                        "`k_bondorder` in handler {handler_name}"
-                    )
 
         if unsupported:
             raise SMIRNOFFHandlersNotImplementedError(unsupported)
@@ -190,6 +183,7 @@ class Interchange(DefaultModel):
             #       move back to the constraint handler dealing with the logic (and
             #       depending on the bond handler)
             if potential_handler_type == SMIRNOFFBondHandler:
+                SMIRNOFFBondHandler.check_supported_parameters(force_field["Bonds"])
                 potential_handler = SMIRNOFFBondHandler._from_toolkit(
                     parameter_handler=force_field["Bonds"],
                     topology=topology,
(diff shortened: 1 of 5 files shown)


openff/interchange/tests/components/test_interchange.py
```diff
@@ -83,6 +73,61 @@ def test_basic_combination(self):
...
+        bogus_parameter_handler = ParameterHandler(version=0.3)
+        bogus_parameter_handler._TAGNAME = "bogus"
+        parsley.register_parameter_handler(bogus_parameter_handler)
+        with pytest.raises(
+            SMIRNOFFHandlersNotImplementedError, match="SMIRNOFF.*bogus"
+        ):
+            Interchange.from_smirnoff(force_field=parsley, topology=top)
+
+    def test_catch_bond_order_interpolation_bonds(self):
+        from openff.toolkit.tests.test_forcefield import xml_ff_bo
```
mattwthompson, line 90: I know this import is a bad pattern - I'd like to change how these XMLs are stored in the toolkit and then make those changes here.

mattwthompson:
@SimonBoothroyd  if you have a moment, this could maybe use a once-over. I'm also not sure what other SMIRNOFF parameters/attributes/handlers I should be worried about catching here; bond interpolation and virtual sites were what stuck out in my head.

openff/interchange/components/interchange.py
```diff
@@ -108,17 +109,20 @@ def _check_supported_handlers(cls, force_field: ForceField):
...
-                unsupported.append(handler)
+            if handler_name not in _SUPPORTED_SMIRNOFF_HANDLERS:
+                unsupported.append(handler_name)
+            if handler_name in {"Bonds", "ProperTorsions"}:
+                handler = force_field[handler_name]
+                if any([p.k_bondorder for p in handler.parameters]):
+                    raise SMIRNOFFHandlersNotImplementedError(
+                        "Bond order interpolation not supported. Found parameters with attribute"
+                        "`k_bondorder` in handler {handler_name}"
+                    )
```
SimonBoothroyd, line 123: I think this is a good idea! I wonder if rather than hard coding the exact unsupported attributes here, we could instead add a `supported_attributes` class method to each potential handler which explicitly states those that are currently supported, and then generalize the check here to make use of those? 

Having this info may also be useful when performing other validations!

SimonBoothroyd (commented):
I think this looks like a great addition @mattwthompson ! I definitely like the approach of 'be explicit about what isn't supported' rather than letting things silently pass through!

I made just one suggestion for you to consider.

openff/interchange/components/interchange.py
mattwthompson: That does sound like a better approach - I've implemented it and it passes all tests!

PR #232, opened 2021-06-22
Fix mypy

mattwthompson:
### Description
It works locally :tm:

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
Darn, the stubs aren't getting picked up. Locally I configured this via `export MYPYPATH=$MYPYPATH:/Users/mwt/software/openff-interchange/stubs/` and it works.

PR #234, opened 2021-06-23
Switch theme to openff-sphinx-theme

Yoshanuikabundi:
### Description
New theme!! Let me know if you have any thoughts, we have a lot of flexibility with this theme.

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
Thanks @Yoshanuikabundi  - I think this looks nice, and the unification is a huge plus. However, fI'd like to park this for a day or two while I get something more sensible off the ground. For example, the API docs are all completely broken, but mostly because everything tries to point to old modules names which no longer exist.

The only detail I could find right now is that some of the text in this sidebar looks clipped ((`compon` -> `components`): image

mattwthompson:
The above could/should probably be fixed with better-named headers. We'll see if I can figure that out

mattwthompson:
Blocked by #235, which should also be good to go

mattwthompson:
What's our status with this, #271, and #255? Let me know if something is waiting on my review, or if you'd prefer any particular order of merging!

Yoshanuikabundi:
I'm meeting with Simon to discuss the theme today and then I'll have more info about this one. As for #271 and #255, I was hoping to hear any of your preferences; we should close one and merge the other. As much as I love the idea of autoapi, I think autosummary is probably superior, especially since we can mock imports and since autodoc_pydantic is planning to [support this use case in the near future](https://github.com/mansenfranzen/autodoc_pydantic/issues/33#issuecomment-894810551).

mattwthompson:
Now that https://github.com/openforcefield/openff-toolkit/pull/979 is merged, I think this should be ready to go?

mattwthompson:
This seems to have slipped through the cracks, but I don't think anything should conflict with this going live.

PR #235, opened 2021-06-23
Overhaul docs

mattwthompson:
### Description
This PR adds a few doc examples that should be rendered in docs and also run in CI via `pytest --doctest-modules`

cc @umesh-timalsina @daico007 @justinGilmer In 1c5b886, I changed some function signatures to be more consistent and use `force_field` over a mix of `forcefield` and `ff`. API break but a `%s///g` kind of break

### Checklist
- [ ] Add tests
- [x] Lint
- [x] Update docstrings


mattwthompson:
I'm scope creeping this one to also include generally updating the docs to the new API, since they've been pretty much ignored for many months.

They build okay (not great) locally, which hopefully bodes well for the build on this PR. Then #234 is next up.

mattwthompson:
Sweet - it built on the first try! https://openff-interchange--235.org.readthedocs.build/en/235/index.html

mattwthompson:
I'm probably leaving this be for a few hours. One outstanding issue is the difference between what I get locally in some places, i.e. this locally 

image

vs built

image

cc: @Yoshanuikabundi

PR #182, opened 2021-05-12
Foyer tests

umesh-timalsina:
### Description
Add tests for foyer parameterized systems added in #114. Depends on #166.

### Some discussion points/comments on the PR.
- Setting gen-pairs to yes in the gmx topfile for OPLSAA Force Field is not necessary or might lead to energy differences.

- We currently have differences in the output top files specifically the default section

OpenFF-Defaults (ethanol system)
[ethanol_openff.txt](https://github.com/openforcefield/openff-system/files/6469374/ethanol_openff.txt)

Parmed Defaults (parameterized ethanol)
[ethanol.txt](https://github.com/openforcefield/openff-system/files/6469370/ethanol.txt)

- Largest energy differences were found for “Angles” and “Electrostatics”.

PR #237, opened 2021-06-23
Add gro decimal precision for `get_gromacs_energies`

daico007:
### Description
We realized some of the differences in our energy tests are due to differences in precision of the positions defined in the gromacs `.gro` file, but that variable could not be changed from the `get_gromacs_energies`. This PR added a `decimal` flag for the `get_gromacs_energies` to address that issue. Related to #182 

### Checklist
- [ ] Add tests
- [ ] Lint
- [x] Update docstrings

### DIFF
--- a/openff/interchange/drivers/gromacs.py
+++ b/openff/interchange/drivers/gromacs.py
@@ -128,7 +128,7 @@ def get_gromacs_energies(
     off_sys: "Interchange",
     mdp: str = "auto",
     writer: str = "internal",
-    decimal: int = 8
+    decimal: int = 8,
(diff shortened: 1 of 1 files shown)


mattwthompson (approved):
LGTM - good to merge once tests pass?

P.S. the other approach is to round positions before exporting: 
https://github.com/openforcefield/openff-interchange/blob/3803eb81b42015df2a0ee075a76e0373d319eb2a/openff/interchange/tests/integration_tests/test_interoperability_pathways.py#L75

I don't know which is a better approach; it's probably useful to have both available.

PR #245, opened 2021-07-08
Update partial charges in some tests

mattwthompson:
### Description
Leaving hanging until a 0.10.0 toolkit release (0.9.3 if I don't get my way on versioning). ~~Changes result from https://github.com/openforcefield/openff-toolkit/pull/948 https://github.com/openforcefield/openff-toolkit/pull/1007~~ hmm that's not merged. I'm not sure where these differences come from ... wrapped toolkit differences?

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,7 +30,7 @@ repos:
         'flake8-pytest-style',
     ]
 - repo: https://github.com/asottile/pyupgrade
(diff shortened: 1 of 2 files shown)


mattwthompson:
False alarm - the difference is simply due to me installing OpenEye toolkits in my development environment.

PR #247, opened 2021-07-08
Improve Amber support, move to using native writers

mattwthompson:
### Description
Starting to work through Amber-specific writers and tests.

### DIFF
--- a/openff/interchange/interop/internal/amber.py
+++ b/openff/interchange/interop/internal/amber.py
@@ -8,6 +8,31 @@ if TYPE_CHECKING:
     from openff.interchange.components.interchange import Interchange
 
 
+def to_prmtop(interchange: "Interchange", file_path: Union[Path, str]):
+    """
+    Write a .prmtop file. See http://ambermd.org/prmtop.pdf for details.
+
+    """
+    if isinstance(file_path, str):
+        path = Path(file_path)
+    if isinstance(file_path, Path):
+        path = file_path
+
+    with open(path, "w") as prmtop:
+        import datetime
+
+        now = datetime.datetime.now()
+        prmtop.write(
+            "%%VERSION  VERSION_STAMP = V0001.000  DATE = "
+            f"{now.month:02d}/{now.day:02d}/{(now.year % 100):02}  "
+            f"{now.hour:02d}:{now.minute:02d}:{now.second:02d}"
+        )
+
+        from openff.interchange.interop.internal.gromacs import _build_typemap
+
+        typemap = _build_typemap(interchange)  # noqa
+
+
 def to_inpcrd(interchange: "Interchange", file_path: Union[Path, str]):
     """
     Write a .prmtop file. See https://ambermd.org/FileFormats.php#restart for details.
--- a/openff/interchange/interop/internal/amber.py
+++ b/openff/interchange/interop/internal/amber.py
@@ -1,0 +1,43 @@
+from pathlib import Path
+from typing import TYPE_CHECKING, Union
+
+import numpy as np
+from openff.units import unit
+
+if TYPE_CHECKING:
+    from openff.interchange.components.interchange import Interchange
+
+
+def to_inpcrd(interchange: "Interchange", file_path: Union[Path, str]):
+    """
+    Write a .prmtop file. See https://ambermd.org/FileFormats.php#restart for details.
+
+    """
+    if isinstance(file_path, str):
+        path = Path(file_path)
+    if isinstance(file_path, Path):
+        path = file_path
+
+    n_atoms = interchange.topology.mdtop.n_atoms  # type: ignore
+    time = 0.0
+
+    with open(path, "w") as inpcrd:
+        inpcrd.write(f"\n{n_atoms:5d}{time:15.7e}\n")
+
+        fmt = "%12.7f%12.7f%12.7f" "%12.7f%12.7f%12.7f\n"
+        coords = interchange.positions.m_as(unit.angstrom)
+        reshaped = coords.reshape((-1, 6))
+        for row in reshaped:
+            inpcrd.write(fmt % (row[0], row[1], row[2], row[3], row[4], row[5]))
+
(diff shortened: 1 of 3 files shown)


davidlmobley:
I'm moving quickly here, so apologies if I'm missing something obvious, but it sounds like the current upshot is that there is something wrong with the prmtop files we are writing, since these are yielding incorrect energies in AMBER... Is that right?

I feel like I must be missing obvious.

But assuming I'm not, and the problem is with the prmtop files we are writing, then I'd probably want to:
a) cross-check against prmtop files for the "equivalent" system written the old way via ParmEd (we've previously validated things like AMBER<->GROMACS round-trips via ParmEd giving identical energies before/after conversion) so we can see how those differ from the ones we're writing
b) failing that, cross-check against converted files in some of the SAMPL repos (SAMPL6 in particular, I believe) which we had extensively validated as giving consistent energies

But maybe I am missing something where this is not just a problem with the prmtop we are writing.

SimonBoothroyd:
@mattwthompson the fix to the LJ and electrostatics issues should be in cea3ed8. The gist of it is that you weren't excluding 1-2 and 1-3 interactions in rings, and also you were ordering the A and B LJ coefficients as upper rather than lower triangle.

> But maybe I am missing something where this is not just a problem with the prmtop we are writing.

As I understand it ParmEd does not de-duplicate parameters when writing the prmtop while interchange does (:tada:) so you have a lot more info in the ParmEd output that can be tedious to sift through.

openff/interchange/interop/internal/amber.py
```diff
@@ -0,0 +1,630 @@
...
+            bond_indices = sorted(bond.atom_indices)
+
+            if atom1.element.atomic_number == 1 or atom2.element.atomic_number == 1:
+                bonds_inc_hydrogen.append(bond_indices[0] * 3)
+                bonds_inc_hydrogen.append(bond_indices[1] * 3)
+                bonds_inc_hydrogen.append(bond_type_index + 1)
+            else:
+                bonds_without_hydrogen.append(bond_indices[0] * 3)
+                bonds_without_hydrogen.append(bond_indices[1] * 3)
+                bonds_without_hydrogen.append(bond_type_index + 1)
```
SimonBoothroyd, line 136:
```suggestion
            bonds_list = bonds_inc_hydrogen if (atom1.element.atomic_number == 1 or atom2.element.atomic_number == 1) else bonds_without_hydrogen:
            
            bonds_list.append(bond_indices[0] * 3)
            bonds_list.append(bond_indices[1] * 3)
            bonds_list.append(bond_type_index + 1)
```

might help avoid a copy and paste!

openff/interchange/interop/internal/amber.py
```diff
@@ -0,0 +1,630 @@
...
+            ]:
+                angles_inc_hydrogen.append(angle_indices[0] * 3)
+                angles_inc_hydrogen.append(angle_indices[1] * 3)
+                angles_inc_hydrogen.append(angle_indices[2] * 3)
+                angles_inc_hydrogen.append(angle_type_index + 1)
+            else:
+                angles_without_hydrogen.append(angle_indices[0] * 3)
+                angles_without_hydrogen.append(angle_indices[1] * 3)
+                angles_without_hydrogen.append(angle_indices[2] * 3)
+                angles_without_hydrogen.append(angle_type_index + 1)
```
SimonBoothroyd, line 171: Same as above

openff/interchange/interop/internal/amber.py
```diff
@@ -0,0 +1,630 @@
...
+        if len(tmp) == 0:
+            tmp.append(0)
+
+        number_excluded_atoms.append(len(tmp))
+        [excluded_atoms_list.append(_) for _ in tmp]
+
+    return number_excluded_atoms, excluded_atoms_list
+
+
+def to_prmtop(interchange: "Interchange", file_path: Union[Path, str]):
```
SimonBoothroyd, line 64: You might want to consider splitting this mono-function into smaller focused ones to enable precise unit testing!

openff/interchange/interop/internal/amber.py
```diff
@@ -0,0 +1,630 @@
...
+            for values in [acoefs, bcoefs, nonbonded_parm_indices]
+            for value in values
+        ), "an internal error occurred"
+
+        prmtop.write("%FLAG NONBONDED_PARM_INDEX\n" "%FORMAT(10I8)\n")
+        text_blob = "".join([str(val).rjust(8) for val in nonbonded_parm_indices])
+        _write_text_blob(prmtop, text_blob)
+
+        prmtop.write("%FLAG RESIDUE_LABEL\n" "%FORMAT(20a4)\n")
+        prmtop.write("FOO\n")
```
SimonBoothroyd, line 407:
```suggestion
```

SimonBoothroyd: Is this a load-bearing FOO?

mattwthompson:
Holy moly, thanks a ton @SimonBoothroyd!

> you weren't excluding 1-2 and 1-3 interactions

Yes, I hadn't considered doing this since it does not seem at all intuitive to track 1-2 and 1-3 pairs in the list of known 1-4 pairs; ~~any pairs that are 1-2 or 1-3 on the short side of rings should also be found by a 1-4 pair-finding algorithm~~ ahhh,  I see here, when a pair is 1-4 along the long side of a ring but 1-2 or 1-3 on the short side, it shouldn't be tracked as a 1-4 at all.  At least that's what I've wrapped my head around - if I'm understanding the code, the variable `known_14_pairs` should maybe be renamed since it functions as a list of pairs that should not be considered, which previously was only the first time a particular 1-4 pair is found but now can be 1-2 and 1-3; the function is perfect though.

> you were ordering the A and B LJ coefficients as upper rather than lower triangle.

I suspected something like this but it's tricky to wrap my head around - thanks for catching!

openff/interchange/interop/internal/amber.py
mattwthompson: Definitely 👍

PR #248, opened 2021-07-09
Support virtual sites in GROMACS export

mattwthompson:
Commit history should clean up if/when #244 is merged

### TODO:
- [ ] Implement "charge increment" behavior
- [ ] Virtual sites need to know what residues they're in
  * Probably obtainable just by querying the residue of the anchor atom - but what happens if a virtual site is added at the boundary between two residues?
 - [ ] `MonovalentLonePair`
 - [ ] `DivalentLonePair`
 - [ ] `TrivalentLonePair`
- [ ] TIP4P example
- [ ] TIP5P example


mattwthompson:
Here's a notebook that adds a sigma hole virtual site to a chloromethane and runs it with GROMACS: [sigma_hole.ipynb.txt](https://github.com/openforcefield/openff-interchange/files/6793684/sigma_hole.ipynb.txt)

It probably won't work for anything more complicated than this, and it also does not apply the "charge increments" from the anchor atoms to the virtual sites.

mattwthompson:
Need to manually add extra exclusions for all atoms that are some number of atoms away from a virtual site or fewer: https://manual.gromacs.org/documentation/current/reference-manual/topologies/molecule-definition.html#exclusions

Also probably need to change `ptype` to `D`?

mattwthompson:
And here's a similarly partially-functional TIP4P implementation: [tip4p_broken.zip](https://github.com/openforcefield/openff-interchange/files/6811720/tip4p_broken.zip)

mattwthompson:
This is getting unwieldy, I'm going to accept the pain of merging an incomplete feature branch instead of suffering through a monstrous merge if/when this is fully implemented.

PR #255, opened 2021-07-21
Automatic API documentation from docstrings

Yoshanuikabundi:
### Description
This is a proof-of-concept of fully automatic API reference generation. The API is generated by [sphinx-autoapi](https://github.com/readthedocs/sphinx-autoapi), which works by parsing the code itself rather than importing it and inspecting it. This should make it easier to compile documentation without installing all the dependencies. The API reference follows the structure of the code itself, and relies on module-level docstrings to guide the reader through the code base.

Adopting this style of reference documentation will take pressure off maintainers to ensure that new code is present in the documentation. If it has a docstring, and it's a public part of a public module (ie, neither it nor any of its ancestors are prefixed with an underscore), it will be documented (modulo [this bug](https://github.com/readthedocs/sphinx-autoapi/issues/298)). On the other hand, it makes it more important to include module-level docstrings (at least a summary line), and might make it more important to be diligent about which parts of the API are public and which are private.

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
This looks really cool! My first impressions are that it looks clean and could be a long-term solution with relatively minor bending of the source code. I'm not sure what feedback to give right now, but I'll follow up later.

j-wags:
Agreed, I really like how this looks. Maybe we can bring this to the Toolkit soon, as well :-)

Yoshanuikabundi:
Ok, Sphinx-autoapi doesn't have support for autodoc_pydantic, so I'm going to have to port this API layout over to autosummary for that to work. That might be for the best, as autosummary is what everything is built around anyway, but I was excited to not need dependencies to build documentation.

mattwthompson:
@Yoshanuikabundi I think the current step in our plan is to have you merge master (now that it includes #267) into this branch and do some amount of review before merging. Feel free to merge yourself if you're happy with it, or let me know how I can help.

Without having thought about it much, I'm ambivalent as to whether Pydantic models must be made pretty in this PR or if that's okay to punt out to a future PR.

Yoshanuikabundi:
I've looked further into treating the Pydantic models properly. It doesn't seem to be possible with autoapi. Autodoc can do it with `autodoc_pydantic`, which produces very nice results, but also seems to have a few bugs that the Interchange codebase exposes. It also needs a little more work or quite a lot of hacking before it works with completely automated API reference generation using autosummary. 

Anyway, that work is all in #271. My preference between autosummary and autoapi probably tends towards autosummary; I prefer autoapi's way of doing things, but autosummary is a core part of Sphinx and is more actively developed and has more of an ecosystem around it.

mattwthompson:
Based on the above, I think we're leaning towards #271 over this (https://github.com/openforcefield/openff-interchange/pull/271#pullrequestreview-728017127).

mattwthompson:
Now that #271 is merged I'm closing this. Feel free to re-open if there is more to be considered.

PR #260, opened 2021-07-23
Initial virtual site support

mattwthompson:
### Description
This PR merges together #258 #248, which diverged from  #244  at some point.

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
https://github.com/openforcefield/openff-toolkit/pull/909 might not be scalable but it would probably be the easiest way to get virtual sites positions ...

mattwthompson:
This is already too big; merging and fixing/expanding feature scope in follow-ups.

PR #228, opened 2021-06-18
Support bond order-based parameter interpolation for harmonic bonds

mattwthompson:

PR #137, opened 2021-03-18
Expand Potential functionality for parameter interpolation

mattwthompson:
### Description
This PR adds a class that can potentially store multiple `Potential` objects. This may be useful for parameter in which the applied parameters (system parameters) are a linear combination of source parameters (force field parameters). This design simply stores the original parameters, as `Potential` objects, and their corresponding weighting. This allows for extracting the system parameters with the same getter as the original `Potential` class, but maintains the original data. Right now, it looks to the user the same as if the original data was not kept (see `test_interpolated_potentials` as of the first commit) but it may be possible to expose methods to update the internal data (weightings and/or parameters).

### Checklist
- [ ] Getters for internal data
- [ ] Setters for internal data (?)
- [x] Update matrix representations
- [ ] Implement sanity checks (i.e. some rows must sum to 1.0)?
- [ ] Update handlers in SMIRNOFF layer?
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

PR #263, opened 2021-07-28
Bond-order based interpolation of torsion force constants

mattwthompson:


mattwthompson:
This is a much smaller changeset than I expect it to be, but
* Most of the machinery already existed in the toolkit
* Wrapping individual parameters was done in #137
* Analogous functionality was aded to harmonic bond forces in  #228 
* Tightly testing this is hard because of run-to-run/machine-to-machine variance in computing fractional bond orders

PR #265, opened 2021-07-29
Add _get_parameters convenience functions

mattwthompson:
### Description

It can sometimes be cumbersome to look up parameters in the current design. In principle, most parameters should be get-able by only knowing some atom indices and the name of an associated handler, but actually doing so takes a few steps.

This snippet shows an individual bond's parameters can be accessed by saving each object along the lookup path (5 lines), collapsing the same process into unreadable golf (1 line), and the same functionality with either the new `Interchange._get_parameters` method 
 or directly with `PotentialHandler._get_parameters` which it calls.

```python3
In [1]: run readme.py
Warning: Unable to load toolkit 'OpenEye Toolkit'. The Open Force Field Toolkit does not require the OpenEye Toolkits, and can use RDKit/AmberTools instead. However, if you have a valid license for the OpenEye Toolkits, consider installing them for faster performance and additional file format support: https://docs.eyesopen.com/toolkits/python/quickstart-python/linuxosx.html OpenEye offers free Toolkit licenses for academics: https://www.eyesopen.com/academic-licensing

In [2]: from openff.interchange.models import TopologyKey

In [3]: topology_key = TopologyKey(atom_indices=(0, 1), associated_handler='Bonds')

In [4]: potential_key = out['Bonds'].slot_map[topology_key]

In [5]: potential = out['Bonds'].potentials[potential_key]

In [6]: parameters = potential.parameters

In [7]: parameters
Out[7]:
{'k': 758.0931772913 <Unit('kilocalorie / angstrom ** 2 / mole')>,
 'length': 1.092888378383 <Unit('angstrom')>}

In [8]: out['Bonds'].potentials[out['Bonds'].slot_map[TopologyKey(atom_indices=(0, 1), associated_handler='Bonds')]].parameters
Out[8]:
{'k': 758.0931772913 <Unit('kilocalorie / angstrom ** 2 / mole')>,
 'length': 1.092888378383 <Unit('angstrom')>}

In [9]: out._get_parameters('Bonds', (0, 1))
Out[9]:
{'k': 758.0931772913 <Unit('kilocalorie / angstrom ** 2 / mole')>,
 'length': 1.092888378383 <Unit('angstrom')>}

In [10]: out['Bonds']._get_parameters((0, 1))
Out[10]:
{'k': 758.0931772913 <Unit('kilocalorie / angstrom ** 2 / mole')>,
 'length': 1.092888378383 <Unit('angstrom')>}

```

There are going to be some sharp edges with the different ways that topology keys can be defined. This also should have a better name to avoid confusing it with matrix representations. For these reasons, I'm keeping it private, but I think it will be useful for powerusers (and me) in the interim.

### Checklist
- [x] Add tests
- [x] Lint
- [x] Update docstrings


mattwthompson:
Test failures are stochastic energy differences. Three separate molecules failing, each only in one run.

PR #266, opened 2021-07-29
LINT: Adopt flake8-typing-only-imports

mattwthompson:
### Description
For cases in which something is imported not to be used but only to be included in a type annotation, mypy allows us to sneak this into a `if TYPE_CHECKING:` block, which leads to that thing not being imported at runtime. This can provide speedups when the thing being imported is potentially heavy, such as `Interchange`. I was doing this manually (read: inconsistently) but found a flake8 plugin to automatically warn when this could be used. I've found some downsides in that it doesn't play too nicely with pydantic, and also is fairly slow.

https://github.com/snok/flake8-type-checking

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,6 +28,7 @@ repos:
         'flake8-bugbear',
         'flake8-absolute-import',
         'flake8-pytest-style',
+        'flake8-typing-only-imports',
     ]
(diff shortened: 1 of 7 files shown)


mattwthompson:
Hmm, the actual plugin seems to have been abandonde/moved into a different project, which produces different behavior. Scratching this.

PR #267, opened 2021-07-30
Add more docstrings

mattwthompson:
### Description
This PR attempts to add docstrings to all public classes, methods, functions, and modules. Those which shouldn't be public are now largely made private. Most aren't what I would characterize as complete, but they do exist.

Linting is handled by `pydocstyle`, with minimal configuration, in order to at least ensure a baseline level of consistency.

This should improve the results of #255, so I would prefer completing this PR before merging that one.

### DIFF
--- a/openff/interchange/components/interchange.py
+++ b/openff/interchange/components/interchange.py
@@ -1,3 +1,4 @@
+"""An object for storing, manipulating, and converting molecular mechanics data."""
 import warnings
 from copy import deepcopy
 from pathlib import Path
@@ -10,7 +11,7 @@ from openff.toolkit.typing.engines.smirnoff import ForceField
 from openff.utilities.utilities import requires_package
 from pydantic import Field, validator
 
-from openff.interchange.components.mdtraj import OFFBioTop
+from openff.interchange.components.mdtraj import _OFFBioTop
 from openff.interchange.components.potentials import PotentialHandler
 from openff.interchange.components.smirnoff import (
     SMIRNOFF_POTENTIAL_HANDLERS,
@@ -50,8 +51,11 @@ class Interchange(DefaultModel):
     """
 
     class InnerSystem(DefaultModel):
+        """Inner representation of Interchange components."""
+
+        # TODO: Ensure these fields are hidden from the user as intended
         handlers: Dict[str, PotentialHandler] = dict()
-        topology: Optional[OFFBioTop] = Field(None)
+        topology: Optional[_OFFBioTop] = Field(None)
         box: ArrayQuantity["nanometer"] = Field(None)  # type: ignore
         positions: ArrayQuantity["nanometer"] = Field(None)  # type: ignore
 
@@ -72,16 +76,20 @@ class Interchange(DefaultModel):
 
     @property
     def handlers(self):
+        """Get the PotentialHandler objects in this Interchange object."""
         return self._inner_data.handlers
 
     def add_handler(self, handler_name: str, handler):
+        """Add a ParameterHandler to this Interchange object."""
         self._inner_data.handlers.update({handler_name: handler})
 
     def remove_handler(self, handler_name: str):
+        """Remove a PotentialHandler in this Interchange object."""
         self._inner_data.handlers.pop(handler_name)
 
     @property
     def topology(self):
+        """Get the OpenFF Topology object in this Interchange object."""
         return self._inner_data.topology
 
     @topology.setter
@@ -90,6 +98,7 @@ class Interchange(DefaultModel):
 
     @property
     def positions(self):
+        """Get the positions of all particles."""
         return self._inner_data.positions
 
     @positions.setter
@@ -98,6 +107,7 @@ class Interchange(DefaultModel):
 
     @property
     def box(self):
+        """If periodic, an array representing the periodic boundary conditions."""
         return self._inner_data.box
 
     @box.setter
@@ -122,11 +132,11 @@ class Interchange(DefaultModel):
     def from_smirnoff(
         cls,
         force_field: ForceField,
-        topology: OFFBioTop,
+        topology: _OFFBioTop,
         box=None,
     ) -> "Interchange":
-        """Creates a new object by parameterizing a topology using the specified
-        SMIRNOFF force field.
+        """
+        Create a new object by parameterizing a topology with a SMIRNOFF force field.
 
         Parameters
         ----------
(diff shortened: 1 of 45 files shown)


mattwthompson:
@Yoshanuikabundi not sure if/how you'd go about reviewing these. Maybe it would be better for this to be merged and any errors corrected in later revisions?

openff/interchange/components/mbuild.py
```diff
@@ -10,11 +11,11 @@
 
 @requires_package("mbuild")
 def offmol_to_compound(off_mol: "Molecule") -> "mb.Compound":
-    """Covert an OpenFF Molecule into an mBuild Compound.
+    """
+    Covert an OpenFF Molecule into an mBuild Compound.
```
Yoshanuikabundi, line 15:
```suggestion
    Convert an OpenFF Molecule into an mBuild Compound.
```

openff/interchange/components/mbuild.py
```diff
@@ -48,11 +49,11 @@ def offmol_to_compound(off_mol: "Molecule") -> "mb.Compound":
 
 @requires_package("mbuild")
 def offtop_to_compound(off_top: "Topology") -> "mb.Compound":
-    """Covert an OpenFF Topology into an mBuild Compound.
+    """
+    Covert an OpenFF Topology into an mBuild Compound.
```
Yoshanuikabundi, line 53:
```suggestion
    Convert an OpenFF Topology into an mBuild Compound.
```

openff/interchange/components/interchange.py
```diff
@@ -292,6 +300,7 @@ def to_top(self, file_path: Union[Path, str], writer="internal"):
             to_top(self, file_path)
 
     def to_lammps(self, file_path: Union[Path, str], writer="internal"):
+        """Write this Interchange to a LAMMPS data file."""
```
Yoshanuikabundi, line 303: Any reason this method says "Write" when others say "Export"? This is the nittiest of nitpicks...

openff/interchange/components/potentials.py
```diff
@@ -143,6 +153,7 @@ def get_system_parameters(self, p=None):
         return jax.numpy.array(q)
 
     def get_mapping(self) -> Dict:
+        """Get a mapping between potentials and array indices."""
```
Yoshanuikabundi, line 156: Array indices in which array?

openff/interchange/components/potentials.py
```diff
@@ -161,12 +172,14 @@ def get_mapping(self) -> Dict:
...
 
     def parametrize(self, p=None):
+        """Return an array of system parameters, given an array of force field parameters."""
         if p is None:
             p = self.get_force_field_parameters()
 
         return self.get_system_parameters(p=p)
 
     def parametrize_partial(self):
+        """Return parametrize() modified with functools.partial gluing on `mapping`."""
```
Yoshanuikabundi, line 182: I'm not sure how familiar the typical Pythonista is with functools...

```suggestion
        """Return a function that will call `self.parametrize()` with arguments specified by `self.mapping`."""
```

This may not be correct as the parametrize method doesn't have a mapping argument and I didn't think `partial` does either so I might just be confused.

openff/interchange/components/smirnoff.py
```diff
@@ -151,25 +152,26 @@ class SMIRNOFFBondHandler(SMIRNOFFPotentialHandler):
 
     @classmethod
     def allowed_parameter_handlers(cls):
+        """Return a list of allowed types of ParameterHandler classes."""
         return [BondHandler]
 
     @classmethod
     def supported_parameters(cls):
+        """Return a list of supported parameter attributes."""
```
Yoshanuikabundi, line 160:
```suggestion
        """Return a list of supported parameter attribute names."""
```

openff/interchange/components/smirnoff.py
```diff
@@ -306,10 +308,12 @@ class SMIRNOFFConstraintHandler(SMIRNOFFPotentialHandler):
 
     @classmethod
     def allowed_parameter_handlers(cls):
+        """Return a list of allowed types of ParameterHandler classes."""
         return [BondHandler, ConstraintHandler]
 
     @classmethod
     def supported_parameters(cls):
+        """Return a list of supported parameter attributes."""
```
Yoshanuikabundi, line 316:
```suggestion
        """Return a list of supported parameter attribute names."""
```

openff/interchange/components/smirnoff.py
```diff
@@ -456,20 +464,21 @@ class SMIRNOFFProperTorsionHandler(SMIRNOFFPotentialHandler):
 
     @classmethod
     def allowed_parameter_handlers(cls):
+        """Return a list of allowed types of ParameterHandler classes."""
         return [ProperTorsionHandler]
 
     @classmethod
     def supported_parameters(cls):
+        """Return a list of supported parameter attributes."""
```
Yoshanuikabundi, line 472:
```suggestion
        """Return a list of supported parameter attribute names."""
```

openff/interchange/components/smirnoff.py
```diff
@@ -563,18 +572,19 @@ class SMIRNOFFImproperTorsionHandler(SMIRNOFFPotentialHandler):
 
     @classmethod
     def allowed_parameter_handlers(cls):
+        """Return a list of allowed types of ParameterHandler classes."""
         return [ImproperTorsionHandler]
 
     @classmethod
     def supported_parameters(cls):
+        """Return a list of supported parameter attributes."""
```
Yoshanuikabundi, line 580:
```suggestion
        """Return a list of supported parameter attribute names."""
```

openff/interchange/components/smirnoff.py
```diff
@@ -770,12 +782,14 @@ def allowed_parameter_handlers(cls):
...
```
Yoshanuikabundi, line 785:
```suggestion
        """Return a list of supported parameter attribute names."""
```

openff/interchange/utils.py
```diff
@@ -85,15 +86,15 @@ def get_partial_charges_from_openmm_system(omm_system):
...
+    """Ensure an OpenFF ForceField is represented as a dict and convert it if it is not."""
     if isinstance(forcefield, ForceField):
         return forcefield._to_smirnoff_data()
     elif isinstance(forcefield, OrderedDict):
         return forcefield
 
 
 def compare_forcefields(ff1, ff2):
-    """Compare dict representations of OpenFF ForceField objects fore equality"""
+    """Compare dict representations of OpenFF ForceField objects fore equality."""
```
Yoshanuikabundi, line 97:
```suggestion
    """Compare dict representations of OpenFF ForceField objects for equality."""
```

Yoshanuikabundi (approved):
I agree! I've had a quick squiz for typos and stuff like that and just have a few suggestions. I think this should be merged ASAP and then I can review it more carefully in situ in #255

I also feel I should point out that a file called `.pre-commit-config.yaml` has been modified. I don't understand what that is and I'm reviewing so I'm pointing it out just in case it's a mistake.

Thanks!

openff/interchange/components/interchange.py
mattwthompson: Not a nitpick - I think the consistency would be valuable

In general, my brain associates "write" with file and "export" with in-memory objects, but that distinction is not robust.

mattwthompson:
```suggestion
        """Export this Interchange to a LAMMPS data file."""
```

mattwthompson:
Thanks for finding the typos! One of my keyboards has a rebellious "d" key which I expected to be the cause of some troubles. Luckily it looks like the number of typos wasn't too bad.

I'm going to go ahead and merge this without totally cleaning up the matrix methods, which I haven't yet had enough coffee to find the right words for. It's also something that's likely subject to a major refactor soon.

> I also feel I should point out that a file called .pre-commit-config.yaml has been modified.

Yeah, the diff on that file is larger than it should be since an auto-update got sucked into this branch. The relevant bit was adding the [`pydocstyle`](http://www.pydocstyle.org/en/stable/usage.html#usage-with-the-pre-commit-git-hooks-framework) hook. (Also, raising the minimum % of docstrings that [`interrogate`](https://github.com/econchick/interrogate) requires.) The framework is basically automation for linting/style/etc tools; opt-in for developers but enforced with a CI hook: https://github.com/openforcefield/openff-interchange/blob/master/docs/developing.md#pre-commit

PR #269, opened 2021-08-02
Use OpenEye Toolkits in tests

mattwthompson:


mattwthompson:
@j-wags any idea why the OpenEye license isn't being detected here? I've tried to match the toolkit's script as closely as possible. The one-liner assertion works on my machine.

j-wags:
Ok, I re-added `openff-interchange` as a repo that has access to the secret, and CI seems to be working now. Kinda a weird UX - I almost left the page before I saw I needed to click "Save changes". Looks like your tests are failing successfully now :-)

mattwthompson:
Thanks @j-wags! It seems like I didn't click "save," whoops.

PR #271, opened 2021-08-04
Automatic API reference with Autosummary

Yoshanuikabundi:
### Description
This is an alternative to #255 that uses autosummary rather than autoapi to produce automatic API reference docs. Theoretically, this supports using autodoc_pydantic to cleanly document pydantic models etc.

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


Yoshanuikabundi:
I've tidied a few things up. I don't think the tests should be documented, so I've added an explicit exclusion for them to the template; I'd rather this was more semantic, but unfortunately autosummary doesn't skip undocumented modules. Can we make tests private by renaming it "_tests"? I didn't attempt this because I thought it would probably break everything.

mattwthompson (approved):
I think I've slowed this PR down enough, and after skimming the HTML output I'm pretty happy with how it looks. There are a few typos/errors in the source text (likely from me) and I would like to see the Pydantic models represented a little more compactly and/or elegantly. But those are smaller fixes that don't need to be addressed urgently or in this PR.


If you still prefer this over #255,  I think this is good to go or let me know if there's other specific feedback you're after. As a matter of logistics, I will likely be without internet until around 22 hours from the time of this post, so feel to merge this yourself (I think you have access to the big green button).

mattwthompson:
@Yoshanuikabundi could you please

* Merge in the `master` branch and fix this issue I came across: https://github.com/openforcefield/openff-interchange/pull/283#issuecomment-900467795
* Document how these files were generated? I assume it's a mixture of automation and hand-crafted art; I wouldn't expect this to be easy work for a newcomer to replicate in another project but I do think there is value in leaving a reasonable amount of breadcrumbs. There were also some developers at the core-devs call yesterday that were interested in this and #255, so it could see adoption in other projects.

After these details are finished up I think this should be good to merge!

Yoshanuikabundi:
Sorry, haven't been able to work as much as usual thanks to my brain and quarantine. I have merged in `master`. Strangely, I'm getting different behaviour locally and in CI on which test modules get included, which is probably because they use different versions of Sphinx. Could I get a list of modules that you want excluded? Is there any reason we can't just mark them as private with an underscore?

I'm thinking I'll write up how this one works in the developers guide, and then describe both it and #255 in a blog post on my personal blog (or the openff blog, I don't mind). Is that what you had in mind?

mattwthompson:
My apologizes if my tone is not constructive - of course, no rush on these items while other more important things are out there.


> Could I get a list of modules that you want excluded?

The ideal listing is

```
tree -L 1
.
├── __init__.py  # Not a module
├── _version.py  # Not a module
├── components
├── conftest.py  # Hidden
├── data  # Hidden (no init)
├── drivers
├── energy_tests  # Hidden (tests)
├── exceptions.py
├── interop
├── interoperability_tests  # Hidden (tests)
├── models.py
├── py.typed  # Not a module
├── testing
├── types.py
├── unit_tests  # Hidden (tests)
└── utils.py
```

If it's simpler to implement a regex-y thing that also hides `testing/` that's fine, too.

>  Is there any reason we can't just mark them as private with an underscore?

I think we discussed this in another thread, but I simply would rather keep the file/directory structure consistent with other projects if possible.

>  Is that what you had in mind?

Yes - this is above and beyond what I expected! If I was doing this I would probably just leave a long comment in a thread and not commit it to the actual doc files; safe to say your idea is better. We'd be happy to host it on the OpenFF blog, since this is pretty cool work you're doing.

Yoshanuikabundi:
> My apologizes if my tone is not constructive - of course, no rush on these items while other more important things are out there.

Nope! I'm just frazzled!

I think the easiest and most robust way to get that set of modules hidden is to just exclude them explicitly. This way we don't risk making a "cutest" or "fittest" or "contested" or whatever module down the line that gets left out of the documentation. So I've added a way to do that in the template. It's configured in conf.py with the `autosummary_context["exclude_modules"]` setting. Now that I've found a clean way to implement it I think this is the best solution after all - it means no one has to modify the template to change this list, configuration all happens in conf.py, and no one has to bend their module layout into a autosummary-shaped hole. And I've documented all of this in the developers guide.

I'm confused why this PR has so many files changed - I thought GH only showed changes between the HEAD and base branches, not every file touched by the HEAD branch. I've verified that the files are all correct, but I've also created a [branch](https://github.com/Yoshanuikabundi/openff-interchange/tree/autosummary-reference2) by cherry picking the actual feature commits on top of the current `master` if it makes you uneasy. 

Other than that, I think this is ready to merge?

mattwthompson:
> I'm confused why this PR has so many files changed 

I'm also confused, but I don't generally put a lot of value in a clean history on a young project since things like `openff/interchange/` -> `openff/system` and `tests` -> `unit_tests`, etc. have made it hard to traverse the history. When things are more stable we'll care more about that, but for now it's a matter of getting features in.


> Other than that, I think this is ready to merge?


I agree! Thanks for putting this together!

PR #283, opened 2021-08-17
Autosummary

mattwthompson:
Kicking the tires on #271 from a branch that has OpenEye license and has #281 merged in


mattwthompson:
Okay, things are mostly building fine and the logic in the template also ignores the three new `foo_tests` "modules." But the new `openff.interchange.testing` module isn't included, probably since I don't know how these files were generated. Here I'd expect the `.testing` module to show up:

image

PR #294, opened 2021-08-30
Add back old charge assignment tests

mattwthompson:
### Description
Some unit tests were inadvertently deleted in https://github.com/openforcefield/openff-interchange/commit/36bf5aaf973dfc44c7aee76415b033ee407b41d8. This adds them back, only to learn that at least one is now broken.


mattwthompson:
This was motivated by working on #295 in which I was pulling in OpenFF Recharge without using it in the source code or tests - whoops!

mattwthompson:
This should be revived - probably restarted altogether - in light of https://github.com/openforcefield/openff-toolkit/pull/1276 (https://github.com/openforcefield/openff-toolkit/actions/runs/2191452887)

PR #304, opened 2021-09-13
Add pytest-xdist to test envs

mattwthompson:
Apparently we are given multi-core VMs by default!?

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources


mattwthompson:
Down from around 10-12 minutes to 6-7 minutes. Pretty good, although I suspect most of the time is wasted by `sqm` or OpenMM building distance matrices.

Issue #307, opened 2021-09-21
Units: Some duck-typed quantities are not processed as expected


mattwthompson:
```
In [1]: import numpy as np

In [2]: from openmm import unit

In [3]: from openff.interchange.types import _from_omm_quantity

In [4]: x = np.float64(4.0) * unit.nanometer

In [5]: _from_omm_quantity(x)
---------------------------------------------------------------------------
UnitValidationError                       Traceback (most recent call last)
<ipython-input-5-6483d9324953> in <module>
----> 1 _from_omm_quantity(x)

~/software/openff-system/openff/interchange/types.py in _from_omm_quantity(val)
     84     #        return val_ * unit.Unit(str(unit_))
     85     else:
---> 86         raise UnitValidationError(
     87             "Found a openmm.unit.Unit wrapped around something other than a float-like "
     88             f"or np.ndarray-like. Found a unit wrapped around type {type(val_)}."

UnitValidationError: Found a openmm.unit.Unit wrapped around something other than a float-like or np.ndarray-like. Found a unit wrapped around type <class 'numpy.float64'>.

In [6]: x._value
Out[6]: 4.0
```

This is a strange result since `x` as a wrapped `np.float64` behaves like a built-in `float` in most use cases. The error stems from the [type-checking logic ](https://github.com/openforcefield/openff-interchange/blob/2448c3360e70a3064d265cfabe9c404947f4d302/openff/interchange/types.py#L71-L89) doing strict checks against expected classes and not allowing for ducky subclasses. That is, 

```
In [8]: isinstance(x._value, float)
Out[8]: True

In [9]: type(x._value) is float
Out[9]: False
```

I decided to be strict because of some gotchas like `isinstance(True, int)` returning `True`. I'm not sure if it would be better to allow subclasses to slip through and work back the gotcha cases or remain initially strict but work in the expected fallback cases. We should let use cases guide this decision.

mattwthompson:
I don't remember this being an issue lately. Also, https://github.com/openforcefield/openff-models/pull/21 and a forthcoming `IntQuantity` PR are hardening some of these classes.

PR #308, opened 2021-09-21
Handle NumPy duck-typing built-ins

mattwthompson:
### Description
This is a possible solution for #307. I ran into this case (a `np.float64` wrapped as an `openmm.unit.Quantity`) while working with ForceBalance. It's somewhat hard-coded to this specific case but should catch some others.

### Checklist
- [x] Add tests
- [x] Lint
- [ ] Update docstrings


mattwthompson:
This gets some tests passing on an experimental fork of ForceBalance. I'll merge soon if no other cases crop up.

https://github.com/leeping/forcebalance/compare/master...mattwthompson:interchange?expand=1

PR #309, opened 2021-09-21
Improve user guide

mattwthompson:
### Description
This PR adds an unknown amount of content to the  user guide.
- [ ] Migration guide

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/docs/using/migrating.md
+++ b/docs/using/migrating.md
@@ -1,0 +1,98 @@
+# Migrating
+
+##
+Where previously the OpenFF Toolkit was used to generate an OpenMM system:
+
+```python
+openmm_system = forcefield.create_openmm_system(topology)
+```
+
+the same objects can be used to generate an `Interchange` object:
+
+```python
+interchange = Interchange.from_smirnoff(force_field=forcefield, topology=topology)
+```
+
+From here, the `Interchange` object can be exported to files for OpenMM:
+
+```python
+openmm_system = interchange.to_openmm()
+```
+
+or other simulation enginegs:
+
+```python
+interchange.to_gro("out.gro")
+interchange.to_top("out.top")
+interchange.to_lammps("out.lmp")
+```
+
+Below is a minimal but complete example parameterizing an ethanol molecule with an OpenFF force field and creating an OpenMM System alongside GROMACS files:
+
+```python
+from openff.toolkit.topology import Molecule
+from openff.toolkit.utils import get_data_file_path
+from openff.toolkit.typing.engines.smirnoff import ForceField
+from openff.interchange.components.interchange import Interchange
+
+sdf_file_path = get_data_file_path("molecules/ethanol.sdf")
+molecule: Molecule = Molecule.from_file(sdf_file_path)
+topology: Topology = molecule.to_topology()
+
+sage = ForceField("openff-2.0.0.offxml")
+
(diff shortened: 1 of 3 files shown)


mattwthompson:
@Yoshanuikabundi I could use some help on the following points, please!

* General smithing of the small amount of writing I've done. Consider yourself to have the keys to the castle; I'll probably accept your suggestions wholesale.
* I've had a hard time putting pen to paper on the docs that users would actually benefit from and it's hard to ask users what docs they want because we don't have many users yet. The migration guide in this PR is largely guided by a one-on-one I had with @pavankum earlier in the week in which I helped him get Interchange into some of his workflows. I expect this general problem to work itself out over time, but any suggestions on how to jump the line / break the chicken-and-egg cycle?
* I manually converted the (small) changelog from rST to Markdown  and then copy-pasted from [here](https://github.com/openforcefield/openff-interchange/releases/tag/v0.1.1). Presumably we should add in links to the PRs and API points, but this also provides an opportunity to ask you: how would you recommend maintaining a changelog from the beginning, assuming you can do whatever you want with it?
* What's the one-liner to have docs auto-build? It's `sphinx autobuild` something or other ... I keep forgetting what it is and where it is so it might be useful to add it to the developer docs.

j-wags:
This is coming together nicely! I'm interested to see what Josh thinks.

> I've had a hard time putting pen to paper on the docs that users would actually benefit from and it's hard to ask users what docs they want because we don't have many users yet. 

For the developer's guide, I wonder if a good approach would be to make something like a rich bulleted list of the high-level design points and possible "gotchas" that you know about as the developer, which Josh could flesh out and ask for clarification on. 

In my experience, the user's guide is going to be hardest. I get a lot of inspiration for items to add to the users guide when I read over the OpenFF toolkit examples, and see the explanations for technical/conversion steps. So one good source of material for the user's guide can be figuring out if there's a common thread between "steps that require a lot of explanation" in the examples, because those can unearth the principles that are relevant for users to understand.

j-wags:
(bookkeeping post) -- @mattwthompson and I discussed how to get started on a user guide today, and we brainstormed a bit in [these meeting notes](https://openforcefield.atlassian.net/wiki/spaces/IN/pages/2134671423/2021+09+29+Thompson+Wagner+check+in). So our current plan is for Matt to provide a high-level outline with bullet points (similar to the one in the linked notes, but with more detail/revised organization) commit it to a markdown file in this PR, and then will ping @Yoshanuikabundi for feedback. 

Josh, feel free to ask for clarification and plan to iterate on this. We've done some cool things with a remote team in OpenFF, but asynchronously writing a users guide from scratch may be a new kind of lift. So let's keep track of lessons-learned from this process as we go!

mattwthompson:
Okay ... I'm a walking writers' block but I've managed to get some of my thoughts to paper. It's not done, nor will it ever be, but it's in a state that I consider suitable for feedback. This can be high-level (telling me what gaps I should aim to cover) or low-level (down to individual smithing). I have the bandwidth to dedicate much more time to this, but I need direction. Alternatively, I'm fine turning this over at any time for round of editing/smithing with the intent of merging it with whatever its current scope is.

Things look okay with minimal spot-checking in my `sphinx-autobuild docs docs/_build/html` tab. I tried to not touch important configurations, just add Markdown content.

PR #312, opened 2021-09-24
Add GROMACS topology reader and InterMol converter

mattwthompson:
### Description
This PR adds a not-fully-scoped GROMACS topology reader. As of today I can round-trip a test file into something that looks decent for the sections that are implemented; I have not yet tested if they will run.

It currently is little more than a converted from an InterMol object, which is responsible for the heavy lifting. I may continue to build out that conversion layer, but I will also start to refactor it out in favor of internal parsers. I have found InterMol's object models to be more straightforward so far, and I hope they are a more feasible starting point for refactoring than ParmEd.

Part of this necessarily chips away at #301, especially while upstream functionality in the toolkit is not yet being worked on.

Note that full - or even just generally useful - GROMACS parsers are blocked by the topology refactor.

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/components/base.py
+++ b/openff/interchange/components/base.py
@@ -1,0 +1,78 @@
+"""
+Base models for engine- and force field-agnostic components.
+"""
+from typing import Literal
+
+from openff.units import unit
+from pydantic import Field
+
+from openff.interchange.components.potentials import PotentialHandler
+from openff.interchange.types import FloatQuantity
+
+
+class BaseBondHandler(PotentialHandler):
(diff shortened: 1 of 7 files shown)


mattwthompson:
Could still
* Improve parsing of multiple types of dihedrals
* Make `interchange.from_gromacs` or something like that
* Refactor bits out into a `_from_intermol` sort of conversion

mattwthompson:
There's a lot more work to do, but I want to do those in different branches.

PR #326, opened 2021-11-01
Add host-guest example

mattwthompson:
### Description
This is the result of exploring how much has changed since Mobley wrote a host-guest example circa 2017. I don't have a nice notebook diff viewer installed here, but `Molecule.{to|from}_openeye(oemol)` is one of the biggest changes. I still have to go fully into OpenEye land to _do_ the docking, but the interface in and out is relatively clean. (It could probably be written up as a on-off `dock(host: Molecule, guest: Molecule)` sort of utility function.)

The last major roadblack is assigning charges; something isn't quite right when directly calling `oequacpac.OEAssignCharges` and `Molecule.assign_partial_charges` chokes while generating conformers.

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
This is now a year old and I have no immediate plans to finish this. If people do want a host-guest example or functionality that makes these systems easier, please reach out.

PR #328, opened 2021-11-04
Add protein-ligand example

mattwthompson:
### Description
This PR adds a protein-ligand example, starting with prepared data files and SMIRNOFF force fields and ending with some simulation inputs. There are several rough edges, but the basic structure is there.

The biopolymer branch of the toolkit is needed:
```shell
$ pip install git+https://github.com/openforcefield/openff-toolkit.git@topology-biopolymer-refactor
```

~~Here is the conda environment I have (including many things not needed here):~~ (old) https://gist.github.com/mattwthompson/b760f5b8b3cd04fea4c4b3c41f7ff73a

Here is a PDF of what I get while running it: [protein_ligand.pdf](https://github.com/openforcefield/openff-interchange/files/7476898/protein_ligand.pdf)
 
### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
Jotting down some known issues, things I ran into while working on this:

* Sometimes positions aren't properly processed when adding objects
* Adding objects together can produce not-so-useful errors when some parts have positions and others don't
* The exports to LAMMPS and GROMACS are really slow, and the Amber export is also not fast
* The ["should I tell the engine to constrain bonds" logic](https://github.com/openforcefield/openff-interchange/blob/65b5c349efdb5ecc9cd5a501eb55e3ddc32ba5e8/openff/interchange/drivers/utils.py#L8) is really slow on the protein (probably because of the large graph?) and probably be reworked

mattwthompson:
@Yoshanuikabundi I'm merging this now so as to avoid it languishing forever, but feel free to take a hacksaw to this notebook as you wish. It's not perfect but I feel like it's relatively representative of how our current infrastructure can handle this sort of workflow. I expect it to be a common entry point for users, so it's worth some extra eyes.

PR #331, opened 2021-11-10
Restore autodoc_pydantic sphinx extension

Yoshanuikabundi:
### Description
- Changes autosummary templates so that autodoc_pydantic can be switched back on (and switches it back on)
- Fixes all remaining documentation warnings (hopefully) and configures RTD to fail on warning, so if anything bad happens to the docs we'll know in a PR
- Moves RTD over to Mamba for faster, more reliable builds

### Checklist
- [x] Implement
- [x] Check that RTD builds correctly
- [ ] Make sure @mattwthompson  checks that my changes to the `T_` type variable in  `openff/interchange/components/smirnoff.py` are appropriate (I removed it because it was causing a spurious error, see https://github.com/sphinx-doc/sphinx/issues/9836 - It was only being used in one place so as far as I understand it was redundant. Pyright was complaining about it)


Yoshanuikabundi:
Hey @mattwthompson, I've had to prepend an underscore to Interchange.InnerSystem to get the docs to build. I think this has to do with a [sphinx bug](https://github.com/sphinx-doc/sphinx/issues/8919). I might be able to work around it by fiddling with the templates if this fix is unacceptable.

.readthedocs.yml
```diff
@@ -3,16 +3,13 @@
...
-python:
-  version: 3.8
-  install:
-    - method: setuptools
-      path: .
+  fail_on_warning: true
 
 conda:
-  environment: docs/environment.yml
+  environment: devtools/conda-envs/docs_env.yaml
```
mattwthompson, line 15: 👍

devtools/conda-envs/docs_env.yaml
```diff
@@ -1,5 +1,7 @@
 name: openff-interchange-docs
 channels:
+  # Skip installing CUDA to speed up builds
+  - jaimergp/label/unsupported-cudatoolkit-shim
```
mattwthompson, line 4: 👍

(though obviously we'll eventually want the fundamental OpenMM-no-CUDA problem fixed upstream so we're not using unsupported things in production!)

docs/index.md
```diff
@@ -22,7 +22,7 @@ maxdepth: 2
 
 using/intro.md
 using/design.md
-using/ouput.md
+using/output.md
```
mattwthompson, line 25: 😬   there's the signature of my work

docs/using/output.md
```diff
@@ -63,9 +54,15 @@ By default, this will separate non-bonded interactions into several different
 `openmm.Force` objects. To combine everything into a single
 `openmm.NonbondedForce`, use the `combine_nonbonded_forces=True` argument.
 
+## Amber
 
+Under construction!
```
mattwthompson, line 59: We can split this out into a separate PR, but this section is no longer under construction

openff/interchange/components/interchange.py
```diff
@@ -56,7 +56,7 @@ class Interchange(DefaultModel):
     .. warning :: This API is experimental and subject to change.
     """
 
-    class InnerSystem(DefaultModel):
+    class _InnerSystem(DefaultModel):
```
mattwthompson, line 59: Good by me; this class is really only partially-implemented, but it's meant to be private-ish and should have probably have the underscore already

openff/interchange/components/smirnoff.py
```diff
@@ -136,7 +134,7 @@ def store_matches(
     @classmethod
     def _from_toolkit(
         cls: Type[T],
-        parameter_handler: T_,
+        parameter_handler: PotentialHandler,
```
mattwthompson, line 137: No issues with renaming, the `[A-Z]_` pattern is me trying to follow convention. I'd like to keep the variable in place, though, since it's the easiest way to keep Mypy happy. Alternatively, getting it passing with this approach would work, though I'm not sure how to do that off the top of my head. This is a base class whose derived classes will probably get subclasses of `PotentialHandler` and I think `TypeVar` is one way to get that to work. Open to other solutions - I always feel behind the curve / just barely staying above water with the creative solutions to type annotations.

```suggestion
        parameter_handler: TP,
```

and the corresponding change on L74 (the UI won't let me suggest changes on deleted code ... 🤔 )

mattwthompson (approved):
This is really slick! I have no idea what's going on with the Jinja templating, but pulling up an [arbitrary model](https://openff-interchange--331.org.readthedocs.build/en/331/_autosummary/openff.interchange.components.smirnoff.SMIRNOFFBondHandler.html#openff.interchange.components.smirnoff.SMIRNOFFBondHandler) shows cleanly rendered docs.

I've found only one thing I'd like changed here - everything is looks great.

Blocker:
* See the comment about Mypy/ `T_` - anything that gets Mypy passing is fine by me

Non-blockers / things to do later:
* Update the Amber section of `using/output.md`
* It would be nice to have this show the functions themselves, or names/aliases, not a memory address-looking thing. I have no idea how to do this and if it takes more than 30 seconds to fix I'd like to punt on it and get this PR merged as-is.
image

docs/index.md
Yoshanuikabundi: Nope! Pretty sure that was me actually

openff/interchange/components/smirnoff.py
Yoshanuikabundi: Oh I thought MyPy would accept subclasses. Can you link me to some examples of that convention? It would help me make the case to fix this bug upstream.

Yoshanuikabundi: Actually I think the problem here might be that there's actually an incorrect type being passed, and that the `bound` argument takes an actual type rather than a string. MyPy's error is:

```
openff/interchange/components/interchange.py:253: error: Argument "parameter_handler" to "_from_toolkit" of "SMIRNOFFPotentialHandler" has incompatible type "List[Any]"; expected "PotentialHandler"  [arg-type]
```

Which AFAIK should not be fixed by having a TypeVar with the PotentialHandler bound. MyPy also [claims](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#class-types) that types should accept their subclasses. I think maybe the argument is incorrectly typed and the TypeVar is masking this because it's incorrectly accepting a string as the Any type? This might be something to leave for a future PR though.

Yoshanuikabundi:
@mattwthompson I've spent 10 minutes looking and can't find a simple way to do that - it's just showing the output of the objects `__repr__` method I think, so fixing it will probably need changes in autodoc_pydantic itself like a post-processing step that converts function reprs to links to the function. Which would probably be a great feature for Sphinx generally now that you mention it.

Have fixed that blocking issue (but see my comments on your review - I think the TypeVar is masking that the type is incorrect), so I'll go ahead and merge!

openff/interchange/components/smirnoff.py
mattwthompson: Looks like my impression of the convention was wrong ... the docs do not tend to have the trailing underscore. I must have developed the habit reading others' code.

I think you're right ... the electrostatics treatment involves taking a list of handlers and probably other issues as well. I'm fine with the false positive as is, it can be fixed in the future as typing further creeps into things.

mattwthompson:
> I've spent 10 minutes looking and can't find a simple way to do that 

Is this re: function names? If so, I understand and don't think it's worth more than the time you spent as we can work on that later.

PR #333, opened 2021-11-11
Make beta/RC tests run automatically

mattwthompson:
This steals the simplicity of https://github.com/openforcefield/openff-toolkit/pull/1111


mattwthompson:
The JAX tests have been failing for several months because one of the feedstocks is broken. It might be fixed soon: https://github.com/conda-forge/jaxlib-feedstock/pull/65

PR #349, opened 2021-12-08
Optionally store velocities

mattwthompson:
### Description
This PR adds an optional `velocities` field to the core object model and updates some exporters accordingly.

Note that exporting to `openmm.Context` is not currently supported, so users will have to `.setVelcoties()` themselves.

### Checklist
- [x] Add velocities to `Interchange` model
- [ ] Write velocities to `.gro` file
- [ ] Write velocities to `.inpcrd` file
- [ ] Write velocities to `.lmp` file
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
Will add this to exports later should they be used

PR #359, opened 2021-12-16
fix issue with python 3.10

mikemhenry:
### Description
You will need to quote the python version when you add python 3.10 eventually

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -29,7 +29,7 @@ jobs:
 
     - uses: conda-incubator/setup-miniconda@v2.1.1
       with:
-        python-version: ${{ matrix.python-version }}
+        python-version: "${{ matrix.python-version }}"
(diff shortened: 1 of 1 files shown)


mattwthompson (approved):
LGTM - I think I saw something about this a few months ago when it was released, but have yet to be burnt by it so I didn't change anything.

mikemhenry:
Yah at first I tried just adding quotes in the list, but needed it where it defined the python version in the yml and still failed, I caught it when I used your GHA yml as inspiration for another repo

mattwthompson:
Yeah, It'll be a while until our stack is compatible with 3.10 - though you're closer to the tap with OpenMM being of the current blockers.

mikemhenry:
:see_no_evil:

PR #375, opened 2022-01-18
Do not modify users' `Topology` objects

mattwthompson:
This is a port of #350 targeting v0.2.x / Toolkit v0.11.x

PR #376, opened 2022-01-19
Move force fields into pytest fixtures

mattwthompson:
### Description
I've found myself re-loading the same force fields several times; pytest fixtures are a cleaner way to do this.


mattwthompson:
Snuck this into #375

PR #382, opened 2022-01-31
Fix RTD build and update Sphinx

### LINKED ISSUE #378: Docs builds broken
@Yoshanuikabundi if you have some bandwidth in the next week or so, could you take a look at the RTD builds? One such broken [build](https://readthedocs.org/projects/openff-interchange/builds/15845380/). The warnings are confusing (why is just one of the submodules not found by autosummary?) and it's not clear at all from the log why the build _fails_, assuming that the autosummary warnings are not supposed to be fatal.

I can build locally using the instructions in the  developer docs (including making a whole new environment) and things look okay on first pass.

Yoshanuikabundi:
### Description
Closes #378, which was due to `openff/interchange/components/foyer.py` having a `foyer` import outside the check that the package exists. This caused imports of this particular module to fail when foyer is unavailable, such as in the RTD environment. Tests that imports are successful on different environments might be worth looking into. 

This PR also updates Sphinx to 4.4.0, which comes with a lot of improvements to the API docs (and has a marginally more informative error message for the above problem)

### Checklist
- [ ] Add tests
- [x] Lint

### DIFF
--- a/devtools/conda-envs/docs_env.yaml
+++ b/devtools/conda-envs/docs_env.yaml
@@ -22,6 +22,6 @@ dependencies:
   - myst-parser
   - numpydoc
   - autodoc-pydantic
-  - sphinx=4.2.0
+  - sphinx>=4.4.0,<5
   - pip:
     - git+https://github.com/openforcefield/openff-sphinx-theme.git@main
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -75,6 +75,7 @@ autosummary_context = {
         "openff.interchange.tests",
     ],
 }
+autosummary_ignore_module_all = False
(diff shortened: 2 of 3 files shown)


mattwthompson:
Thanks! LGTM.

Out of curiosity, was the issue only evident after the bump to 4.4, or did I miss something in the logs prior? I was scratching my head at why the builds were failing - I wouldn't have thought of bumping the sphinx version.

mattwthompson (approved):

Yoshanuikabundi:
4.4.0 says "couldn't import this module" and provides a cut down traceback, whereas earlier versions just say "this module doesn't exist". So its definitely more evident after the bump!

PR #385, opened 2022-02-03
 Error out if passed a force field missing required Electrostatics

mattwthompson:
### Description
A quick pass at safeguarding against https://github.com/openforcefield/openff-toolkit/issues/716

> The desired behavior would be to raise an exception if the ESHandler doesn't exist, but the system has partial charges that will interact.

### Checklist
- [x] Add tests
- [x] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/components/toolkit.py
+++ b/openff/interchange/components/toolkit.py
@@ -1,9 +1,13 @@
 """Utilities for processing and interfacing with the OpenFF Toolkit."""
-from typing import Dict, Union
+from typing import TYPE_CHECKING, Dict, Union
 
 import numpy as np
-from openff.toolkit.topology import Molecule, Topology
-from openff.toolkit.utils.collections import ValidatedList
+from openff.toolkit.topology import Topology
+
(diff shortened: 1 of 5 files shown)


openff/interchange/components/toolkit.py
```diff
@@ -93,3 +97,32 @@ def _combine_topologies(topology1: Topology, topology2: Topology) -> Topology:
...
+        "GBSA",
+    ]
+
+    # A more robust solution would probably take place late in the parameterization
+    # process, but this solution should cover a vast majority of cases with minimal
+    # complexity. Most notably this will *not* behave well with
+    #   * parameter handler plugins
+    #   * future additions to -built-in handlers
+    #   * handlers that _could_ assign partial charges but happen to not assign
+    #       any for some particular topology
```
j-wags, line 122: These are fantastic details to capture. Maybe this could be a design note for interchange plugins - Like, one potentialhandler can specify that it requires the existence of another (eg ToolkitAM1BCC requires Electrostatics?)

PR #392, opened 2022-02-11
Remove `_OFFBioTop`

mattwthompson:
### Description
Requires the completion of https://github.com/openforcefield/openff-toolkit/pull/1179

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
Docs will fail while we pull from feature branches. I don't think it's worth the effort to keep `docs_env.yaml` in sync for every merge.

PR #395, opened 2022-02-21
Update constraint behavior to match toolkit

### LINKED ISSUE #394: Bug: TIP3P constraints in Sage are not processed
This appears to happen for water but not typical hydrogen-containing bonds (whether the other atom is oxygen or not)
```
from openff.units import unit
from openff.toolkit.topology.molecule import Molecule
from openff.toolkit.typing.engines.smirnoff.forcefield import ForceField


topology = Molecule.from_smiles("CO").to_topology()
topology.box_vectors = [4, 4, 4] * unit.nanometer
print(force_field.create_openmm_system(topology, use_interchange=False).getNumConstraints())
print(force_field.create_openmm_system(topology, use_interchange=True).getNumConstraints())

topology = Molecule.from_smiles("O").to_topology()
topology.box_vectors = [4, 4, 4] * unit.nanometer
print(force_field.create_openmm_system(topology, use_interchange=False).getNumConstraints())
print(force_field.create_openmm_system(topology, use_interchange=True).getNumConstraints())
```
Out:
```
4
4
3
0
```

mattwthompson:
### Description
Resolves #394 

### Checklist
- [x] Add tests
- [x] Lint
- [ ] Update docstrings


mattwthompson:
Yesterday @j-wags and I decided on a path forward
* For now, move forward with this update of matching the toolkit's behavior in a way that configures and/or exposes a toggleable switch for turning this behavior on and off
* After the OpenFF stack gets released with biopolymer support, open a PR that proposed to modify this and gather input from relevant stakeholders (OpenMM developers). It's unclear if a change would be an improvement, but it would be useful to clarify what users in general expect here.

This assumes that both behaviors are acceptable in general, something I currently think is the case. But it makes things easier to match the toolkit's behavior because that's the current reference.

PR #398, opened 2022-02-25
Add safeguard test for TIP3P parameters in Sage

mattwthompson:
This test duplicates behavior that will be fixed upstream (and by nature won't pass until it is) https://github.com/openforcefield/openff-toolkit/issues/1199


mattwthompson:
This is fixed upstream (#1200) but won't be in the right branch until it's ported and/or everything is merged back in.

PR #404, opened 2022-03-04
Split out examples in CI, trim some heavy/unused test dependencies

mattwthompson:
I _think_ that using https://github.com/codecov/codecov-action removes the need to install the `codecov` conda package.

### DIFF
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -1,5 +1,6 @@
 name: openff-interchange-env
 channels:
+  - jaimergp/label/unsupported-cudatoolkit-shim
   - conda-forge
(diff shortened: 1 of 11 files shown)


devtools/conda-envs/test_env.yaml
```diff
@@ -22,20 +23,17 @@ dependencies:
   - mdtraj
   - intermol
   - nglview
-  - openff-evaluator
```
mattwthompson, line 25: I found with some `conda-tree` foo that this was bringing in a lot of things that weren't used - in particular Yank, which pulls down the entire Jupyter ecosystem for some reason. It is used in tests and I'd like to bring it back or move the functionality I need out of it.

PR #409, opened 2022-03-09
Support cosine squared angles

mattwthompson:
### Description
This PR adds support for [cosine squared](https://hoomd-blue.readthedocs.io/en/stable/module-md-angle.html#hoomd.md.angle.cosinesq) angle potentials and may or may not be a springboard to user-defined potentials.

### Checklist
- [x] Add tests
- [x] Lint
- [ ] Update docstrings


mattwthompson:
Something I've run into while swapping out the `HarmonicAngleForce` for a `CustomAngleForce` is that the latter drops unit-checking. 

In `HarmonicAngleForce`:
```python3
    def addAngle(self, particle1, particle2, particle3, angle, k):
        r"""
        addAngle(self, particle1, particle2, particle3, angle, k) -> int
        Add an angle term to the force field.

        Parameters
        ----------
        particle1 : int
            the index of the first particle forming the angle
        particle2 : int
            the index of the second particle forming the angle
        particle3 : int
            the index of the third particle forming the angle
        angle : double
            the equilibrium angle, measured in radians
        k : double
            the harmonic force constant for the angle, measured in kJ/mol/radian^2

        Returns
        -------
        int
            the index of the angle that was added
        """

        if unit.is_quantity(angle):
            angle = angle.value_in_unit(unit.radian)
        if unit.is_quantity(k):
            k = k.value_in_unit(unit.kilojoule_per_mole/(unit.radian*unit.radian))


        return _openmm.HarmonicAngleForce_addAngle(self, particle1, particle2, particle3, angle, k)
```

And in `CustomAngleForce`:
```python3
    def addAngle(self, *args):
        r"""
        addAngle(self, particle1, particle2, particle3, parameters=std::vector< double >()) -> int
        Add an angle term to the force field.

        Parameters
        ----------
        particle1 : int
            the index of the first particle connected by the angle
        particle2 : int
            the index of the second particle connected by the angle
        particle3 : int
            the index of the third particle connected by the angle
        parameters : vector< double >
            the list of parameters for the new angle

        Returns
        -------
        int
            the index of the angle that was added
        """
        return _openmm.CustomAngleForce_addAngle(self, *args)
```

I'd love to not require the user to units of each parameter alongside everything else, but there does need to be some way of ensuring the energy function is evaluated properly.

mattwthompson:
Re: above comment, this has something to do with custom forces being implemented in C++, not Python. Just have to be mindful of it.

PR #410, opened 2022-03-09
Trim some long tests

mattwthompson:
### Description
I thought I included this in #404 but it seems I did not.


openff/interchange/tests/energy_tests/test_energies.py
```diff
@@ -141,6 +141,7 @@ def test_energies_single_mol(
     @needs_gmx
     @needs_lmp
     @pytest.mark.slow()
+    @pytest.mark.skip(reason="Revisit after performannce optimizations in writers")
```
mattwthompson, line 144: This one in particular is pretty bad, taking roughly 8 minutes on my hardware. I'll look into fixing it another time.

Issue #408, opened 2022-03-09
Docs updates


Yoshanuikabundi:
# TODOS:
- [x] Rearrange and expand user guide (#428)
- [ ] Polish examples

# Explanation

Hey @mattwthompson I've just gone through the protein-ligand example from #328 - It looks really good! it's not too long and everything's well explained. I'll organise it with headings and might rephrase or re-organise things but my big notes are about the API.

First,  I love that the `+` operator works with `Interchange` objects!!!

Secondly, I've talked to @j-wags about this a little, but `Interchange` seems to me to basically be `Topology` on steroids (plus an FF and positions). In this example in particular, the `Topology` is only really constructed to write a file and construct the Interchange. I wonder if allowing `Interchange.from_smirnoff()` to accept a list of `Molecule` objects as an alternative to a `Topology` could simplify the API? 

Even if it uses `Topology` under the hood, my suspicion is that this would replace most usages of `Topology` by users. I think this would be a huge win for teaching, because `Topology` is a pretty nebulous term in MD world, and the type at the moment is really the only intermediate type. Without `Topology`, the main workflow for applying a force field takes `Molecule` objects, a `ForceField` object, a `position` array, and a `boxvectors` array, and combines them into an `Interchange` object. This is a beautiful workflow with no intermediate steps where each input corresponds exactly to a comp chem concept and half of them are numpy arrays that everyone has used a million times before. 

So my suggestion here is:

```diff
 @classmethod
 def from_smirnoff(
     cls,
     force_field: ForceField,
-    topology: Topology,
+    topology: Union[Topology, List[Molecule]],
     box=None,
 ) -> "Interchange":
 ```
 
 (And being able to specify positions in this method probably makes sense too, though it wouldn't come up in this example)

Would you accept PRs for this change?

Finally, my energies at the end were a bit strange:

```
OpenMM Energies:

Bond:          		11279.70703125 kJ / mol
Angle:         		5386.8837890625 kJ / mol
Torsion:       		7612.54052734375 kJ / mol
Nonbonded:     		None
vdW:           		5095140.218750465 kJ / mol
Electrostatics:		-20698.04012266005 kJ / mol

AMBER Energies:

Bond:          		2695.9208 kcalories_per_mole
Angle:         		1287.4972 kcalories_per_mole
Torsion:       		1819.4409 kcalories_per_mole
Nonbonded:     		None
vdW:           		5095170.807108 kjoule_per_mole
[...]

PR #411, opened 2022-03-10
Allow from_smirnoff to take a list of Molecules instead of a Topology

Yoshanuikabundi:
### Description
As discussed in #408, this simplifies an important route through the API by removing the need for the user to construct a `Topology` directly. This should allow many users to remove `Topology` from their mental model of the OpenFF ecosystem.

### Checklist
- [x] Add tests
- [x] Lint
- [x] Update docstrings

### DIFF
--- a/openff/interchange/components/interchange.py
+++ b/openff/interchange/components/interchange.py
@@ -3,9 +3,10 @@ import time
 import warnings
 from copy import deepcopy
 from pathlib import Path
-from typing import TYPE_CHECKING, Dict, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Union
 
 import numpy as np
+from openff.toolkit.topology.molecule import Molecule
 from openff.toolkit.topology.topology import Topology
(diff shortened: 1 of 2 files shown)


mattwthompson (approved):
Thanks - I now appreciate the simplicity. I moved the check inside the validator, which may or may not be an issue in the future for non-SMIRNOFF/OpenFF objects.

openff/interchange/components/interchange.py
```diff
@@ -196,15 +202,16 @@ def from_smirnoff(
             >>> from openff.toolkit.typing.engines.smirnoff import ForceField
             >>> mol = Molecule.from_smiles("CC")
             >>> mol.generate_conformers(n_conformers=1)
-            >>> top = Topology.from_molecules([mol])
```
mattwthompson, line 199: This is no different than your proposal but seeing it in code helped me appreciate how this unlocks some uses cases that completely skip the need for users to understand what a `Topology` is - but of course adds no friction to users that do want to make an manipulate them.

PR #414, opened 2022-03-11
Add non-OpenEye jobs to CI matrix

mattwthompson:
I realize that for a while now, CI runners _always_ had access to OpenEye toolkits. I'm not interested in exhaustively testing toolkit wrapper behaviors (that's a job for the OpenFF Toolkit) but I do want to ensure a developer without an OpenEye license can run the test suite without major surprises.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,6 +24,12 @@ jobs:
         python-version:
           - "3.8"
           - "3.9"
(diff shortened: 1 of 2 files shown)


mattwthompson:
The matrix is configured from yoinking parts of the toolkit's current coniguration: https://github.com/openforcefield/openff-toolkit/pull/1125

PR #415, opened 2022-03-14
More virtual site fixes

mattwthompson:
This is built from #244 using a yet-to-be-published test suite written by @j-wags


mattwthompson:
I'm looking at how the exclusion policies are implemented. One of them (`minimal`) relies on knowing what atom is the singular, most direct "parent" of any given virtual particle. If I'm reading the SMIRNOFF [spec](https://openforcefield.github.io/standards/standards/smirnoff/#virtualsites-virtual-sites-for-off-atom-charges) correctly, it looks like this atom is always the first (1-index in the spec, 0-index in any 0-based indexing scheme). But if I'm reading the source code of the toolkit, it's [not always](https://github.com/openforcefield/openff-toolkit/blob/921fbe2d48f3069adc3af6e919349dee6deefa26/openff/toolkit/typing/engines/smirnoff/parameters.py#L5810-L5816) the first atom, sometimes it's the second atom (or in an improper?).

j-wags:
Thanks for noting this. I think that fixing this may require more than just updating the exclusions - The root causes of https://github.com/openforcefield/openff-toolkit/issues/1221 and https://github.com/openforcefield/openff-toolkit/issues/1222 may be that that the Toolkit's implementation of Di- and TrivalentLonePair sites have been using the wrong indexing all along (relative to the spec). So while it would be great to fix the indexing problem altogether, it would be a ton of work and may actually be disruptive to the fitting team.

I'd advocate for either of the following courses of action:
* Match the toolkit's current behavior, even if it's wrong
* Match what the toolkit "currently thinks" the central atom is, regardless of whether it agrees with the spec

mattwthompson:
I'll admit I wasn't considering either possibility, and was hoping to do it "the right way." Though that may not be practical here, as we've discussed. I'll have to think about this some more.

mattwthompson:
Something's off with the application of charges:

In
```
toolkit_charges = list()

tmp = ff.create_openmm_system(mol.to_topology())
for force in tmp.getForces():
    if type(force) == openmm.NonbondedForce:
        for i in range(tmp.getNumParticles()):
            toolkit_charges.append(force.getParticleParameters(i)[0]._value)

interchange_charges = [
    v.m
    for v in Interchange.from_smirnoff(ff, [mol])[
        "Electrostatics"
    ].charges_with_virtual_sites.values()
]

[*zip(toolkit_charges, interchange_charges)]
```

Out:
```
[(-0.5461000055074692, -0.3827500082552433),
 (0.6320499926805496, 0.8180400155484677),
 (-0.4150100082159042, -0.5919000007212162),
 (0.08643001317977905, 0.0325099915266037),
 (0.5426300078630447, 0.4241000019013882),
 (-0.30000000000000004, -0.30000000000000004)]

mattwthompson:
Oops - I think the above is due to https://github.com/openforcefield/openff-toolkit/pull/1220 pending

mattwthompson:
Geometries are not accurate but this branch is the de facto default branch now and I'm merging to avoid branch divergence.

PR #432, opened 2022-04-05
Refresh examples

Yoshanuikabundi:
### Description
This PR refreshes the examples as discussed in #408.

So far I've just done the protein-ligand example. I've re-written it to have a more linear flow, use Interchange as much as possible, and reduce intermediate work with other packages. I've also updated bits like adding LAMMPS to the energy tests, added big red warnings when code isn't production ready, and formatted everything with headings and dependency installation instructions.

@mattwthompson Let me know which other examples need updating!


examples/protein_ligand/protein_ligand.ipynb
```diff
@@ -209,7 +213,11 @@
     "# This will take 1-2 minutes to load\n",
     "ff14sb = ForceField(\"ff14sb_off_impropers_0.0.2.offxml\")\n",
     "ff14sb.deregister_parameter_handler(\"ImproperTorsions\")\n",
```
mattwthompson, line 215:
```suggestion
```
This can be dropped with version 0.0.2 - something I forgot to propagate out. You had no real way of knowing this but it caught my eye here.

mattwthompson:
Changes look good to me on a skim through!


The examples I'd like some feedback/consideration on are:
* `packed_box.ipynb` seems to me now like it's a mixture of too basic and too focused on system preparation to be useful here. But on the other hand I can see that there's value in examples with these relatively simple use cases, showing _that_ they're simple to prepare in some cases.
* Two or three of `smirnoff_argon.ipynb`, `smirnoff_ethanol.ipynb`, and `SMIRNOFF_simulation.ipynb` should probably be removed; the latter is adapted from a toolkit example of the same name and is particularly focused on how to OpenMM things.  The others are each from an era of this project in which getting a single molecule out to a simulation engine was novel. They don't look so useful now. Of all these, I think the instructive value I want to remain is a sort of basic, bare minimum functionality example about how to go from a molecule(s) and force field and get something running in a user's engine of choice. I'd like to see one of these adapted to that purpose or, potentially, something new written if you think starting fresh is better.
* Please take a pass through `parameter_replacement.ipynb` and `parameter_splitting.ipynb`; I don't recall touching them in the past year but I think the functionality is important enough to warrant your attention. I'm reminded of [this ](https://openff-interchange.readthedocs.io/en/latest/using/handlers.html#inspecting-an-assigned-parameter)section of the docs you recently wrote - thanks for bringing that to light.
* No need to work on the Foyer or JAX examples now; those won't be a focus for the first batch of users.

I'd also be curious if you feel any functionality is missing from the docs (not that that would need to be resolved here). The current examples seem relatively comprehensive but I could certainly have a blind spot.

Sorry about so many failing tests - those are mine to fix - causing noise where we like signal. As long as the examples run they can be ignored here.

examples/protein_ligand/protein_ligand.ipynb
Yoshanuikabundi: I tried this during the rewrite and it caused an error during export, but retrying it now it seems to work. Must've just accumulated one too many inconsistencies in that notebook kernel! Thanks.

Yoshanuikabundi:
> packed_box.ipynb seems to me now like it's a mixture of too basic and too focused on system preparation to be useful here. But on the other hand I can see that there's value in examples with these relatively simple use cases, showing that they're simple to prepare in some cases.

I think this is an important example to have somewhere because that trick going through OpenMM for multi-component PDBs is probably pretty important. So I've refreshed and updated it a bit. Jeff and I talked about adding an API point somewhere that does this, and I think Interchange is the place for it because otherwise someone has to write code to take the positions and put them in `conformers[0]` for each molecule in a topology. In Interchange assigning this is much easier.

This example actually inspired me to move some stuff around `protein_ligand` so that each component is a bit more self contained. So between packed_box and protein_ligand the examples now cover 4 ways of setting up a system, as well as how to combine systems created in different ways. This seems good to me! `Protein_ligand` will get even better at this job when Rosemary comes out and we can combine molecules built in different ways with the same FF rather than combining interchanges. There's probably room for a "first principles" interchange construction example which goes from, like, SMILES string to gas phase interchange, but I think in practice won't this be fairly rare for users to do?

> Two or three of smirnoff_argon.ipynb, smirnoff_ethanol.ipynb, and SMIRNOFF_simulation.ipynb should probably be removed; the latter is adapted from a toolkit example of the same name and is particularly focused on how to OpenMM things. The others are each from an era of [this ](https://openff-interchange.readthedocs.io/en/latest/using/handlers.html#inspecting-an-assigned-parameter)project in which getting a single molecule out to a simulation engine was novel. They don't look so useful now. Of all these, I think the instructive value I want to remain is a sort of basic, bare minimum functionality example about how to go from a molecule(s) and force field and get something running in a user's engine of choice. I'd like to see one of these adapted to that purpose or, potentially, something new written if you think starting fresh is better.

I've written an example that takes the system from `packed_box` through to a simulation with OpenMM, and a second example for Amber, and I'll write one for LAMMPS soon and then GROMACS when it's ready. The idea is that they just take a fully speced Interchange and show how to run a simulation in each engine. So then we have `protein_ligand` and `packed_box` for system construction, and an example for each output engine.

I've used GROMACS and OpenMM before but I'm a noob at Amber and LAMMPS. My Amber example is broken, it seems to be heating/blowing up for no reason - I'd guess some energy minimization or a heat ramp would fix it but OpenMM didn't need it, do you have any ideas about what else could be going on?

> Please take a pass through parameter_replacement.ipynb and parameter_splitting.ipynb

 Will do!

> I'd also be curious if you feel any functionality is missing from the docs

Will keep an eye out but I only know about features that I find in the code or someone tells me about! The obvious thing would be going backwards, like amber parameters -> Interchange, but I don't think this is really ready for primetime yet?

Yoshanuikabundi:
Shout out to pre-commit CI for constantly saving my butt

Yoshanuikabundi:
How come `_write_lammps_input`, `_write_mdp_file`, `amber._write_input_file` etc are all private? It seems like they're important for making sure the force field is being accurately simulated. What's the plan there?

mattwthompson:
> How come `_write_lammps_input`, `_write_mdp_file`, amber.`_write_input_file` etc are all private? It seems like they're important for making sure the force field is being accurately simulated. What's the plan there?

Correct, they are crucial for ensuring force fields are accurately simulated. They should be made public at some point, but there is yet no plan for doing so. They're private now because we need them for internal use but can't really offer stability there.

> There's probably room for a "first principles" interchange construction example which goes from, like, SMILES string to gas phase interchange, but I think in practice won't this be fairly rare for users to do?

Do you mean something like this?

```
out = Interchange()
out.topology.add_molecule(Molecule.from_smiles(smiles))
bond_handler = SMIRNOFFBondHandler._from_toolkit(out.topology, force_field['Bonds'])
# repeat for each other handler
# etc.
```
because I'd agree that's a bit obtuse, at least in the current design. Going through toolkit objects to make a gas-phase single-molecule system is routine for the fitting team, though, but I think we have that covered well.

> I've used GROMACS and OpenMM before but I'm a noob at Amber and LAMMPS. My Amber example is broken, it seems to be heating/blowing up for no reason - I'd guess some energy minimization or a heat ramp would fix it but OpenMM didn't need it, do you have any ideas about what else could be going on?

I'm no Amber expert but I can have a look at this. Feel free to split this file off into a separate PR so as to not block this one. I grabbed [this](https://github.com/openforcefield/openff-interchange/blob/7f045194cff7b6dcf37868b9949c972f7583cc2a/examples/amber/amber.ipynb) file and ran it blindly; while it did not error out, the last frame does look suspicious.

> Will keep an eye out but I only know about features that I find in the code or someone tells me about!

Fair point! Right now user quests tend to be more exotic than we can handle (or may ever handle) like a sort of interoperability decathlon of different tools scattered across OpenFF/Amber/OpenMM ecosystems. I'll keep my ear to the ground and convey requests that I think would be a good fit to tackle.

>  The obvious thing would be going backwards, like amber parameters -> Interchange, but I don't think this is really ready for primetime yet?

Correct - though this is likely the first major feature after the 0.11.0/0.2.0 releases.

Are there other major changes you'd like to tackle in this PR? I think my feedback so far has pretty much consisted of nodding my head and confirming your questions, but I haven't gone through line-by-line to address anything unresolved you may (or may not) have come across there.

examples/parameter_replacement/parameter_replacement.ipynb
```diff
@@ -0,0 +1,316 @@
...
+    "[`store_potentials`]: https://openff-interchange.readthedocs.io/en/stable/_autosummary/openff.interchange.components.potentials.PotentialHandler.html#openff.interchange.components.potentials.PotentialHandler.store_potentials"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "pars_sys.handlers[\"Angles\"].store_matches(sage_ff[\"Angles\"], topology=pars_sys.topology)"
```
Yoshanuikabundi, line 245: This line was previously:

```python
sys_100.handlers[\"Angles\"].store_matches(parsley_100["Angles"], topology=top)
```

I changed it to

```python
sys_100.handlers[\"Angles\"].store_matches(parsley_130["Angles"], topology=sys_100.topology)
```

because I think it was an error, `store_matches` takes the new parameter handler as input, not the old one. Can you confirm this was an error?

examples/parameter_replacement/parameter_replacement.ipynb
```diff
@@ -0,0 +1,316 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "pars_sys.handlers[\"Angles\"].slot_map.clear()\n",
+    "pars_sys.handlers[\"Angles\"].potentials.clear()"
```
Yoshanuikabundi, line 226: I thought it would be good to show users how to clear old parameter handlers, in case someone tries to add a parameter handler with different SMIRKS covering the same chemistry. This is what I came up with, is there a better way?

Yoshanuikabundi:
Finished and ready for review!

mattwthompson (approved):
Nice work! This is just the sort of changeset I had in mind, and reading through the protein-ligand example has me excited for a near future in which the process is smoother.

I used a third-party service (https://app.reviewnb.com/openforcefield/openff-interchange/pull/432/) to eyeball the diffs. I think it requires logging in and might have other features, all I did was use it to make diffs more readable (you'd think GitHub would have something like this ... maybe it's in the hopper). I didn't run any locally since CI is passing and there aren't many plausible errors that wouldn't cause an exception or exit code. Everything I saw was an improvement over the existing code and I couldn't find anything obvious areas of improvement.

> because I think it was an error, store_matches takes the new parameter handler as input, not the old one. Can you confirm this was an error?

Yes, that was an error, although the result would happen to be the same between those versions.

> I thought it would be good to show users how to clear old potential handlers, in case someone tries to add a parameter handler with different SMIRKS covering the same chemistry. This is what I came up with, is there a better way or is there anything else that should be cleared?

I don't think there's a distinctly better way to do it and haven't thought much about how this should be handled at the API level. Just clearing those two dicts should be enough here, I think.

Non-actionable feedback:
* Nice find using the Python interface to LAMMPS! I hadn't even considered using it (I think it was less mature when I was using LAMMPS in grad school) and it's installed with the conda package!
* Also +1 for the `.gitignore` files in subdirectories
* Also +1 for the streamlined file structure
* The use of `get_data_file_path("systems/packmol_boxes/propane_methane_butanol_0.2_0.3_0.5.pdb")` is not desirable, in my opinion, and I only used those PDB files out of convince. At some point (not now) we should move to use files stored here.
* I notice a lot of import blocks are not yet taking advantage of your `from openff.toolkit import Molecule, ForceField` shortcut. That doesn't need to be fixed now, but I hope over time we'll make that shift. I remember bit import blocks being somewhat daunting when I was just starting out, and I think that change is a nice plus.
* "The `Molecule.from_pdb()` method will soon be renamed `Molecule.from_polymer_pdb()` to better describe its purpose." Interesting, I was unaware of this coming change.

mattwthompson:
I read through our back-and-forth and I don't see any loose threads you were planning to clean up - feel free to merge whenever you wish.

PR #440, opened 2022-04-13
New (broken) Amber example

Yoshanuikabundi:
### Description
This is the Amber example that is currently broken. I've broken it into its own PR as suggested to avoid blocking the other examples PR.

I've never used Amber before. OpenMM and LAMMPS both simulate identical positions and parameters with no trouble, but Amber blows up. It doesn't crash because the velocities are clamped at each step, but the simulation heats very quickly. Possibly unrelatedly, the [protein_ligand](https://github.com/openforcefield/openff-interchange/blob/8f04ecdf6545b39de428659c7593774acfbb6fc6/examples/protein_ligand/protein_ligand.ipynb) example demonstrates that Amber's electrostatic potential energy is about 4 times that of LAMMPS or OpenMM.

Previous discussion in #408 and #432


j-wags:
Ok, @Yoshanuikabundi and I looked at this a lot today, and were unable to determine what's going wrong here. 

The changes we tried which were successful by any measure were:

* Cranking up the `gamma_ln` to >100 could keep the simulation from exploding

The changes we tried which were not successful by any measure were:

* Truncating atom names and atom type names to a max of 4 characters (we noticed that they overflowed their fields)
* Setting the initial and/or final temperature to different values, including 0
* Cranking up the box size by a factor of 10
* Changing the fswitch to 8 and the cutoff to 9
*  Changing ntc to 2 and 3 (hbond constraints, all crashed due to large deviation from constraints)

Next steps would be to try:
* Minimizing to see if it ever stabilizes
* reducing the number of atoms to keep any 4-character name/type fields from overflowing (or implementing logic to know whern there are identical atoms)
* Seeing whether the "explodey" molecules have their atoms in different orders from each other (maybe bonds in later isntances of a molecule are getting mangled by a reference vs. topology order thing?)

mattwthompson:
Moving this into #1046

PR #443, opened 2022-04-21
Mdconfig docs

Yoshanuikabundi:
### Description
This PR adds book-style documentation to the new `MDConfig` object and adds it to the LAMMPS example. The Amber example is still in PR and there is no GROMACS example until performance improves, so other uses of `MDConfig` are not yet in examples.

### Questions

- The LAMMPS example crashes with `special_bonds lj 0 0.5 1 coul 0 0.8333333333 1`, as outputted by MDConfig. The previous value of `special_bonds lj 0.0 0.0 0.5 coul 0.0 0.0 0.8333333333`, which IIRC was previously produced by `_write_lammps_input_file()`, works with the other changes to the sample file. I think this is an error/regression in MDConfig? I tried changing the 0s and 1s to 0.0s and 1.0s in case it was an int/float error but no dice. I've left the new value in the example as a reminder not to merge until we sort this out.
- Have we considered something like `Interchange.create_mdconfig()` instead of `MDConfig.from_interchange(...)`? This would save users an import and be less verbose
- Would it be possible to take a file object instead of a file name in the `MDConfig` export methods? Users will immediately want to either manually read or otherwise edit these files nearly every time they create them

### DIFF
--- a/docs/using/output.md
+++ b/docs/using/output.md
@@ -7,7 +7,7 @@ source of truth is the rich chemical information in the `Interchange`
 object, and exported files are tools to perform some operation.
 
 (sec-mdconfig)=
-### Run control/config files
+## Run control/config files
 
 SMIRNOFF force fields include several parameters that many MD engines do not
 include as part of their topologies. These values are essential for accurately
--- a/docs/using/output.md
+++ b/docs/using/output.md
@@ -6,6 +6,32 @@ stored in an `Interchange`; they support a design where the principle
 source of truth is the rich chemical information in the `Interchange`
 object, and exported files are tools to perform some operation.
 
+(sec-mdconfig)=
+### Run control/config files
+
+SMIRNOFF force fields include several parameters that many MD engines do not
+include as part of their topologies. These values are essential for accurately
+simulating output from Interchange, but they are configured in the same files
+that are used for general control of simulation runtime behavior. As a result,
+Interchange cannot simply provide complete versions of these files.
+
+Instead, Interchange provides [`MDConfig`], a class that writes stub versions of
+MD engine run input files. These files must be modified and completed before
+they can be used to run a simulation.
+
+`MDConfig` can be constructed from an existing Interchange:
+
+```python
+from openff.interchange import Interchange
+from openff.interchange.components.mdconfig import MDConfig
+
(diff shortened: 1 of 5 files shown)


mattwthompson:
Thanks for opening this; I can see now that I should have been a bit more careful before merging an experiment into the main branch.

This deserves a bit more care than I can give this week, but my quick thoughts are
1. It sure looks like you've correctly identified a regression.
2. I hadn't thought too much about this, the current design is what felt natural when I was writing it up the first time. I can see why `Interchange.create_mdconfig()` is smoother from the user perspective. I'd like to hold off on deciding this until a little bit more work is done on the examples (I can tackle this in the near future), using information gathered from that work to inform a decision.
3. I think this sounds good, would it mean the methods can't take strings anymore? I haven't worked with file objects much so I'm not so familiar. I'd happily take a PR doing this but I wouldn't rate it as more important than other work that's probably on your plate at the moment.

mattwthompson (approved):
Okay, I looked through the actual changes to docs and they look great. I'm going to try to fix the bug you found and with any luck I won't break other things along the way.

Yoshanuikabundi:
>> Would it be possible to take a file object instead of a file name in the MDConfig export methods?
>
> I think this sounds good, would it mean the methods can't take strings anymore? I haven't worked with file objects much so I'm not so familiar. I'd happily take a PR doing this but I wouldn't rate it as more important than other work that's probably on your plate at the moment.

I think you could go either way. Not accepting strings simplifies implementation but adds boilerplate when you actually want to write to disk. You could write the function polymorphically something like: 

```python
from pathlib import Path
from typing import TextIO
...
class ...
    def _export_whatever_inner(self, outfile: TextIO):
        # Code to run on open file object goes here

    def export_whatever(self, outfile: Union[Path, str, TextIO]):
        try:
            outfile = open(outfile, 'r')
        except TypeError:
            self._export_whatever_inner(outfile)
        else:
            with outfile:
                self._export_whatever_inner(outfile)
```

This pattern is verbose but it ensures that (1) the output file is closed by export_whatever if and only if it was opened by export_whatever; (2) `TypeError`s from the inner function don't get caught by the outer try block; (3) minimal code is duplicated; (4) end users can do whatever weird duck typing hackery they want.

PR #445, opened 2022-04-28
Add solvated ligand example

mattwthompson:
### Description
I took a few wrong turns and ended up on an old-but-still-relevant example that never quite got polished up (https://github.com/openforcefield/openff-toolkit/issues/844). I think it would have a good home here since putting a ligand in a box of water in something people have a general interest in doing. Arguably it duplicates some effort of `examples/packed_box`.

### DIFF
--- a/examples/ligand_in_water/ligand_in_water.ipynb
+++ b/examples/ligand_in_water/ligand_in_water.ipynb
@@ -1,0 +1,279 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "d051f4a7-9a72-418b-b068-8d2e330a8db6",
+   "metadata": {},
+   "source": [
+    "# Solvating and equilibrating a ligand in a box of water\n",
+    "\n",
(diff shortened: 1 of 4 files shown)


examples/ligand_in_water/ligand_in_water.ipynb
```diff
@@ -0,0 +1,279 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "54b39faf-eae9-4550-92bc-68794432ec59",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "if False:\n",
+    "    interchange = Interchange.from_smirnoff(force_field=sage, topology=topology)\n",
+    "    system = interchange.to_openmm(combine_nonbonded_forces=True)\n",
```
mattwthompson, line 152: Something is broken in `MDConfig` parsing constraints, whether I use the constrained or unconstrained version of Sage it chokes on something.

PR #447, opened 2022-04-29
Fix intramolecular scale factors in `write_lammps_input`

mattwthompson:
### Description
In #442 I broke the LAMMPS input/run file writer by mangling the intramolecular scale factors. SMIRNOFF handlers keep track of 1-3, 1-4, and 1-5, but LAMMPS's `special_bonds` command cares about 1-2, 1-3, and 1-4. Ctrl+F on the [diff](https://github.com/openforcefield/openff-interchange/pull/442/files) makes it more obvious. In an attempt to prevent future developers from being as confused as I was, I replaced the lists with some extremely verbose and explicit dicts.

Thanks to @Yoshanuikabundi for isolating this bug.

### Checklist
- [ ] Add tests
- [x] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/components/mdconfig.py
+++ b/openff/interchange/components/mdconfig.py
@@ -157,12 +157,29 @@ class MDConfig(DefaultModel):
             lmp.write("improper_style cvff\n")
 
             # TODO: LAMMPS puts this information in the "run" file. Should it live in MDConfig or not?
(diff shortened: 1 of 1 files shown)


mattwthompson:
Merged as a part of #443

PR #452, opened 2022-05-12
Assorted changes for toolkit parity

mattwthompson:


openff/interchange/components/smirnoff.py
```diff
@@ -1521,17 +1584,55 @@ def store_matches(
...
+                molecule_index = molecule.atom_index(atom)
+                topology_index = topology.atom_index(atom)
+
+                molecule_charges[molecule_index] = topology_charges[topology_index]
+
+            charge_sum = sum(molecule_charges)
+            formal_sum = molecule.total_charge.m
 
-                        self.slot_map[topology_key] = potential_key
+            if abs(charge_sum - formal_sum) > 0.01:
```
mattwthompson, line 1622: To help our memory, this value was selected to match long-standing behavior in the toolkit:

https://github.com/openforcefield/openff-toolkit/blob/fade767977cda3c2d70399ac38644aa5428414fe/openff/toolkit/typing/engines/smirnoff/parameters.py#L3994-L3997

PR #481, opened 2022-07-06
Fix code coverage

mattwthompson:
I don't know how many times I'll have to nurse CodeCov into doing what I want, but here goes another attempt:
- [x] Fix reporting paths (`/*/*/interoperability_tests`, `/*/*/energy_tests/`, `/*/*/unit_tests/`) that do not exist in the project
- [x] Try to actually ignore `if TYPE_CHECKING:` lines as specified in the setup

### DIFF
--- a/.github/workflows/ci_minimal.yaml
+++ b/.github/workflows/ci_minimal.yaml
@@ -24,8 +24,7 @@ jobs:
           - "3.8"
           - "3.9"
     env:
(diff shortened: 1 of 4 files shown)


mattwthompson:
Okay, the codecov web app didn't pick up on the change in the default branch 🤷 🤷‍♂️ 🤷‍♀️ had to fix that myself

PR #491, opened 2022-08-02
Fix nightly and beta testing workflows

mattwthompson:


mattwthompson:
Something is different with how OpenMM reports energies on Linux only.

PR #492, opened 2022-08-02
Use micromamba in CI

mattwthompson:
It's supposed to be faster and appears to be close to a drop-in replacement


mattwthompson:
This gives us a free 2-3 minutes in most CI jobs, or about half the time `setup-miniconda` takes

PR #494, opened 2022-08-04
Faster parameter lookup

j-wags:
### Description
Uses ParameterList slicing instead of `get_parameter` to look up parameters by their SMIRKS. Cuts `ff.create_interchange` runtime by about 40%. 

Test code:

```python
%load_ext snakeviz

from openff.toolkit import Molecule, OpenEyeToolkitWrapper, RDKitToolkitWrapper, ForceField, Topology
from openff.toolkit.utils import get_data_file_path

ff = ForceField('openff-2.0.0.offxml', 'ff14sb_off_impropers_0.0.3.offxml')

protein = Molecule.from_polymer_pdb(get_data_file_path('proteins/T4-protein.pdb'))
ala = Molecule.from_polymer_pdb(get_data_file_path('proteins/MainChain_ALA.pdb'))
top = Topology()
top.add_molecule(protein)
for i in range(5):
    top.add_molecule(ala)
water = Molecule.from_smiles('O')
for i in range(50):
    top.add_molecule(water)
```

```python
%%snakeviz
interchange = ff.create_interchange(top)
```

Before these changes:

Screen Shot 2022-08-04 at 3 29 02 PM

After these changes:

Screen Shot 2022-08-04 at 3 34 21 PM

### DIFF
--- a/openff/interchange/components/smirnoff.py
+++ b/openff/interchange/components/smirnoff.py
@@ -285,7 +285,6 @@ class SMIRNOFFBondHandler(SMIRNOFFPotentialHandler):
             self.potentials = dict()
         for topology_key, potential_key in self.slot_map.items():
             smirks = potential_key.id
-            #parameter = parameter_handler.get_parameter({"smirks": smirks})[0]
             parameter = parameter_handler.parameters[smirks]
             if topology_key.bond_order:
                 bond_order = topology_key.bond_order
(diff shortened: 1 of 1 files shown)


mattwthompson (approved):
Nice catch ... I think I was completely unaware of the performance gains here (even though I might have been the one to implement this? https://github.com/openforcefield/openff-toolkit/commit/6bd1e133907094d7d4c798efb7fc59bb290e207f)

These changes seem atomic enough and pass all tests, I can't think of anything else to consider.

PR #500, opened 2022-08-10
Defer JAX imports until used

mattwthompson:


mattwthompson:
I'm not sure why this doesn't work, but it doesn't.

PR #502, opened 2022-08-10
Collate virtual particles in OpenMM exports

mattwthompson:
Based on https://github.com/openmm/openmm/issues/1931 and a few dozen hours of tinkering, we have come to the conclusion that we cannot place virtual particles at the end but must collate them within the topology (and by extension the positions vector and force objects). At the moment, every virtual particle in a molecule is associated with the last residue in the molecule; this is fine for most systems but this assumption will fail when polymers are involved.


mattwthompson:
I still need to update the `System` creation, which will take more than a few minutes because of my decision a while ago to add atoms and virtual particles in separate functions.

mattwthompson:
Pending (though not necessarily blocked by) resolution of https://github.com/openmm/openmm/issues/3736

PR #503, opened 2022-08-16
Update virtual sites in OpenMM export

mattwthompson:
new boss same as the old boss


mattwthompson:
Will fix the examples later

PR #504, opened 2022-08-16
Pin `jax`/`jaxlib` to avoid 0.3.16

mattwthompson:


mattwthompson:
I have no idea how JAX keeps getting pulled in ... I even went as far as running

```
$ CONDA_SUBDIR=linux-64 mamba env create --file devtools/conda-envs/test_env.yaml
```

on my mac to see if it gets pulled in (it does not)

mattwthompson:
Should be fine for a bit while grabbing 0.3.14 and probably upstream will flesh things out in time: https://github.com/conda-forge/jaxlib-feedstock/issues/139#issuecomment-1219844683

Issue #507, opened 2022-08-19
Integrate external regression tests into CI suite


mattwthompson:
Basically I have some bad implementations at what https://github.com/openforcefield/interchange-regression-testing sets out to do. It would be useful to have a few tens of of molecules from that data set split out into here so that potential regressions are caught before the release process.

PR #508, opened 2022-08-19
Update regression/energy tests

mattwthompson:
Closes https://github.com/openforcefield/openff-interchange/issues/507 and helps https://github.com/openforcefield/openff-toolkit/issues/1051

- [ ] Remove older tests that assumed on `ForceField.create_openmm_system(topology)` may produce a different result than `Interchange.from_smirnoff(force_field, topology).to_openmm()`
- [ ] Scrub from the annals of history other questionable tests I've written in the past
- [ ] Write out a few XMLs with version 0.10.6 of the toolkit
- [ ] Add new tests that compare differences between the above files and newly-generated ones


mattwthompson:
The regression tests should probably be their own action, but I'll do that another time. I've already spent too much time on this.

PR #510, opened 2022-08-23
Port new `ensure_unique_atom_names` behavior from Toolkit

Yoshanuikabundi:
### Description
Once [#1368](https://github.com/openforcefield/openff-toolkit/pull/1368) is merged into the Toolkit, the `ensure_unique_atom_names` argument of `Topology.to_openmm()` will accept the name of a hierarchy iterator and ensure atom names are unique within elements of that iterator, rather than within molecules. This helps preserve atom names of proteins. This PR ports the same behavior to Interchange.

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


Yoshanuikabundi:
Hmm these test failures are very cryptic. This PR does depend on some new features in the Toolkit PR, so maybe we should check again after that's merged?

mattwthompson:
Up to you - we could wait until they're merged upstream or point this PR to that feature branch

Yoshanuikabundi:
Looks like I was misreading the longest-time test report as failiing tests! Test failures are from my copying the input topology to prevent new atom names from changing the input (matching the new Toolkit behavior). This was a problem because the `_virtual_site_parent_molecule_mapping()` function returned a map to actual `Molecule` objects, which are copied with the topology and therefore cannot be found in the duplicate topology. 

Every use of the function could be very easily refactored to use molecule indices instead, so I did that. In most cases, this basically amounts to moving a use of `Topology.molecule_index` from the calling function into `_virtual_site_parent_molecule_mapping()`. In `to_gro`, the molecule is used directly in a dict, so this use of `Topology.molecule_index` is basically wasted - but I think this is small performance potatoes at the moment.

I'm happy to remove the topology copy and revert these latter changes if you prefer!

I'm also curious - would it be possible to just call `Topology.to_openmm()`, and then add the virtual sites in on top of that? This might reduce code duplication between Interchange and the Toolkit.

openff/interchange/interop/openmm/_topology.py
```diff
@@ -9,42 +9,50 @@
...
     from openff.toolkit.topology.molecule import Bond
     from openmm import app
 
     from openff.interchange.interop._virtual_sites import (
         _virtual_site_parent_molecule_mapping,
     )
 
-    topology = interchange.topology
+    # Copy topology to avoid modifying input (eg, when generating atom names)
+    topology = Topology(interchange.topology)
```
mattwthompson, line 30: In a perfect world I'd like to avoid copying large objects [^1] but I can see the value here. IIUC the only modifications happening here are tinkering with the atom names, but 
* I started thinking about what it would take to implement a mapping between the original and "modified as requested" atom names and could tell I was too lazy to see that idea to completion 
* It stands to reason there are more modifications that we'll be involved in in the future ...


[^1]: I have not done the profiling to understand what's really taking up space in memory - but I figure a `Topology` of a large protein in water will be large using our current object models

mattwthompson (commented):
Thanks for the contribution; I didn't follow the discussion in the toolkit PR to the end but I trust you and Jeff arrived at a good design, and thanks for doing the work of porting it over here. The changes look good to me as I read them and tests are passing, but I would like https://github.com/openforcefield/openff-interchange/pull/512 merged in before approving this as there's enough changes to possibly overlap in bad ways. It's also not required, and I'll leave the decision to you, but if there are any tests that should be ported from `openff/toolkit/tests/test_topology.py` now would be a good time to get that started.

> Every use of the function could be very easily refactored to use molecule indices instead

I think in principle there shouldn't be any difference between using molecules or their indices, but IIRC there were unworkable performance issues with `Topology.molecule_index` when I was first working on that function. There's still an overhead associated with looking up molecules by their indices (compared to just _having_ them there), although it's not as bad now (nor significant compared to `Topology` construction):

```python
In [1]: from openff.toolkit import *

In [2]: water = Molecule.from_mapped_smiles("[H:2][O:1][H:3]")

In [3]: %%timeit
   ...:
   ...: topology = Topology.from_molecules(10000 * [water])
   ...:
   ...:
22.8 s ± 1.18 s per loop (mean ± std. dev. of 7 runs, 1 loop each)


In [5]: topology = Topology.from_molecules(10000 * [water])

In [6]: %%timeit
   ...:
   ...: [topology.molecule_index(molecule) for molecule in topology.molecules]
   ...:
   ...:
3.28 s ± 83.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

I don't think there's currently much reason to care about this difference now, though.

> I'm also curious - would it be possible to just ...

Probably. My memory is hazy but I recall getting confused about how OpenMM allows or doesn't allow you to add particles to residues/chains after they're created, if there are restrictions or gotchas, etc. This soured my mood enough that the fundamentally bad decision to port the code directly looked better. (At the time, I was wishy-washy on whether virtual sites go at the end of the OpenMM objects or if they should be collated through - that added complexity and indecision made everything harder. Now, they're collated, and while it's trickier to implement and maintain that is my preference moving forward.) I don't feel strongly about making this change now (I think the next set of releases will necessarily see huge changes to the `Topology` class) - but if you do and have the bandwidth I'd be happy to take a PR.

Yoshanuikabundi:
> Probably. My memory is hazy but I recall getting confused about how OpenMM allows or doesn't allow you to add particles to residues/chains after they're created, if there are restrictions or gotchas, etc. This soured my mood enough that the fundamentally bad decision to port the code directly looked better. (At the time, I was wishy-washy on whether virtual sites go at the end of the OpenMM objects or if they should be collated through - that added complexity and indecision made everything harder. Now, they're collated, and while it's trickier to implement and maintain that is my preference moving forward.) 

Ahhh that makes sense. Maybe one day it'll be better to leave this functionality to Interchange and have the toolkit use it with some sort of dummy force field.

> but if you do and have the bandwidth I'd be happy to take a PR.

lol nope 😬

mattwthompson (approved):
I didn't look through the tests as I trust you ported them accurately - feel free to merge when you're happy with this (depending on your work schedule and how the rest of the release progresses, I may jump the gun and do it myself)

PR #513, opened 2022-08-24
Update ligand in water notebook

mattwthompson:
### Description

This PR updates the ligand-in-water notebook after we figured out the reasons for the simulations crashing (https://github.com/openmm/openmm/issues/3736) and virtual particle ordering.

@Yoshanuikabundi do you think it would be worth including short GROMACS simulations in this notebook? One one hand, it's a common user need (https://github.com/openforcefield/openff-toolkit/issues/1328) and not _that_ much of a leap from running the same state in OpenMM. On the other hand, it takes a significant amount of cruft to run GROMACS simulations in notebooks and the scope boundary here is hard to define - should the same be done when we support many engines?

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/examples/ligand_in_water/ligand_in_water.ipynb
+++ b/examples/ligand_in_water/ligand_in_water.ipynb
@@ -36,8 +36,7 @@
     "from openff.units import unit\n",
     "from openff.units.openmm import from_openmm, to_openmm\n",
     "\n",
-    "from openff.interchange import Interchange\n",
-    "from openff.interchange.interop.openmm import to_openmm_positions"
+    "from openff.interchange import Interchange"
    ]
   },
   {
@@ -231,11 +230,15 @@
     "    integrator = openmm.LangevinIntegrator(\n",
     "        300 * openmm.unit.kelvin,\n",
     "        1 / openmm.unit.picosecond,\n",
-    "        0.1 * openmm.unit.femtoseconds,\n",
+    "        1 * openmm.unit.femtoseconds,\n",
     "    )\n",
     "\n",
     "    simulation = openmm.app.Simulation(topology, system, integrator)\n",
     "    simulation.context.setPositions(positions)\n",
+    "\n",
+    "    # https://github.com/openmm/openmm/issues/3736#issuecomment-1217250635\n",
+    "    simulation.minimizeEnergy()\n",
+    "\n",
     "    simulation.context.setVelocitiesToTemperature(300 * openmm.unit.kelvin)\n",
     "    simulation.context.computeVirtualSites()\n",
     "\n",
@@ -302,7 +305,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "run_simulation(simulation, n_steps=2000)"
+    "run_simulation(simulation, n_steps=5000)"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


Yoshanuikabundi:
> should the same be done when we support many engines?

My gut says that from the perspective of someone using Interchange, the engine shouldn't matter much - and most users will already know how to use their engine of choice. It's just a question of picking the right API points to export the stuff you need, and this is already covered by the individual engine examples and the docs. Examples are hard enough to maintain as it is; we should avoid the combinatorial explosion of matching all examples to all engines. OpenMM is the obvious choice for examples; not only is it the best supported by Interchange, it's also the best supporter of Jupyter.

I guess we should try to make the individual engine examples hit as many features as possible though.

mattwthompson:
Okay, thanks for the thoughts. I'll leave GROMACS out of this example for now, but think more about how to include it in testing and provide some way to help communicate the functionality out to users.

examples/ligand_in_water/ligand_in_water.ipynb
```diff
@@ -36,8 +36,7 @@
     "from openff.units import unit\n",
```
Yoshanuikabundi, line 36: <p>Obviously I'm biased, but I'd suggest doing this at the same time as the box vectors (which are kinda from one perspective a bit useless without positions anyway) and using the new <code>Topology.set_positions()</code> method. This also reduces the duplicated work in running TIP4P.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openforcefield/openff-interchange/pull/513/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

Yoshanuikabundi, line 36: <p>An alternative to having this in an appendix would be to run the whole notebook with two parallel systems. This would emphasize how easy it is to change the force field; instead of seeing a whole appendix with new code, there would just be a few doubled up lines.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openforcefield/openff-interchange/pull/513/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

Yoshanuikabundi, line 36: <pre class="ql-syntax" spellcheck="false">Line #1.    from openff.interchange.interop.openmm import to_openmm_positions
</pre><p> </p><p>I think we should encourage users to import from public modules.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openforcefield/openff-interchange/pull/513/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='28'/>

Yoshanuikabundi:
We could have a blurb in the examples like, "This notebook demonstrates simulation with OpenMM, but it works just as well with ..."?

examples/ligand_in_water/ligand_in_water.ipynb
mattwthompson: <p>Oops yep, good catch - forgot to update this one in the other PR.</p>

mattwthompson: <p>I really like this idea!</p>

mattwthompson:
The problem is I'm not quite sure it works just as well with GROMACS, and committing some time to expanding this notebook is synonymous with getting it to actually work for more than zero steps, produce a reasonable-looking trajectory, etc. For now I'm pushing that work into the next release..

PR #514, opened 2022-08-24
Re-organize `openff.interchange.interop.openmm` API

mattwthompson:
### Description
While working on #513 I noticed that `from openff.interchange.interop.openmm._topology import to_openmm_topology` is ... not exactly peak user experience. Here I reorganized `openmm/__init__.py` to include some of the functions I'd expect most users to need to access.

I also added a strong warning in `from_openmm` and hid some imports away in helper functions since I don't think those will be used much soon.

cc: @Yoshanuikabundi - another ping! Hooray! I think this is what you have in mind for `__all__` usage?
### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/interop/openmm/__init__.py
+++ b/openff/interchange/interop/openmm/__init__.py
@@ -1,25 +1,21 @@
 """Interfaces with OpenMM."""
 from pathlib import Path
-from typing import Union
+from typing import TYPE_CHECKING, Union
 
 import openmm
-from openff.toolkit.topology import Topology
-from openff.units import unit as off_unit
-from openff.units.openmm import from_openmm as from_openmm_quantity
 from openmm import unit
 
-from openff.interchange.components.potentials import Potential
-from openff.interchange.constants import _PME
 from openff.interchange.exceptions import UnsupportedImportError
-from openff.interchange.interop.openmm._nonbonded import _process_nonbonded_forces
-from openff.interchange.interop.openmm._valence import (
-    _process_angle_forces,
-    _process_bond_forces,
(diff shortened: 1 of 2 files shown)


openff/interchange/interop/openmm/__init__.py
```diff
@@ -1,25 +1,21 @@
...
-    _process_angle_forces,
-    _process_bond_forces,
-    _process_constraints,
-    _process_improper_torsion_forces,
-    _process_torsion_forces,
-)
-from openff.interchange.models import PotentialKey, TopologyKey
+
+__all__ = [
+    "to_openmm",
```
mattwthompson, line 11: I'm open to renaming this `to_openmm_system` or something

mattwthompson: cc: @Yoshanuikabundi

Yoshanuikabundi: I think that's probably a better name but it doesn't matter much.

Yoshanuikabundi:
> cc: @Yoshanuikabundi - another ping! Hooray! I think this is what you have in mind for __all__ usage?

Yes! Perfect! Though `get_partial_charges_from_openmm_system` and `get_nonbonded_force_from_openmm_system` have been left out - was that deliberate?

mattwthompson:
Not necessarily; I think they could each be private but they could also each go away at some point. I don't consider them hidden nor a part of the public API. I'll probably revisit in the future and remove them if they're not used outside of testing.

PR #515, opened 2022-08-25
Defer pandas imports

mattwthompson:
### Description
Unfortunately the wins won't happen until https://github.com/hgrecco/pint/pull/1492 is finished.

```python3
>>> import sys, rdkit, openff.toolkit
>>> "pandas" in sys.modules
False
>>> import openff.units
>>> "pandas" in sys.modules
True
```

### Checklist
- [ ] Add tests
- [x] Lint
- [ ] Update docstrings


mattwthompson:
Blocked by https://github.com/hgrecco/pint/pull/1492

mattwthompson:
This should no longer be needed after Pint 0.21, but will need to test to be sure

PR #562, opened 2022-11-29
Do not call `setup.py` directly

mattwthompson:
See i.e. https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html


mattwthompson:
Cool, this drops a few deprecation warnings in logs.

PR #566, opened 2022-12-06
Refactor `EnergyReport`

mattwthompson:
### Description
`EnergyReport` was developed over time based on specific use cases and had some sketchy behavior in some corner cases that were not directly tested. This rewrites most of it from the ground up and adds unit tests for each of these cases, separate these unit tests out from the energy tests of the various drivers and writers.

### API changes
* `EnergyReport.update_energies` is now `EnergyReport.update`
* `EnergyReport.diff` is added

### Behavior changes
* `EnergyReport` now only allows the following keys in  the dict comprising its `.energies` attribute:
  *  `"Bond"`
  *  `"Angle"`
  *  `"Torsion"`
  *  `"vdW"`
  *  `"Electrostatics"`
  *  `"Nonbonded"`
* `EnergyReport.compare` and `EnergyReport.diff` now assume the two objects either both split out non-bonded terms or both do not.

### Checklist
- [x] Add tests
- [x] Lint
- [x] Update docstrings


mattwthompson:
Debugging code

```
import openmm

def f(system):
    for force in system.getForces():
        if isinstance(force, openmm.NonbondedForce):
            print(
                force.getUseSwitchingFunction(),
                force.getNonbondedMethod(),
                force.getCutoffDistance(),
                force.getSwitchingDistance(),
                force.getUseDispersionCorrection(),
                force.getNumExceptions(),
                force.getEwaldErrorTolerance(),
                force.usesPeriodicBoundaryConditions(),
            )
            # print([force.getParticleParameters(i)[0]._value for i in range(force.getNumParticles())])

        return force
f(system), f(interchange.to_openmm(combine_nonbonded_forces=True))

PR #591, opened 2023-02-21
Support double exponential force fields

mattwthompson:
### Description

### Checklist
- [ ] Add tests
  - [x] Test against reference implementation
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -38,3 +38,5 @@ dependencies:
   - types-setuptools
   - types-pkg_resources
   - pandas-stubs >=1.2.0.56
+  - pip:
+    - git+https://github.com/jthorton/de-forcefields
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -26,6 +26,7 @@ Please note that all releases prior to a version 1.0.0 are considered pre-releas
 
 ### New features
 * #589 For convenience, per-parameter variables are now stored on `SMIRNOFFCollection.potential_parameters` and its subclasses.
+* #591 Adds support for custom `SMIRNOFFCollections` via a plugin interface.
 
 ## 0.2.3 - 2022-11-21
 
--- a/openff/interchange/components/potentials.py
+++ b/openff/interchange/components/potentials.py
@@ -103,6 +103,10 @@ class Collection(DefaultModel):
     """Base class for storing parametrized force field data."""
 
     type: str = Field(..., description="The type of potentials this handler stores.")
+    is_plugin: bool = Field(
+        False,
+        description="Whether this collection is defined as a plugin.",
+    )
     expression: str = Field(
         ...,
         description="The analytical expression governing the potentials in this handler.",
--- a/openff/interchange/exceptions.py
+++ b/openff/interchange/exceptions.py
@@ -98,7 +98,7 @@ class UnsupportedCombinationError(BaseException):
     """General exception for something going wrong in Interchange object combination."""
 
 
-class PluginIncompatibilityError(BaseException):
+class PluginCompatibilityError(BaseException):
     """A plugin is incompatible with the current version of Interchange in the way it is called."""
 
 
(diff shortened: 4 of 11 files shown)


jthorton:
Hi @mattwthompson is the plan to directly support this in interchange or should this be moved to `smirnoff-plugins`? I am keen to get plugins working with the toolkit again and would love to update `smirnoff-plugins` if there is a guide on what needs to be implemented!

mattwthompson:
I don't have a strong preference; the plugins themselves are fairly lightweight so it doesn't make much of a difference where they're packaged. I need them here at least for testing purposes, it would be too slow to need to make updates in two places while I'm testing all of this out. I need something for long-term testing so I might keep something in the `plugins/` directory here.

mattwthompson:
I used this script to get the energy of a single ethanol molecule in vacuum. It's comprehensive test but it gets past a number of barriers (creating a `System` and `Simulation` with most of the parameters matching.

```python3
"""
$ CONDA_SUBDIR=osx-64 mamba env create --name double-exponential "openff-toolkit=0.10.6" "smirnoff-plugins" openeye -c conda-forge -c openeye
...
$ mamba activate double-exponential
$ pip install git+https://github.com/jthorton/de-forcefields
"""
import openmm
import openmm.unit
from openff.toolkit.topology import Molecule
from openff.toolkit.typing.engines.smirnoff import ForceField

force_field = ForceField("de-force-1.0.0.offxml", load_plugins=True)

molecule = Molecule.from_smiles("CCO")
molecule.generate_conformers(n_conformers=1)

topology = molecule.to_topology()
topology.box_vectors = [4, 4, 4] * openmm.unit.nanometer

system = force_field.create_openmm_system(topology)

simulation = openmm.app.Simulation(
    topology.to_openmm(),
    system,
    openmm.LangevinIntegrator(300, 1, 0.002),
    openmm.Platform.getPlatformByName("CPU"),
)

simulation.context.setPositions(molecule.conformers[0])
simulation.context.setPeriodicBoxVectors(*topology.box_vectors)

state = simulation.context.getState(getEnergy=True)

print(state.getPotentialEnergy().in_units_of(openmm.unit.kilojoule_per_mole))

with open('reference.xml', 'w') as f:
    f.write(openmm.XmlSerializer.serialize(system))
```


When I run this today, I get 13.591709748611304 kJ/mol, which is pretty close to what I see in the corresponding unit test. It's important that OpenEye is installed and available to get that number, otherwise the partial charges are very different and I get or 37.9516622967221. (Looking at CI [logs](https://github.com/openforcefield/openff-interchange/actions/runs/4247603129/jobs/7385826270), there's an extra ~0.001 kJ/mol difference for just this tiny molecule _just for switching from macOS to Linux_!)

Two quirks that are currently included in this code:

I separate out the 1-4 coluombic interactions into a separate `openmm.CustomBondForce`. The current `smirnoff-plugins` implementation just sets the interactions directly on the `openmm.NonbondedForce` that stores the charges. In both cases, the intermolecular electrostatics are handled by an `openmm.NonbondedForce` with sigmas and epsilons zeroed out. Eyeballing the XMLs shows the same parameters, just organized differently. My choice here doesn't help anything in OpenMM world but it's the only way I know of to match GROMACS's ability to separate out 1-4 coulombic interactions from the rest of the electrostatics.

The current DE implementation pre-computes parameters for ease of compute, i.e. $\epsilon_i$ and $\epsilon_j$ provided from the force field are processed as $\sqrt\epsilon_i$ and $\sqrt\epsilon_j$, same with $r_{min,1}, r_{min,1}$ actually being  $r_{min,1} /2,  r_{min,1}/2$. This is helpful so that OpenMM doesn't need to compute a bunch of square roots, but confused me until I was able to trace everything back through. I went back and forth on how this should be implemented; right now it's more legible to me but without pre-computing, which is probably the worse of the two options. I need to double-check that the parameters are being evaluated correctly for the 1-4 interactions.

jthorton:
This is awesome great work @mattwthompson !

PR #605, opened 2023-03-03
Improve support for virtual sites in plugins exported to OpenMM

mattwthompson:
### Description

Also move Buckingham support from first-class to a plugin.

### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/interop/openmm/_nonbonded.py
+++ b/openff/interchange/interop/openmm/_nonbonded.py
@@ -86,10 +86,9 @@ def _process_nonbonded_forces(
     else:
         molecule_virtual_site_map = defaultdict(list)
 
-    # Mapping between OpenFF "particles" and OpenMM particles (via inddex). OpenFF objects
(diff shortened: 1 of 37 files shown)


openff/interchange/interop/openmm/_nonbonded.py
```diff
@@ -650,28 +672,27 @@ def _create_multiple_nonbonded_forces(
...
+        openff_indices = (
+            interchange.topology.atom_index(atom1),
+            interchange.topology.atom_index(atom2),
+        )
 
-        # If the interactions are both zero, assume this is a 1-2 or 1-3 interaction
-        if q._value == 0 and eps._value == 0:
-            pass
-        else:
-            # Assume this is a 1-4 interaction
```
mattwthompson, line 673: This was fun to replace - this assumption fails when there's no vdW interactions!

PR #613, opened 2023-03-10
Support SMIRNOFF plugins modifying OpenMM objects without boostrapping

mattwthompson:
### Checklist
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/interop/openmm/__init__.py
+++ b/openff/interchange/interop/openmm/__init__.py
@@ -110,6 +110,19 @@ def to_openmm(
         particle_map=particle_map,
     )
 
+    for collection in interchange.collections.values():
+        if collection.is_plugin:
+            try:
+                collection.modify_openmm_forces(
+                    interchange,
+                    openmm_sys,
+                    add_constrained_forces=add_constrained_forces,
+                    constrained_pairs=constrained_pairs,
+                    particle_map=particle_map,
+                )
+            except NotImplementedError:
+                continue
+
     return openmm_sys
 
 
--- a/openff/interchange/interop/openmm/__init__.py
+++ b/openff/interchange/interop/openmm/__init__.py
@@ -55,7 +55,7 @@ def to_openmm(
 
     Returns
     -------
-    openmm_sys : openmm.System
+    system : openmm.System
         The corresponding OpenMM System object
 
     """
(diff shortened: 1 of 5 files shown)


mattwthompson:
This is a script that implements Lennard-Jones-like 12-6-4 ion [parameters from the Merz group](https://pubs.acs.org/doi/10.1021/jp505875v) a few years back. Since the functional form only adds one term, all that I needed to add was a single `CustomNonbondedForce` to cover that $\frac{C_{ij}}{r^4}$ term. (The partical charge and 12-6 terms are added on to the existing Electrostatics and vdW handlers.)  This script just builds a small system and runs an energy minimization as a sanity-check that MD simulations shouldn't immediately crash.

There's a bit more housekeeping that's needed to make OpenMM happy, and it's not optimized for LOC - it's optimized for flexibility. The `modify_openmm_forces` method is passed the already-constructed `Interchange` object, the already-constructed `openmm.System` object, and some other variables that may or may not be helpful. **Users can write plugins that do anything they want in that method** - add, modify, and delete forces, particles, and who knows what else. This method does not expect a return value since the system can be modified in-place.

```python3
import numpy
import openmm
import openmm.app
import openmm.unit
from openff.toolkit import ForceField, Molecule, Topology
from openff.units import unit

from openff.interchange.drivers.openmm import _get_openmm_energies

topology = Topology.from_molecules(
    [Molecule.from_smiles(smiles) for smiles in ["O", "[#13+3]"]]
)

force_field = ForceField("tip3p-1.0.0.offxml", load_plugins=True)

lj_handler = force_field.get_parameter_handler("vdW")
c4_ion_handler = force_field.get_parameter_handler("C4Ion")

AL_SMIRKS = "[#13X0+3:1]"

force_field.get_parameter_handler("vdW").add_parameter(
    {
        "smirks": AL_SMIRKS,
        "id": "n-12-6-4-Al3+",
        "rmin_half": unit.Quantity(1.369, unit.angstrom),
        "epsilon": unit.Quantity(0.01128487, unit.kilocalorie_per_mole),
    },
)

force_field.get_parameter_handler("LibraryCharges").add_parameter(
    {
        "smirks": AL_SMIRKS,
        "id": "n-12-6-4-Al3+",
        "charge1": unit.Quantity(3, unit.elementary_charge),
    },
)

c4_ion_handler.add_parameter(
    {
        "smirks": AL_SMIRKS,
        "id": "lj4-12-6-4-Al3+",
        "c": unit.Quantity(399, unit.kilocalorie_per_mole * unit.angstrom**4),
    },
)

system = Interchange.from_smirnoff(force_field, topology).to_openmm(
    combine_nonbonded_forces=True
)

with open("ions.xml", "w") as f:
    f.write(openmm.XmlSerializer.serialize(system))

positions = openmm.unit.Quantity(
    numpy.array(
        [
            [-1.992, 0.918, 0.221],
            [-2.099, 1.091, -0.748],
            [-2.723, 1.432, 0.647],
            [-1.908, 4.685, -0.664],
        ]
    ),
    openmm.unit.angstrom,
)

simulation = openmm.app.Simulation(
    topology.to_openmm(),
    system=system,
    integrator=openmm.LangevinIntegrator(1, 1, 300),
)

simulation.context.setPositions(positions)
simulation.minimizeEnergy()

minimized_positions = simulation.context.getState(getPositions=True).getPositions(
    asNumpy=True
)
print(minimized_positions)
initial_energy = _get_openmm_energies(
    system,
    box_vectors=unit.Quantity(50 * numpy.eye(3), unit.angstrom).to_openmm(),
    positions=positions,
    platform="CPU",
)["Nonbonded"]
minimized_energy = _get_openmm_energies(
    system,
    box_vectors=unit.Quantity(50 * numpy.eye(3), unit.angstrom).to_openmm(),
    positions=minimized_positions,
    platform="CPU",
)["Nonbonded"]

print(
    f"Initial energy:   {str(round(initial_energy, 3))}\n"
    f"Minimized energy: {str(round(minimized_energy, 3))}"
)

# Initial energy:   67.294 kilojoule / mole
# Minimized energy: -428.041 kilojoule / mole
```

The final geometry has the Al3+ ion close enough to the oxygen that Avogadro 2 draws a bond - I think this is an expected result given how strong those interactions should be.

image

mattwthompson:
Looking into this a bit more, I am re-considering whether or not this is a good example. I can't figure out of this force field is using TIP3P or a modified version of TIP3P with an added C4 term. It's not clear whether the water-water interactions are modified, though I don't think they are. It would be hard to add ion-oxygen interactions without also adding oxygen-oxygen interactions or also updating infrastructure to add out-of-spec non-bonded terms.

I found a [tutorial](https://ambermd.org/tutorials/advanced/tutorial20/12_6_4.htm) that uses this force field it but it's deep in Amber tools that I find hard to use.

PR #614, opened 2023-03-12
Support SMIRNOFF GBSA parameters

mattwthompson:
### Description
This PR adds `SMIRNOFFGBSACollection` and its associated method creating an OpenMM force.

### Checklist
- [x] Add tests
- [x] Lint
- [ ] Update docstrings


mattwthompson:
I spent some time looking into how best to test this but couldn't find a concise way. The [toolkit's GBSA tests](https://github.com/openforcefield/openff-toolkit/blob/0.10.6/openff/toolkit/tests/test_forcefield.py) are incredibly complicated and I'd like to avoid replicating all of that logic in new tests.

I'll put it out in the next beta and see if users report anything off.

PR #633, opened 2023-03-24
Begin supporting InterMol-styled `MDConfig`

### LINKED ISSUE #626: Add high-level method for transforming into "InterMol settings"
**Description**
As @mrshirts [demonstrated](https://pubmed.ncbi.nlm.nih.gov/27787702/), the only practical way to compare single-point energies across different engines is to modify some settings related to _how_ the calculations happen (but not modifying the actual parameters, i.e. partial charges). It would be useful to support, for testing purposes, post-creation modification of `Interchange` objects to match the MD settings used in InterMol's test suite. This might involve making a copy in order to keep the initial state clean. This would help streamline projects such as #615 by controlling for most potential causes of energy differences that are NOT simply a matter of parameters being written out differently.

This is clearly in conflict with the philosophy of SMIRNOFF - modifying PME electrostatics prescribed by the force field into simple cut-off electrostatics is unsupported - but it's clearly useful in testing. It's trivial to discover false positives (i.e. #552) in which two engines are given the same charges but report different electrostatics energies, particularly with large systems. It should be documented that this is for testing purposes only and not generally recommended for use.

mattwthompson:
### Description

Resolves #626
### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings
- [ ] Check writing
  - [ ] GROMACS inputs
  - [ ] LAMMPS inputs
  - [ ] Amber inputs
  - [ ] OpenMM objects
- [ ] Incorporate into existing tests


mattwthompson:
I'm still figuring out some of these details. Frustratingly,

* SMIRNOFF does not support cut-off electrostatics (https://github.com/openforcefield/standards/issues/6) so I'm out in the deep end.
* I thought OpenMM didn't support cut-off electrostatics without reaction-field attenuation, but I've been running some simple tests that calls that into doubt.

Getting this working will also require getting some more cases of non-bonded settings working in LAMMPS and probably a couple of other engines. I'll try to figure out exactly what's needed and spin those off as separate tasks.

mattwthompson:
This isn't complete but I fear it'll drift over time if I leave it as a large PR, so I'm going to break this into a few smaller ones.

PR #634, opened 2023-03-27
Improve plugin docs

mattwthompson:
### Description

The current documentation and examples for plugins in `v0.3.0beta4` could use some improvement.

https://docs.openforcefield.org/projects/interchange/en/latest/using/plugins.html

### Checklist

- [ ] Add tests
- [x] Lint
- [ ] Update docstrings

### DIFF
--- a/docs/_templates/autosummary/base.rst
+++ b/docs/_templates/autosummary/base.rst
@@ -1,0 +1,15 @@
+{% block title -%}
+
+{{ ("``" ~ objname ~ "``") | underline}}
+
+{%- endblock %}
+{% block base %}
+
+.. currentmodule:: {{ module }}
+
+.. auto{{ objtype }}:: {{ objname }}
+    {% if objtype in ["attribute", "data"] -%}
+    :no-value:
+    {%- endif %}
+
+{%- endblock %}
--- a/docs/_templates/autosummary/class.rst
+++ b/docs/_templates/autosummary/class.rst
@@ -1,4 +1,4 @@
-{{ objname | escape | underline }}
+{{ ("``" ~ (objname | escape) ~ "``") | underline('=')}}
 
 .. currentmodule:: {{ module }}
 
--- a/docs/_templates/autosummary/module.rst
+++ b/docs/_templates/autosummary/module.rst
@@ -1,8 +1,9 @@
 {% if fullname == "openff." ~ objname -%}
-{{ fullname | escape | underline(line="=")}}
+{%- set title = fullname | escape -%}
 {%- else -%}
-{{ objname | escape | underline(line="=")}}
+{%- set title = objname | escape -%}
 {%- endif %}
+{{ ("``" ~ title ~ "``") | underline('=')}}
 
 .. automodule:: {{ fullname }}
    :no-members:
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -65,6 +65,8 @@ intersphinx_mapping = {
         None,
     ),
     "mdtraj": ("https://www.mdtraj.org/1.9.5/", None),
+    "setuptools": ("https://setuptools.pypa.io/en/latest/", None),
+    "openmm": ("http://docs.openmm.org/latest/api-python/", None),
 }
 
 mermaid_init_js = """mermaid.initialize({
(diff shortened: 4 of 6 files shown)


mattwthompson:
https://openff-docs--634.org.readthedocs.build/projects/interchange/en/634/using/plugins.html

Yoshanuikabundi:
Hi Matt! I'm looking this over and noticing some things, like for instance that there are no API docs for `SMIRNOFFCollection`. I can use this opportunity to give the docs a once over (they look like they can use some TLC) or just review these changes - what would you prefer and what kind of timeframe do you have for release?

Yoshanuikabundi (commented):
OK It's really hard to test and write suggestions for Markdown in reviews, and there was some low-hanging fruit for fixing up some API docs issues that I noticed, so I've opened #637 as my review.

This write-up is really, really good. The writing is clear and descriptive, and the example is fantastic - shows all the details, uses the same language as the write-up so its easy to follow, and it's short! All I really did is add a paragraph or two introducing things more gently, tweak some phrasing, and add links everywhere.

I really like how you've designed the collection ABCs - all the abstract methods are public, so they are visible in the docs, and this page makes understanding them pretty simple. My only suggestion for the code would be to move from `*args, **kwargs` to full, typed signatures everywhere (eg, in `modify_openmm_forces`), but it's a small point that is more than adequately explained by the docs you've written.

I've also taken the liberty of re-exporting `SMIRNOFFCollection` from the `plugins` module (it's necessary to add it to `__all__` in this case so that Autosummary knows its supposed to be public), fixing Sphinx's issues around it (this "jax" dependency seems to directly link to CUDA, which isn't installed in the docs environment - we don't seem to use it directly, so mocking it does no harm), and very slightly tweaking the API templates for more consistency and a nicer look.

It might also be worth linking to the plugin page from the docstrings of `Collection` and `SMIRNOFFCollection`, and making sure all the public stuff has docstrings. Just in case you don't know, attributes/fields can be given docstrings by putting the string right after the line:

```python
class foo:
    some_attribute: str = "hi"
    """This is the docstring for the above attribute"""
```

So there is no limit to the amount of busywork a developer can do in documenting things!!!

Let me know if there's anything else you'd like me to take a look at before release!

mattwthompson:
> It's really hard to test and write suggestions for Markdown in reviews

I hear ya ... I go back and forth between wrapped and unwrapped lines, neither of which is great for diffs. This would be easier if I spelled as well as I did in middle school, oops.

> My only suggestion for the code would be ...

Did you see this anywhere but the definition of `modify_openmm_forces` in the base class? And is it best practice to have the full function signature on the base class even though it's never called (on that class)?

> attributes/fields can be given docstrings by

Is this a consequence of a Sphinx plugin/extension/thingy, something built into Pydantic, or more Python magic I don't know? And is this recommended or different than using `Field(..., description="blah")`? I picked this up from reading some code (I think QCArchive?) ages ago:

```python
class Foo:
    some_attribute: str = "hi"
    """This is the docstring for the above attribute"""


class Bar:
    some_attribute: str = Field(
        "bye",
        description="This is the docstring for the above attribute",
    )

Yoshanuikabundi:
> I go back and forth between wrapped and unwrapped lines, neither of which is great for diffs. 

Mmm. I think I've settled on unwrapped (breaking at paragraphs). It's a pain with the basic diff highlighting, but GitHub can usually highlight the parts of the line that have changed. Much better than when an edit causes the whole paragraph to re-flow.

> Did you see this anywhere but the definition of modify_openmm_forces in the base class?

Nope, don't think so

> And is it best practice to have the full function signature on the base class even though it's never called (on that class)?

Yeah I think so, especially if the function signature forms part of the abstract interface. It just makes it so much clearer what you expect subclasses to implement, whether you're looking at code or docs. If subclasses are supposed to use whatever arguments they like then it's less important, but if subclasses should implement a particular interface I think that interface should be on the base class wherever possible. 

Documenting (abstract) base classes that users will inherit from is definitely still new for me though, and I'm still thinking through the best ways to do it.

> Is this a consequence of a Sphinx plugin/extension/thingy, something built into Pydantic, or more Python magic I don't know? 

Built in to Python! The general rule is something like any string literal expression immediately following an object's declaration is its docstring. For classes and functions that goes at the top of the inner block because the `def` or `class` line is the declaration, but for attributes and so forth the declaration is the `foo = bar` line.

> And is this recommended or different than using Field(..., description="blah")? I picked this up from reading some code (I think QCArchive?) ages ago:

Autodoc_pydantic will document either! The advantage of `description` is that it gets included in the Pydantic schema. The advantage of the docstring approach is that more IDEs can display it. I wish Pydantic could interpret a docstring as the description if a field is missing a description, but apparently that's hard (and of course now I can't find the relevant issue).

PR #637, opened 2023-03-29
Review for #634

Yoshanuikabundi:
### Description

This PR is a review for #634, see that PR for discussion.


docs/using/plugins.md
```diff
@@ -1,65 +1,88 @@
...
+Custom interactions (i.e., interactions other than 12-6 Lennard-Jones, harmonic bonds, or harmonic angles) can be introduced into the OpenFF stack via plugin interfaces. To create a plugin, create a subclass of both [`ParameterHandler`] and [`SMIRNOFFCollection`] for each new type of physical interaction in the model, then expose them to Interchange via the `openff.toolkit.plugins.handlers` and `openff.interchange.plugins.collections` [SetupTools entry points], respectively.
 
-Currently, these custom interactions can only be exported to OpenMM. There are two routes for this export: one using existing Interchange machinery and another that allows for completely custom behavior.
+Parameter handlers are responsible for the system- and engine-independent aspects of parametrization; in other words, they define the force field itself. By contrast, a SMIRNOFF collection handles the parametrization of a particular system and the preparation of that system for simulation with an external MD engine. The base classes for parameter handlers are defined in the OpenFF Toolkit, while collections are defined in Interchange.
```
mattwthompson, line 5: Since the first paragraph introduces something new and this is a helpful back-step to clarify the existing infrastructure, it'd be nice to communicate "the OpenFF stack ships with everything SMIRNOFF already, you're only responsible for writing the custom stuff you want to tack on." The last sentence has some implication of the first clause but I think I don't think that's the main message. I couldn't find a way to add this in without being repetitive

docs/using/plugins.md
```diff
@@ -1,65 +1,88 @@
...
-  * inherit from `ParameterHandler`
-  * be available in the `openff.toolkit.plugins.handlers` entry point group
-  * contain a class that inherits from `ParameterType` (a "type class")
-  * define a class attribute `_TAGNAME` that is used for serialization and handy lookups
-  * define a class attribute `_INFOTYPE` that defines the "type class"
+  * inherit from [`ParameterHandler`]
+  * be available in the `openff.toolkit.plugins.handlers` [entry point group]
+  * contain a class that inherits from [`ParameterType`] (a "type class")
+  * define a class attribute `_TAGNAME: str` that is used for serialization and handy lookups
+  * define a class attribute `_INFOTYPE` that specifies the handler's "type class"
```
mattwthompson, line 30:
```suggestion
  * define a class attribute `_INFOTYPE: Type[ParameterType]` that specifies the handler's "type class"
```
might be too much? not sure if this is even the correct annotation ... I hadn't considered using type annotations in-line in prose before seeing you do this!

docs/using/plugins.md
```diff
@@ -1,65 +1,88 @@
...
-  * define a class method `allowed_parameter_handlers` that returns an iterable of `ParameterHandler` subclasses that it can process
-  * define a class method `supported_parameters` that returns an interable of parameters that it expects to store (i.e. `smirks`, `k`, `length`, etc.)
-  * override methods of `SMIRNOFFCollection` (`store_matches`, `store_potentials`, `create`) as needed
+  * inherit from [`SMIRNOFFCollection`] and therefore be written in the style of a [Pydantic model].
+  * define a field [`is_plugin: bool = True`]
+  * be available in the `openff.interchange.plugins.collections` [entry point group]
+  * define a field [`type: str`], which we suggest match the corresponding `ParameterHandler._TAGNAME`
+  * define a field [`expression: str`], a mathematical formula defining how it contributes to the overall potential energy
+  * define a class method [`allowed_parameter_handlers`] that returns an iterable of the `ParameterHandler` subclasses that it can process
+  * define a class method [`supported_parameters`] that returns an iterable of parameters that it expects to store (i.e. `"smirks"`, `"k"`, `"length"`, etc.)
```
mattwthompson, line 68: Out of scope here, but I think a sore point in our infrastructure is that "parameter" can easily refer to two or more things, one of which contains the other

```
x = sage["vdW"].parameters[-1]
y = sage["vdW"].parameters[-1].sigma
```

if, using conversational English, I was trying to describe something like "grab this parameter, do this thing, ..." you'd have no clue which I was referring to. I intuitively think that `y` is more accurately described as a "parameter" but I have a hard time saying that indexing into `ParameterHandler.ParameterList` doesn't also give you a "parameter" !

I don't  help the cause on the Interchange side of things - while I find this is natural to work with:

```python3
In [15]: [*out["vdW"].potentials.values()][-1]
Out[15]: Potential(parameters={'sigma': <Quantity(0.534539231, 'angstrom')>, 'epsilon': <Quantity(1.23259997e-05, 'kilocalorie / mole')>}, map_key=None)

In [16]: [*out["vdW"].potentials.values()][-1].parameters
Out[16]:
{'sigma': 0.5345392308836691 <Unit('angstrom')>,
 'epsilon': 1.232599966667e-05 <Unit('kilocalorie / mole')>}
```
it's hard to argue that it doesn't further overload the word.

Totally out of scope here - wanted to jot my thoughts on this wart _somewhere_ - feel free to move this to an issue somewhere if you think it's feasible for us to make progress on this.

mattwthompson (approved):
Thanks for the edits - I've read through the rendered docs and am happy with it. Looking over other comments briefly before merging

docs/using/plugins.md
Yoshanuikabundi: Mmm true. This seems like its also at the root of how we have a type named `ParameterType`, which we apparently sometimes use as `Type[ParameterType]`... Maybe if we had another word that meant the `x = sage["vdW"].parameters[-1]` parameter we could rename that type to match and simplify things.

docs/using/plugins.md
Yoshanuikabundi: I think it is too much, especially since it might not be clear to newcomers what `Type[foo]` means. I think the fact that it specifies a class rather than an instance is clear enough, especially given the examples, and also we do later define a type class... should be fine.

PR #643, opened 2023-04-05
Automatically detect API changes

mattwthompson:
https://mkdocstrings.github.io/griffe/checking/#detected-breakages


mattwthompson:
Cool, if I remove an arbitrary method it complains:
image

Issue #647, opened 2023-04-06
Support multiple `ParameterHandler`s in one `SMIRNOFFCollection`


wwilla7:
**Description**
When creating custom `ParameterHandler`s via [plugins](https://docs.openforcefield.org/projects/interchange/en/latest/using/plugins.html), if there are multiple handlers in one collection, interchange currently returns RuntimeError [here](https://github.com/openforcefield/openff-interchange/blob/main/openff/interchange/smirnoff/_create.py#L59-L60):
```
if len(parameter_handlers) != 1:
    raise RuntimeError
```
I have a temporary workaround [here](https://github.com/openforcefield/openff-interchange/compare/main...wwilla7:openff-interchange:pol) that works for [my case](https://github.com/wwilla7/MPID_plugin/blob/main/mpid_plugin/nonbonded.py#L84-L97), but it needs to be designed more carefully. If this is a desired feature, I can put in more effort and create a PR.

mattwthompson:
I think this is a totally valid use case - that exception probably stems from a previous implementation of that handler-collection mapping that has little reason to exist permanently.

Your changes look like a good start - I'd like to build out some more tests around them to ensure they work for other cases as well. Could you open them as a PR and then I'll take it over?

wwilla7:
Thanks for the response. I have created the [PR#648](https://github.com/openforcefield/openff-interchange/pull/648).

mattwthompson:
Resolved in #651, should land in version 0.3.1 in late April. Thanks for the contribution @wwilla7!

PR #648, opened 2023-04-06
Support multiple `ParameterHandlers` in one `SMIRNOFFCollection`

wwilla7:
### Description

Responds to [Issue#647](https://github.com/openforcefield/openff-interchange/issues/647)

I have left comments in the code block regarding:
1. If more than one collection is loaded by the plugin.
2. If there is only one handler in the collection.

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/smirnoff/_create.py
+++ b/openff/interchange/smirnoff/_create.py
@@ -58,7 +58,7 @@ for collection_plugin in load_smirnoff_plugins():
 
     # if len(parameter_handlers) != 1:
     #     raise RuntimeError
-    
+
     for parameter_handler in parameter_handlers:
         if parameter_handler in load_handler_plugins():
             _SUPPORTED_PARAMETER_HANDLERS.add(parameter_handler._TAGNAME)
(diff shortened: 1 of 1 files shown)


openff/interchange/smirnoff/_create.py
```diff
@@ -348,13 +349,30 @@ def _plugins(
...
+    
+    collection = collection_class.create(
+        parameter_handler=handlers,
+        topology=topology,
+    )
+
+    interchange.collections.update(
+        {
+            # handler_class._TAGNAME: collection,
+            collection.type: collection, # Since handlers have different tagnames, I would match collection type instead of handler
```
mattwthompson: I think this is a great idea

mattwthompson:
@wwilla7 I made a couple of changes to get this working with the existing tests; I don't think your case should be affected but I don't have it integrated into these tests yet. (Some tests may still fail from your fork since we don't expose the OpenEye license to forks, but if everything here is good I'll move these changes to a separate PR that can access it.) Could you pull down the latest changes and let me know if they work with your plugin?

I pushed directly to your branch, so pulling from your remote to your local clone should bring the changes in. Just make sure 08f9e3b37acf03f0bbe75d10a077cc4055e1a2c9 is the most recent commit in your `git log` and either re-install Interchange or ensure you were using an editable install (i.e. `pip install -e .`).

wwilla7:
Hi, I pulled and tested it with my plugin. Everything works fine, thanks for following up on this!

mattwthompson:
Superseded by #651

PR #663, opened 2023-04-18
Improve speed of Amber export

mattwthompson:
### Description

The current Amber exporter relies heavily on `Topology.atom(index)`, which is slow and scales poorly with larger system sizes. This PR takes some shortcuts around this and also begins to shuffle a monolithic (~500 LC) function into smaller modular bits.

From benchmarking a medium-sized system:

```python
from openff.interchange import *
from openff.toolkit import *

n = 100

molecule = Molecule.from_smiles("CCO")
topology = Topology.from_molecules(n * [molecule])

interchange = Interchange.from_smirnoff(force_field, topology)
```

The speedup of `interchange.to_prmtop` and the various `_get_` functions are as follows:

```
%%snakeviz
interchange.to_prmtop("tmp.prmtop")
# 1.75 total -> 0.514
# 0.814 dihedral -> 0.271
# 0.608 angle -> 0.0188
# 0.133 bond -> 0.0116
```

There's still some optimization to be had with how exclusions are prepared, but I might not get to that here.

### Checklist

- [x] Avoid `Topology.atom` when only the atomic number is needed
- [x] Try to use tuples when iterators are not modified after creation
- [x] Move Amber module
- [x] Maintain old import path
- [x] Refactor some of `to_prmtop` out to smaller functions
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
Now down to 0.332, 0.0524, 0.0209, 0.0129

mattwthompson:
For larger systems, `_pair_in_known_14_pairs` is the bottleneck.

PR #673, opened 2023-04-21
Add vectorized representation example

mattwthompson:
### Description

This PR splits off what was previously part of a larger example that is no longer supported.

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
I thought about making this [prettier](https://rich.readthedocs.io/en/stable/tables.html) but didn't try too hard to make it happen.

PR #680, opened 2023-04-24
Fix text wrapping and exclusions in Amber export

### LINKED ISSUE #678: Atom names can be mangled in Amber export
**Description**

It's possible for whitespace to disrupt text wrapping, resulting in blocks like this that some other parsers choke on:

```
%FLAG ATOM_NAME
%FORMAT(20a4)
 N1  C1  C2  N2  C3  C4  H1  C5  H2  C6  O1  C7  H3  H4  H5  C8  O2  C9  H6  H7
  H8  C10 H9  H10 N3  C11 C12 C13 H11 C14 H12 C15 C16 C17 H13 N4  H14 C18 H15
C19 H16 C20 H17 N5  C21 H18 C22 H19
```

**Reproduction**

Stems from #676

### LINKED ISSUE #679: Duplicate excluded atom pairs in Amber export
**Description**

In some polycyclic compounds, the logic for collecting 1-4 pairs can produce duplicates. This does not obviously produce problems with sander, but ParmEd and therefore OpenMM reject parsing it.

**Reproduction**

Stems from #676

mattwthompson:
### Description

Resolves #678
Resolves #679

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -15,6 +15,11 @@ Please note that all releases prior to a version 1.0.0 are considered pre-releas
 
 * #671 Adds `Interchange.to_gromacs` which writes both GROMACS topology and coordinate files.
 
+### Bugfixes
+
+* #680 Fixes #678 in which, in some cases, text wrapping in Amber files was mangled.
+* #680 Fixes #679 in which atom exclusion lists in Amber files were written incorrectly.
+
 ## 0.3.1 - 2023-04-19
 
 ### Behavior changes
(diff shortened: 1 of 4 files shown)


mattwthompson:
The problem here is that I wasn't following the weird behavior of specifying an atom's `NUMBER_EXCLUDED_ATOMS` entry as 1 when it has no excluded atoms, though I was including the nonsense "0" atom in the corresponding `EXCLUDED_ATOMS_LIST`. I don't understand why it has to be this way, but it does.

> If an atom has no excluded atoms (either because it is a monoatomic ion or
all atoms it forms a bonded interaction with have a smaller index), then it
is given a value of 1 in this list which corresponds to an exclusion with (a
non-existent) atom 0 in EXCLUDED ATOMS LIST

https://ambermd.org/prmtop.pdf

PR #684, opened 2023-04-27
Fix docs environment and tidy up pack_box docstring

Yoshanuikabundi:
### Description

Toolkit 0.13.0 broke the docs environment - it would solve but then `from openff.toolkit import Molecule, ForceField, Topology` would raise an ImportError. I just copied over the deps from the test environment and it seems to work, so hopefully this is some change in a dependency that is already reflected in the forge recipe. Also tidied up the pack_box docstring.

### Checklist

- [x] Lint
- [x] Update docstrings

### DIFF
--- a/devtools/conda-envs/docs_env.yaml
+++ b/devtools/conda-envs/docs_env.yaml
@@ -7,14 +7,30 @@ dependencies:
   # Base depends
   - python
   - pip
-  - pydantic
-  - pint
-  - openff-toolkit-base >=0.11.3
-  - openff-models >=0.0.4
+  - numpy >=1.21
+  - pydantic <2.0.0a0
   - openmm >=7.6
+  - openff-toolkit >=0.11.3
+  - openff-models >=0.0.4
+  - openff-nagl
+  - unyt
+  - mbuild
+  - foyer >=0.11.3
   - mdtraj
   - intermol
   - jax
+  - nbval
+  - nglview
+  - ipywidgets >7,<8  # https://github.com/nglviewer/nglview/issues/1032
+  - mdanalysis
+  - gromacs >=2021=nompi*
+  - lammps
+  # https://github.com/conda-forge/quippy-feedstock/issues/15
+  - panedr
+  - mypy =1.2
+  - typing-extensions
+  - types-setuptools
+  - pandas-stubs >=1.2.0.56
   # readthedocs dependencies
   - myst-parser
   - numpydoc
@@ -23,4 +39,5 @@ dependencies:
   - sphinxcontrib-mermaid
   - sphinx-notfound-page
   - pip:
+    - git+https://github.com/jthorton/de-forcefields
     - git+https://github.com/openforcefield/openff-sphinx-theme.git@main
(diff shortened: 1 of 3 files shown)


Yoshanuikabundi:
I'm also happy to tweak pack_box so that it returns a topology rather than an MDTraj trajectory if that would be useful!

Yoshanuikabundi:
I've simplified the logic calculating the expected volume and added an argument to avoid adding buffers to the box size - in my (limited) experience the buffers aren't necessary and they're making self-assembly simulations a bit of a pain.

mattwthompson:
Super confused about how the release broke the docs builds. I guess the size of this environment isn't a problem? The builds are much longer (~150 to ~400 seconds) but I don't really care about that.

openff/interchange/components/packmol.py
```diff
@@ -629,34 +623,37 @@ def pack_box(
...
-        equal to the length of `molecules`.
+        equal to the length of ``molecules``.
     structure_to_solvate: str, optional
         A file path to the PDB coordinates of the structure to be solvated.
     center_solute: bool
-        If `True`, the structure to solvate will be centered in the
-        simulation box. This option is only applied when `structure_to_solvate`
-        is set.
+        If ``True``, the structure to solvate will be placed in the center of
+        thesimulation box. This option is only applied when
```
mattwthompson, line 631:
```suggestion
        the simulation box. This option is only applied when
```

devtools/conda-envs/docs_env.yaml
```diff
@@ -7,14 +7,30 @@ dependencies:
...
+  - openff-nagl
+  - unyt
+  - mbuild
+  - foyer >=0.11.3
   - mdtraj
   - intermol
   - jax
+  - nbval
+  - nglview
+  - ipywidgets >7,<8  # https://github.com/nglviewer/nglview/issues/1032
```
mattwthompson, line 24: FTR this is fixed upstream but they haven't made a release yet

devtools/conda-envs/docs_env.yaml
```diff
@@ -7,14 +7,30 @@ dependencies:
...
-  - pydantic
-  - pint
-  - openff-toolkit-base >=0.11.3
-  - openff-models >=0.0.4
+  - numpy >=1.21
+  - pydantic <2.0.0a0
   - openmm >=7.6
+  - openff-toolkit >=0.11.3
+  - openff-models >=0.0.4
+  - openff-nagl
```
mattwthompson, line 15:
```suggestion
  - openff-toolkit >=0.13
  - openff-models >=0.0.4
```

mattwthompson: I'm probably reviewing my own changes here, the test environment is a pain in the rear to simultaneously keep up to date and minimal to the use case

mattwthompson (approved):
The changes to the `packmol` module look fine to me and render well with a quick spot check

It'd be nice to slim down the docs environment again but that's not a timely priority

mattwthompson:
> I'm also happy to tweak pack_box so that it returns a topology rather than an MDTraj trajectory if that would be useful!

I think this would be a pure improvement - if you have time and choose to do this, could you do it in another PR though?

mattwthompson:
(And to be explicit - please merge this yourself if/when you're happy with it, no need to wait for me!)

PR #692, opened 2023-04-29
Tag some features as opt-in experimental

mattwthompson:
### Description

This is an idea I've considered for a while but never tried. The problem is that many features need extensive testing to be considered complete yet need to exist in the package for a long time in order for that testing to be feasible. Throwing warnings around is not ideal as they can easily be ignored by users. Me personally and repeatedly communicating the nature of these features scales poorly and guarantees that some people aren't informed what is and is not currently trustworthy. Hiding things behind "private" import paths implies a lack of API stability but does not communicate whether or not something is trustworthy and leaves us more or less in the same position as before.

With this changeset, features tagged with an `@experimental` decorator require an environment variable `INTERCHANGE_EXPERIMENTAL=1` to be set. This requires users to explicitly opt-in with a non-negligible hurdle if using one of these features. Power users and heavy testers can set this in their shell profile without their work being interrupted.

Here's what it looks like to use:

```python3
In [1]: from openff.interchange._experimental import experimental
   ...:
   ...:
   ...: @experimental
   ...: def f():
   ...:     pass
   ...:
   ...:
   ...: f()

ExperimentalFeatureException              Traceback (most recent call last)
Cell In[1], line 9
      4 @experimental
      5 def f():
      6     pass
----> 9 f()

File ~/software/openff-interchange/openff/interchange/_experimental.py:15, in experimental.<locals>.wrapper(*args, **kwargs)
     13 def wrapper(*args, **kwargs):
     14     if os.environ.get("INTERCHANGE_EXPERIMENTAL", "0") != "1":
---> 15         raise ExperimentalFeatureException(
     16             f"\n\tFunction or method {func.__name__} is experimental. This feature is not "
     17             "complete, not yet reliable, and/or needs more testing to be considered suitable "
     18             "for production.\n"
     19             "\tTo use this feature on a provisional basis, set the environment variable "
     20             "INTERCHANGE_EXPERIMENTAL=1.",
     21         )
     23     return func(*args, **kwargs)

ExperimentalFeatureException:
	Function or method f is experimental. This feature is not complete, not yet reliable, and/or needs more testing to be considered suitable for production.
	To use this feature on a provisional basis, set the environment variable INTERCHANGE_EXPERIMENTAL=1.
```
Thoughts @j-wags @Yoshanuikabundi?

### Checklist

- [x] Add tests
- [x] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/_experimental.py
+++ b/openff/interchange/_experimental.py
@@ -1,6 +1,6 @@
+import os
 from functools import wraps
 from inspect import cleandoc
-import os
 
 from openff.interchange.exceptions import ExperimentalFeatureException
 
--- a/openff/interchange/_experimental.py
+++ b/openff/interchange/_experimental.py
@@ -10,15 +10,15 @@ def experimental(func):
     Decorate a function as experimental, requiring environment variable opt-in.
 
     To use the wrapped function, set the environment variable INTERCHANGE_EXPERIMENTAL=1.
-    """
 
+    """
     firstline, _, remainder = cleandoc(func.__doc__).partition("\n")
 
     func.__doc__ = (
         "🧪 "
-        + firstline
-        + "\n\n.. admonition:: Experimental\n\n    This object is experimental and should not be used in production.\n"
-        + remainder
+        f"{firstline}"
+        "\n\n.. admonition:: Experimental\n\n    This object is experimental and should not be used in production.\n"
+        f"{remainder}"
     )
 
     @wraps(func)
--- a/openff/interchange/_experimental.py
+++ b/openff/interchange/_experimental.py
@@ -1,3 +1,5 @@
+from functools import wraps
+from inspect import cleandoc
 import os
 
 from openff.interchange.exceptions import ExperimentalFeatureException
@@ -10,6 +12,16 @@ def experimental(func):
     To use the wrapped function, set the environment variable INTERCHANGE_EXPERIMENTAL=1.
     """
 
+    firstline, _, remainder = cleandoc(func.__doc__).partition("\n")
+
+    func.__doc__ = (
+        "🧪 "
+        + firstline
+        + "\n\n.. admonition:: Experimental\n\n    This object is experimental and should not be used in production.\n"
+        + remainder
+    )
+
+    @wraps(func)
     def wrapper(*args, **kwargs):
         if os.environ.get("INTERCHANGE_EXPERIMENTAL", "0") != "1":
             raise ExperimentalFeatureException(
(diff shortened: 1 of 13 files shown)


openff/interchange/_experimental.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def experimental(func):
+    """
+    Decorate a function as experimental, requiring environment variable opt-in.
+
+    To use the wrapped function, set the environment variable INTERCHANGE_EXPERIMENTAL=1.
+    """
+
+    def wrapper(*args, **kwargs):
```
Yoshanuikabundi, line 13: The `functools.wraps` decorator tells Python that a function is a wrapper around some other function, which preserves the signature and docstring and so on. PyRight seems to be able to figure it out on its own, but Sphinx couldn't.

We can also dynamically modify the docstring to add a test-tube symbol and a warning (or whatever). The call to `cleandoc` is important in case the docstring is indented (as almost all docstrings are) or starts with a newline.

```suggestion
from functools import wraps
from inspect import cleandoc
import os

from openff.interchange.exceptions import ExperimentalFeatureException


def experimental(func):
    """
    Decorate a function as experimental, requiring environment variable opt-in.

    To use the wrapped function, set the environment variable INTERCHANGE_EXPERIMENTAL=1.
    """

    firstline, _, remainder = cleandoc(func.__doc__).partition("\n")

    func.__doc__ = (
        "🧪 "
        + firstline
        + "\n\n.. admonition:: Experimental\n\n    This object is experimental and should not be used in production.\n"
        + remainder
    )

    @wraps(func)
    def wrapper(*args, **kwargs):
```

Yoshanuikabundi (commented):
I love this idea! I think this would be especially useful for adding experimental features to the public API without taking on the risk that users... use it. One risk is that users get used to setting the environment variable for everything, so I think we'd have to be careful when we use these features in examples and also take care to remove the decorator quickly when it becomes clear that the feature works. One potential policy might be to only use it in examples that are clearly marked as experimental, and NOT to include setting the env var in the example to force the users to acknowledge that it's being used.

Also - removing the "- git+https://github.com/jthorton/de-forcefields" dependency from the docs environment seems to fix the build for me locally without disrupting anything

mattwthompson:
> One risk is that users get used to setting the environment variable for everything

I'm worried about this as well, I thought briefly about having different tiers of danger but I couldn't think of a good way to implement it. Expanding this to multiple different variables seems like a bad idea. I don't love the idea of continuously emitting more warnings/tacking on headers to docstrings and pages but that's the best I can come up with.

mattwthompson:
Looks good, albeit this is the only case as the others are each hidden behind an underscore:

image

PR #695, opened 2023-05-01
FIX: Do not assume vdW handler exists

mattwthompson:
### Description

This PR fixes an assumption in the OpenMM export logic that a vdW collection exists with that name as a non-plugin.

### Checklist

- [ ] Test in Evaluator test environment
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
This is needed for https://github.com/openforcefield/openff-evaluator/pull/512 but something else is also needed.

PR #706, opened 2023-05-03
Update solvent example

mattwthompson:
### Description

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
This makes the examples take a couple of minutes longer, which I'm a little grumpy about, but don't want to fix now.

PR #707, opened 2023-05-04
Packmol improvements

Yoshanuikabundi:
### Description

This PR adds a bunch of features to the packmol wrapper. I probably shouldn't have spent time on it at this stage but it seems to work and it's much easier to use than the old wrapper. Up to you @mattwthompson if you want to merge this now, or after the meeting, or put it in the vignettes repo, or throw it away or whatever - just putting it here so you have the option!

I may have gone a bit overboard but I've tried to stage commits so you can just take the ones you like and leave the later ones off. I've tested manually on a lipid-water system and solvating a protein, but haven't written or changed any unit tests - I can if you decide you want to merge this. Here's what the PR does:

- pack_box now returns an OpenFF Topology instead of an MDTraj Trajectory
- The box size buffers to support PBCs have been overhauled; they're now tied to tolerance and the volume is not scaled automatically by the 1.1 fudge factor. This is a bit more principled and less surprising (to me). It also means that the generated simulation box actually has the requested density/size, but packing still converges and it still energy minimizes and simulates.
- the `structure_to_solvate` argument is now `topology_to_solvate` and takes a topology instead of a PDB file. I've removed a lot of now-unnecessary code that handles residue names and the like; there was no obvious, principled way to port them and I figure we should dogfood the Toolkit's handling of this
- MDTraj is no longer a dependency
- We can now pack triclinic boxes (like rhombic dodecahedra) - packmol only supports rectangular prisms so this works by computing the rectangular brick representation of the box and packing that. The `topology_to_solvate` is wrapped into the brick for packing so that packmol sees all of it, but the original coordinates are later restored so that all molecules are whole in the output. 
- The arguments to specify box vectors have been overhauled. It now takes a `box_shape` argument instead of `box_aspect_ratio`, and a `box_vectors` argument instead of `box_size`. They do what you expect.
- If `box_vectors` and `mass_density` are not specified, box vectors are now taken from `topology_to_solvate`
- The code has been generally streamlined and simplified
- a `solvate_topology` method has been added that takes a topology and adds saline and a box with customizable salt concentration, buffer distance, and shape. You can just call this on a protein topology and get solvated, salted, neutralised starting coordinates that actually work with reasonable defaults in a rhombic dodecahedron. This could be extended to support other solvent mixtures or ions if desired.
- unit image distance arrays specifying the shape of the major box types (cube, rhombic dodecahedron with a square in the XY plane, rhombic dodecahedron with a hexagon in the XY plane) have been added to make it easy to generate box vectors (just multiply by the desired image distance)

Headlines:
- Inputs and outputs `openff.toolkit.Topology` instead of `mdtraj.Trajectory` and PDB files
- Less surprising box sizes
- No more MDTraj
- Triclinic boxes that just work
- `solvate_topology` with reasonable defaults
- Net reduction in LoC

Demo notebook (I can put this in examples if desired): [pack_mol_demo.zip](https://github.com/openforcefield/openff-interchange/files/11392068/pack_mol_demo.zip)

solvate_topology

### Checklist

- [ ] Add tests
- [x] Lint
- [x] Update docstrings

### DIFF
--- a/openff/interchange/components/_packmol.py
+++ b/openff/interchange/components/_packmol.py
@@ -6,9 +6,10 @@ import shutil
 import subprocess
 import tempfile
 from distutils.spawn import find_executable
-from typing import Optional
+from typing import Callable, Optional
 
 import numpy as np
+from numpy.typing import ArrayLike, NDArray
 from openff.toolkit import Molecule, RDKitToolkitWrapper, Topology
 from openff.units import Quantity, unit
 from openff.utilities.utilities import requires_package, temporary_cd
@@ -22,6 +23,46 @@ UNIT_CUBE = np.asarray(
         [0, 0, 1],
     ],
 )
+"""A regular square prism with image distance 1.0 and volume 1.0.
+
+A cubic box is very simple and easy to visualize, but wastes substantial space
+and computational time. A rhombic dodecahedron that provides the same separation
+between solutes has only about 71% the volume, and therefore requires simulation
+of much less solvent.
+"""
+
+RHOMBIC_DODECAHEDRON = np.array(
+    [
+        [1.0, 0.0, 0.0],
+        [0.0, 1.0, 0.0],
+        [0.5, 0.5, np.sqrt(2.0) / 2.0],
+    ],
+)
+"""
+Rhombic dodecahedron with square XY plane cross section, image distance 1.0, and volume ~0.71.
+
+The rhombic dodecahedron is the most space-efficient triclinic box for a
+spherical solute, or equivalently for a solute whose rotations sweep out a
+sphere. Its volume is $\\frac{1}{2}\\sqrt{2}$. A square intersection with the
+XY plane allows the first two box vectors to be parallel to the x- and y-axes
+respectively, which is simple to picture and appropriate for soluble systems.
+"""
+
+RHOMBIC_DODECAHEDRON_XYHEX = np.array(
+    [
+        [1.0, 0.0, 0.0],
+        [0.5, np.sqrt(3.0) / 2.0, 0.0],
+        [0.5, np.sqrt(3.0) / 6.0, np.sqrt(6.0) / 3.0],
+    ],
+)
+"""
+A rhombic dodecahedron with hexagonal XY cross section, image distance 1.0, and volume ~0.71.
+
+The rhombic dodecahedron is the most space-efficient triclinic box for a
+spherical solute, or equivalently for a solute whose rotations sweep out a
+sphere. A hexagonal intersection with the XY plane is convenient for membrane
+simulations,
+"""
 
 
 def _find_packmol() -> Optional[str]:
@@ -46,8 +87,8 @@ def _validate_inputs(
     molecules: list[Molecule],
     number_of_copies: list[int],
     topology_to_solvate: Optional[Topology],
-    box_aspect_ratio: Optional[Quantity],
-    box_size: Optional[Quantity],
+    box_shape: NDArray,
+    box_vectors: Optional[Quantity],
     mass_density: Optional[Quantity],
 ):
     """
@@ -62,30 +103,30 @@ def _validate_inputs(
         equal to the length of `molecules`.
     topology_to_solvate: Topology, optional
         The OpenFF Topology to be solvated.
-    box_size : openff.units.Quantity, optional
-        The size of the box to generate in units compatible with angstroms.
-        If `None`, `mass_density` must be provided.
+    box_vectors : openff.units.Quantity,
+        The box vectors to fill in units compatible with angstroms. If `None`,
+        `mass_density` must be provided.
     mass_density : openff.units.Quantity, optional
         Target mass density for final system with units compatible with g / mL.
          If `None`, `box_size` must be provided.
-    box_aspect_ratio: list of float, optional
-        The aspect ratio of the simulation box, used in conjunction with
-        the `mass_density` parameter.
+    box_shape: NDArray
+        The shape of the simulation box, used in conjunction with the
+        `mass_density` parameter. Should have shape (3, 3) with all positive
+        elements.
 
     """
-    if box_size is None and mass_density is None:
+    if box_vectors is None and mass_density is None:
         raise PACKMOLValueError(
-            "Either a `box_size` or `mass_density` must be specified.",
+            "Either a `box_vectors` or `mass_density` must be specified.",
         )
 
-    if box_size is not None and len(box_size) != 3:
+    if box_vectors is not None and box_vectors.shape != (3, 3):
         raise PACKMOLValueError(
-            "`box_size` must be a openff.units.unit.Quantity wrapped list of length 3",
+            "`box_vectors` must be a openff.units.unit.Quantity Array with shape (3, 3)",
         )
 
-    if box_aspect_ratio is not None:
-        assert len(box_aspect_ratio) == 3
-        assert all(x > 0.0 for x in box_aspect_ratio)
+    assert box_shape.shape == (3, 3)
+    assert np.all(np.linalg.norm(box_shape, axis=-1) > 0.0)
 
     if len(molecules) != len(number_of_copies):
         raise PACKMOLValueError(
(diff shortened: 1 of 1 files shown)


Yoshanuikabundi:
Ok, I think I've fixed the one failing example - I've moved it to a rhombic dodecahedral box (just to show off that we can) and fixed up some prose that I think was out of date. I've also switched the `# NBVAL_SKIP` comments to tags so that users don't see them - see https://nbval.readthedocs.io/en/latest/#Using-tags-instead-of-comments. I think this is ready for review (but I thought that before so I'll check back when the examples CI finishes)

mattwthompson:
Hmm ...

image

mattwthompson:
This seems a little goofy

image

mattwthompson:
If I naively apply a cube root in `_box_from_density`, I get some reasonable-looking box vectors:


image

This fails to map out to OpenMM, however:

```python
ValueError                                Traceback (most recent call last)
Cell In[10], line 1
----> 1 topology.to_openmm()

File ~/mambaforge/envs/openff-interchange-env/lib/python3.10/site-packages/openff/utilities/utilities.py:80, in requires_package.<locals>.inner_decorator.<locals>.wrapper(*args, **kwargs)
     77 except Exception as e:
     78     raise e
---> 80 return function(*args, **kwargs)

File ~/mambaforge/envs/openff-interchange-env/lib/python3.10/site-packages/openff/toolkit/topology/topology.py:1911, in Topology.to_openmm(self, ensure_unique_atom_names)
   1908 if off_topology.box_vectors is not None:
   1909     from openff.units.openmm import to_openmm
-> 1911     omm_topology.setPeriodicBoxVectors(to_openmm(off_topology.box_vectors))
   1912 return omm_topology

File ~/mambaforge/envs/openff-interchange-env/lib/python3.10/site-packages/openmm/app/topology.py:252, in Topology.setPeriodicBoxVectors(self, vectors)
    250         raise ValueError("Second periodic box vector must be in the x-y plane.");
    251     if vectors[0][0] <= 0*nanometers or vectors[1][1] <= 0*nanometers or vectors[2][2] <= 0*nanometers or vectors[0][0] < 2*abs(vectors[1][0]) or vectors[0][0] < 2*abs(vectors[2][0]) or vectors[1][1] < 2*abs(vectors[2][1]):
--> 252         raise ValueError("Periodic box vectors must be in reduced form.");
    253 self._periodicBoxVectors = deepcopy(vectors)

ValueError: Periodic box vectors must be in reduced form.
```

I'm not sure why this is choking, but I think it's reasonable to expect the toolkit to handle it (https://github.com/openforcefield/openff-toolkit/issues/1645)

This was somewhat confusing to me before I noticed the default behavior of `pack_box` was changed away from using a cubic cell. To make the example run, I'll just tell it to use a cube.

mattwthompson (commented):
Everything here seems good; I won't claim to grok every detail but most things seem to be working nicely. I'm inclined to merge this more or less as-is, though I will sleep on it. Merging #731 doesn't seem to have broken things. 

Two issues I've come across are
* `examples/packed_box/packed_box.ipynb` is stalling in the GROMACS evaluation. I poked around a bit and found nothing obvious - could you have a look?
* The Amber and LAMMPS exports don't (explicitly) know about non-rectangular boxes. Could you either implement this or (preferred) error out and raise stub issue?

Yoshanuikabundi:
While testing `_compute_brick_from_box_vectors` I found that it produced brick representations that were not axis-aligned in some circumstances. Instead of rotating every atom in the box, I've just required that box vectors be in OpenMM reduced form - this simplifies the rest of the wrapping code too.

Yoshanuikabundi:
I also had to create a function that has a lot of `and` statements:

```python
    return (
        [ay, az] == [0, 0]
        and bz == 0
        and ax > 0
        and by > 0
        and cz > 0
        and ax >= 2 * np.abs(bx)
        and ax >= 2 * np.abs(cx)
        and by >= 2 * np.abs(cy)
    )
```

Black was formatting this like so, but flake8 didn't like it coz it still prescribes to the old pep8 where the `and` is supposed to go on the right hand sign. I've told flake8 to ignore the relevant warning, and everything works on my machine, but CI is not happy :/

Yoshanuikabundi:
> This fails to map out to OpenMM, however:

This is just because of the naive cube root - it works now that we're only cube rooting the volume/volume fraction (and not the box vectors).

> This was somewhat confusing to me before I noticed the default behavior of pack_box was changed away from using a cubic cell. To make the example run, I'll just tell it to use a cube.

I haven't changed this back (even though it now works with triclinic boxes)

Yoshanuikabundi:
> examples/packed_box/packed_box.ipynb is stalling in the GROMACS evaluation. I poked around a bit and found nothing obvious - could you have a look?

I'm getting...

```
-------------------------------------------------------
Program:     gmx grompp, version 2021.3-bioconda
Source file: src/gromacs/fileio/groio.cpp (line 230)

Fatal error:
Something is wrong in the coordinate formatting of file gmx.gro. Note that gro
is fixed format (see the manual)

For more information and tips for troubleshooting, please check the GROMACS
website at http://www.gromacs.org/Documentation/Errors
-------------------------------------------------------
```

I get this with either a rectangular or dodecahedral box :(

Looks like some of the generated residue names are 6 characters when they should be 5, and that's throwing the rest of the row out of whack:

```
    1MOL0    Cl  996   0.312   5.422   3.135
    1MOL0     C  997   0.305   5.592   3.084
    1MOL0    Cl  998   0.168   5.616   2.973
    1MOL0    Cl  999   0.284   5.696   3.228
    1MOL0    Cl 1000   0.457   5.636   3.002
  201MOL200    C 1001   1.897   5.028   1.310
  201MOL200    C 1002   1.807   5.137   1.358
  201MOL200    C 1003   1.671   5.133   1.326
  201MOL200    C 1004   1.586   5.234   1.370
  201MOL200    C 1005   1.637   5.340   1.446
```

Sounds like a bug in the `to_gro` code that was previously being hidden by `pack_box` generating residue names?

Yoshanuikabundi:
>  If you want to write such a file in your own program without using the GROMACS libraries you can use the following formats:
>
>C format
>    "%5d%-5s%5s%5d%8.3f%8.3f%8.3f%8.4f%8.4f%8.4f" 

...

```python
>>> "%-5s" % "123456"
'123456'
```

Thanks GROMACS.

I fixed this by trimming the residue name on the rhs of the % operator, and I've done the same for the atom name (which is the only other string in the fixed format section).

I can now run a single point energy with GROMACS but it's still not showing up in the `get_summary_data` table.

Yoshanuikabundi:
```
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

openff/interchange/interop/gromacs/export/_export.py:275:13: B028 No explicit stacklevel keyword argument found. The warn method from the warnings module uses a stacklevel of 1 by default. This will only show a stack trace for the line on which the warn method is called. It is therefore recommended to use a stacklevel of 2 or greater to provide more information to the user.
openff/interchange/interop/gromacs/export/_export.py:321:13: B028 No explicit stacklevel keyword argument found. The warn method from the warnings module uses a stacklevel of 1 by default. This will only show a stack trace for the line on which the warn method is called. It is therefore recommended to use a stacklevel of 2 or greater to provide more information to the user.
```

Whatever you say boss

Yoshanuikabundi:
> The Amber and LAMMPS exports don't (explicitly) know about non-rectangular boxes. Could you either implement this or (preferred) error out and raise stub issue?

Done: #752, #753, [298348b](https://github.com/openforcefield/openff-interchange/pull/707/commits/298348b0d47aba8188b455ccaa3114aea820c390), [d8e4c47](https://github.com/openforcefield/openff-interchange/pull/707/commits/d8e4c47be7143142c3cbf9aa99a411f61954c94f)

openff/interchange/_tests/unit_tests/components/test_packmol.py
```diff
@@ -225,18 +307,49 @@ def test_solvate_structure():
...
-        structure_to_solvate=benzene.to_topology(),
+        box_vectors=50 * numpy.identity(3) * unit.angstrom,
+        solute=benzene.to_topology(),
     )
 
     assert topology.n_molecules == 11
     assert len([*topology.unique_molecules]) == 2
+
+
+def solvate_topology():
```
mattwthompson, line 327: The coverage report is complaining about `solvate_topology` not being covered, yet I recall writing a test to do it.

🤦

mattwthompson:
[`flake8-bugbear`](https://pypi.org/project/flake8-bugbear/) borders on over-opinionated but its value so far has outweighed its nuisances

Issue #713, opened 2023-05-08
`Interchange.to_openmm_simulation()`


Yoshanuikabundi:
Don't worry about this until after the meeting, just wanted to put this down so I don't forget.

**Description**
In https://github.com/openforcefield/openff-docs/pull/5 we discussed an example that goes from a SMILES string to a prepared set of OpenMM system/topology/positions objects, and how it would be nice to be able to actually run the simulation as a part of this. I was wondering if it would be possible to one day include an `Interchange.to_openmm_simulation()` method, which would take an `openmm.Integrator` as an argument and produce an OpenMM `Simulation`, ready to go. So a bare-bones SMILES-to-simulation would go from:

```python
from openff.toolkit import Molecule, Topology, ForceField
from openff.interchange import Interchange
import openmm

paracetamol = Molecule.from_smiles("CC(=O)NC1=CC=C(C=C1)O")
paracetamol.generate_conformers(n_conformers=1)

topology = Topology.from_molecules([paracetamol])

force_field = ForceField("openff-2.0.0.offxml")

interchange = Interchange.from_smirnoff(force_field, topology)

openmm_system = interchange.to_openmm()
openmm_topology = interchange.to_openmm_topology()
openmm_positions = interchange.positions.to_openmm()

simulation = openmm.app.simulation.Simulation(
    openmm_topology, 
    openmm_system, 
    openmm.VerletIntegrator(2.0 * openmm.unit.femtosecond),
)
simulation.context.setPositions(openmm_positions)
simulation.minimizeEnergy()
simulation.context.getState(getEnergy=True).getPotentialEnergy()
```

to 

```python
from openff.toolkit import Molecule, Topology, ForceField
from openff.interchange import Interchange
import openmm

paracetamol = Molecule.from_smiles("CC(=O)NC1=CC=C(C=C1)O")
paracetamol.generate_conformers(n_conformers=1)

topology = Topology.from_molecules([paracetamol])

force_field = ForceField("openff-2.0.0.offxml")

interchange = Interchange.from_smirnoff(force_field, topology)

simulation = interchange.to_openmm_simulation(
    openmm.VerletIntegrator(2.0 * openmm.unit.femtosecond),
)
simulation.minimizeEnergy()
simulation.context.getState(getEnergy=True).getPotentialEnergy()
```

[...]

PR #725, opened 2023-05-23
Add `Interchange.to_openmm_simulation`

mattwthompson:
### Description

Implements #713

### Checklist

- [ ] Ensure NPT example runs an NPT simulation
- [x] Add tests
- [x] Update examples
- [x] Lint
- [x] Update docstrings


openff/interchange/components/interchange.py
```diff
@@ -367,6 +385,59 @@ def to_openmm_topology(
             ensure_unique_atom_names=ensure_unique_atom_names,
         )
 
+    def to_openmm_simulation(
+        self,
+        integrator: "openmm.Integrator",
+        combine_nonbonded_forces: bool = True,
+        add_constrained_forces: bool = False,
+        **kwargs,
+    ) -> "openmm.app.Simulation":
```
Yoshanuikabundi, line 394:
```suggestion
    ) -> "openmm.app.simulation.Simulation":
```

This is the magic incantation needed to let Sphinx find the documentation link :/

openff/interchange/components/interchange.py
```diff
@@ -367,6 +385,59 @@ def to_openmm_topology(
...
+        integrator : subclass of openmm.Integrator
+            The integrator to use for the simulation.
+        combine_nonbonded_forces : bool, default=False
+            If True, an attempt will be made to combine all non-bonded interactions into a single
+            openmm.NonbondedForce.
+            If False, non-bonded interactions will be split across multiple forces.
+        add_constrained_forces : bool, default=False,
+            If True, add valence forces that might be overridden by constraints, i.e. call `addBond` or `addAngle`
+            on a bond or angle that is fully constrained.
+
```
Yoshanuikabundi, line 411:
```suggestion
            on a bond or angle that is fully constrained.
        **kwargs
            Further keyword parameters are passed on to :py:meth:`Simulation.__init__() <openmm.app.simulation.Simulation.__init__>`

```

openff/interchange/components/interchange.py
Yoshanuikabundi, line 394: Does it also make sense to have something like an `additional_forces` argument? It could take user-created Force objects and add them to the system before the simulation is created. Would be useful for forces that implement pressure coupling etc. Might be more discoverable than `simulation.system.addForce`?

examples/ligand_in_water/ligand_in_water.ipynb
Yoshanuikabundi, line 1: Is 

```python
simulation.context.getSystem().addForce(barostat)
```
equivalent to
```python
simulation.system.addForce(barostat)
```
? I'm not sure coz I feel like I've run into problems with it, but maybe I was having problems adding forces after doing an energy minimization or something. If it is, it might be a simpler way to write that.

Yoshanuikabundi (approved):
Beautiful: 

```python

from openmm import VerletIntegrator

paracetamol = Molecule.from_smiles("CC(=O)NC1=CC=C(C=C1)O")
paracetamol.generate_conformers(n_conformers=1)

topology = Topology.from_molecules([paracetamol])

interchange = Interchange.from_smirnoff(force_field, topology)

simulation = interchange.to_openmm_simulation(VerletIntegrator(0.02))

print(simulation.context.getState(getEnergy=True).getPotentialEnergy())
simulation.minimizeEnergy()
print(simulation.context.getState(getEnergy=True).getPotentialEnergy())
```

I just have a few suggestions, mostly to documentation, but this looks great.

openff/interchange/components/interchange.py
```diff
@@ -367,6 +385,59 @@ def to_openmm_topology(
...
+        Returns
+        -------
+        simulation : openmm.app.Simulation
+            The OpenMM simulation object, possibly with positions set.
+
+        """
```
Yoshanuikabundi, line 417:
```suggestion
        Returns
        -------
        simulation : openmm.app.Simulation
            The OpenMM simulation object, possibly with positions set.

        Examples
        --------
        
        Create an OpenMM simulation with a Langevin integrator:
        
        >>> import openmm
        >>>
        >>> simulation = interchange.to_openmm_system(
        ...     openmm.LangevinMiddleIntegrator(
        ...         293.15 * openmm.unit.kelvin,
        ...         1.0 / openmm.unit.picosecond,
        ...         2.0 * openmm.unit.femtosecond,
        ...     )
        ... )
        
        Add a barostat:
        
        >>> simulation.system.addForce(
        ...     openmm.MonteCarloBarostat(
        ...         1.00 * openmm.unit.bar, 
        ...         293.15 * openmm.unit.kelvin, 
        ...         25,
        ...     )
        ... )
        
        """
```

If we stick with the current signature, it would be nice to include examples here.

examples/ligand_in_water/ligand_in_water.ipynb
mattwthompson: Not having tested it, I strongly suspect they're similar. The shorter version is more readable, I think, since somebody forgot exactly what's contained within each object

openff/interchange/components/interchange.py
mattwthompson: I'm neutral on this; I see the benefit to the user experience in making the `Simulation` object in one go, but in principle it opens the door to potential post-Interchange customization sneaking into the Interchange step

openff/interchange/components/interchange.py
mattwthompson: My favorite docexamples are the ones that other people write

examples/ligand_in_water/ligand_in_water.ipynb
Yoshanuikabundi: Yeah I had a similar experience doing the micelle self assembly vignette. Density should definitely change in an NPT simulation. If there's no way to add forces after creating a simulation, then I think that `additional_forces` argument has a lot of value.

Yoshanuikabundi: Yeah it's in the OpenMM FAQ: https://github.com/openmm/openmm/wiki/Frequently-Asked-Questions#why-does-it-ignore-changes-i-make-to-a-system-or-force

Apparently once a Context is created you can't change the System unless you call `context.reinitialize()`. So I guess setting the positions in `to_openmm_simulation` makes it impossible to add forces, or a context is created in `Simulation.__init__`. So maybe if after adding forces we do a `system.context.reinitialize(preserveState=True)`?

Yoshanuikabundi: [Issue](https://github.com/openmm/openmm/issues/3759) discussing the topic. I think we should add that call to `reinitialize` to both the notebook and the docstring example, or add the `additional_forces` argument.

openff/interchange/components/interchange.py
mattwthompson: https://github.com/openforcefield/openff-interchange/issues/730

examples/ligand_in_water/ligand_in_water.ipynb
mattwthompson: Makes sense from the perspective of the implementation, though a bit of a sharp edge. Thanks for finding this!

d6c0d9dc893dbcd1fb2175da7aa512d7363047b8

mattwthompson:
Thanks for the review and iterations @Yoshanuikabundi!

PR #731, opened 2023-05-26
Export non-rectangular boxes to GROMACS

mattwthompson:
### Description

This PR allows for the export of non-rectangular boxes to GROMACS.

https://manual.gromacs.org/archive/5.0.3/online/gro.html

As far as I can tell, the implementation is simple because GROMACS wants box vectors in a similar structure to how they're stored in OpenMM style; no converting out to lengths and angles is necessary.

### Checklist

- [x] Add tests, ensuring the box vectors loaded from GROMACS files by OpenMM match what are read from PDB files
  - [x] Cubic box
  - [x] Rhombic dodecahedron
  - [x] Truncated octahedron
- [x] Lint
- [x] Update docstrings


Yoshanuikabundi:
GROMACS supports triclinic boxes of any shape, but it does *not* support arbitrary triclinic boxes - they have to be oriented a particular way and satisfy some inequalities that you can always get to by manipulating box vectors. See [section 3.2](https://manual.gromacs.org/current/reference-manual/algorithms/periodic-boundary-conditions.html) of the reference manual, equations 10-12. Looks like these conditions are [shared](http://docs.openmm.org/latest/userguide/theory/05_other_features.html#periodic-boundary-conditions) by OpenMM, so maybe you handle them somewhere else, and maybe they're universal enough that they can be the user's job to figure out, but I thought I'd make sure you know.

Everything looks good, I'm just trying to install GROMACS to double check that there are no surprises! Will update you when I've done that.

Yoshanuikabundi:
Also - GROMACS actually does support [taking starting coordinates for a simulation](https://manual.gromacs.org/current/onlinehelp/gmx-grompp.html) from a PDB file, it's just that a GRO file can include velocities and have higher precision positions. I used to do all my GROMACS prep with PDB because it's so much easier to visualize. So `to_gro` is definitely nice to have but people can struggle along without it. Sorry, that probably would've been more useful information when you started writing it...

mattwthompson:
This is all very useful information, thanks!

> maybe you handle them somewhere else

Nope! It's convenient for this scope that I can take OpenMM-style box vectors and get whatever GROMACS needs, but the validator for `Interchange.box` doesn't include any domain-specific checks.

This implies we should add a test to make sure `to_pdb` doesn't mangle this, even though the actual work it done by OpenMM. It seems to now:

```python
In [16]: force_field = ForceField("openff-2.0.0.offxml")

In [17]: topology = Topology.from_pdb("openff/interchange/_tests/data/dodecahedron.pdb")

In [18]: force_field.create_interchange(topology).to_pdb("test.pdb")

In [19]: !head test.pdb
REMARK   1 CREATED WITH OPENMM 8.0, 2023-05-30
CRYST1   30.000   30.000   30.000 120.00 120.00  90.00 P 1           1
HETATM    1  H3  ACE A   1      -5.284  -2.981   5.320  1.00  0.00           H
HETATM    2  CH3 ACE A   1      -4.851  -3.877   5.426  1.00  0.00           C
HETATM    3  H1  ACE A   1      -5.412  -4.481   6.144  1.00  0.00           H
HETATM    4  H2  ACE A   1      -4.840  -4.406   4.474  1.00  0.00           H
HETATM    5  C   ACE A   1      -3.431  -3.679   5.966  1.00  0.00           C
HETATM    6  O   ACE A   1      -2.705  -2.809   5.481  1.00  0.00           O
ATOM      7  N   ALA A   2      -2.983  -4.446   6.956  1.00  0.00           N
ATOM      8  CA  ALA A   2      -1.682  -4.415   7.599  1.00  0.00           C

**Comment by mattwthompson** [comment]
I can add checks similar to what OpenMM lays out [here](http://docs.openmm.org/latest/userguide/theory/05_other_features.html#periodic-boundary-conditions) if desired

**Comment by Yoshanuikabundi** [comment]
> I can add checks similar to what OpenMM lays out [here](http://docs.openmm.org/latest/userguide/theory/05_other_features.html#periodic-boundary-conditions) if desired

I think that would be a very reasonable invariant to establish for the entire OpenFF stack (CC @j-wags). I don't think it would inconvenience many users as many MD engines require them already, and it would greatly simplify any direct processing of box vectors (as in #707, for example). Maybe it would be problematic for loading PDB files with crystal records that are from actual crystals? We could probably write a function to automatically put box vectors in this format rather than just rejecting them if they're not? That would involve rotating coordinates themselves in the general case though.

I'll locally test and review this today - hopefully by the time you see this.

**Comment by mattwthompson** [comment]
> Maybe it would be problematic for loading PDB files with crystal records that are from actual crystals? 

I'm not sure how intentionally this use case is really supported

File: openff/interchange/_tests/data/README.md [hunk 1]
```diff
@@ -72,3 +72,20 @@ struct.save('packed-argon.pdb')
 `complex.gro`
 
 - [Source](https://raw.githubusercontent.com/samplchallenges/SAMPL6/master/host_guest/SAMPLing/CB8-G3-0/GROMACS/complex.gro)
+
+- `cube.pdb`
+- `dodecahedron.pdb`
+- `octahedron,pdb`
```
Yoshanuikabundi, line 78:
```suggestion
- `octahedron.pdb`
```

Yoshanuikabundi (approved):
Looks great! Found one typo in a readme. I tried preparing a dodecahedral box in gromacs, importing it with `Topology.from_pdb`, and then exporting it and running GROMPP and it worked beautifully. The PBC lines of the GROs produced by Interchange and GROMACS are similar but not identical:

GROMACS:
```
   9.40650   9.40650   6.65140   0.00000   0.00000   0.00000   0.00000   4.70325   4.70325
```

Interchange:
```
  9.4065000  9.4065000  6.6513999  0.0000000  0.0000000  0.0000000  0.0000000 -4.7032500 -4.7032500
```

I'm not sure what to make of the minus signs. They seem to tessellate correctly in VMD, just with the sign swapped in the Y and Z directions, so I guess this is a degeneracy in the PDB length-angle representation?

Transposing the box and re-exporting, GROMPP gives a clear error:

```
ERROR 1 [file from_interchange.top, line 63319]:
  Only triclinic boxes with the first vector parallel to the x-axis and the
  second vector in the xy-plane are supported.
```

So I don't think we necessarily need to worry about checking for that here, at least in this PR.

I've also checked that the relevant docstrings are still correct and they are. I think this is good to merge!

mattwthompson:
Thanks for the review - any guesses if the vectors with negatives signs are correct? If so we should open a stub issue for the paper trail. Not gonna let that holds us back, though

PR #754, opened 2023-06-21
Fix protein_ligand example's call to pack_box

Yoshanuikabundi:
### Description

The `protein_ligand` example is broken on 0.3.6 - it uses the old argument names for `pack_box`.

This PR gets everything to run without errors, but at least for me the summary data tables were not displaying.

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/examples/protein_ligand/protein_ligand.ipynb
+++ b/examples/protein_ligand/protein_ligand.ipynb
@@ -55,7 +55,7 @@
     "from openff.units import unit\n",
     "\n",
     "from openff.interchange import Interchange\n",
-    "from openff.interchange.components._packmol import pack_box\n",
+    "from openff.interchange.components._packmol import UNIT_CUBE, pack_box\n",
     "from openff.interchange.drivers import (\n",
     "    get_amber_energies,\n",
     "    get_gromacs_energies,\n",
@@ -389,7 +389,7 @@
     "xyz = protein.conformers[0]\n",
     "centroid = xyz.sum(axis=0) / xyz.shape[0]\n",
     "protein_radius = np.sqrt(((xyz - centroid) ** 2).sum(axis=-1).max())\n",
-    "box_size = np.ones(3) * protein_radius * 2 + 2 * unit.nanometer\n",
+    "box_vectors = UNIT_CUBE * (protein_radius * 2 + 2 * unit.nanometer)\n",
     "\n",
     "# Pack the box with an arbitrary number of water\n",
     "n_water = 1000\n",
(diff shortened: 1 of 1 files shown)


Yoshanuikabundi:
Looks like its [ignored in CI](https://github.com/openforcefield/openff-interchange/blob/4ccc56300e04613e4fb2f767b3d313b29cd34a2e/.github/workflows/examples.yaml#L73) - probably because it's relatively slow?

mattwthompson:
Well that's probably the case. Hard to find stuff when it's right under your nose sometimes.

> probably because it's relatively slow

Probably, yeah. It's probably tinker-able a little bit, and the runtime could also be mitigated by running `nbval` in parallel like I believe I've seen you do. (I'll give that a go next week if you don't beat me to it, and you probably have better things to do anyway)

Yoshanuikabundi:
GitHub Actions' runners only have a single core available - multithreading things might still help a bit if you oversubscribe cores to make sure the CPU is always flat out, but you're not going to get crazy speedups and it could just as easily mess with the cache and slow things down. The reason I implemented it was so that I could run things fast *locally* on my 24-thread computer.

Having said that, something weird is definitely going on - I can run every notebook from both the Toolkit and Interchange in about 20 minutes without OpenEye in the Openff-Docs action: https://github.com/openforcefield/openff-docs/actions/runs/5332778351/jobs/9662411769

That seems to be running sequentially from the output (each notebook is done before the next notebook starts), which aligns with the code (multithreaded over 1 thread/available core * 1 core = 1 thread == sequential).

Just the Toolkit notebooks take longer than that in Toolkit CI, and Interchange CI is an extra 10 minutes, and Interchange skips notebooks that I run. Adding those times together double counts Mamba time... but that only accounts for 2 minutes. There are a couple of notebooks running on Toolkit CI that I skip, but I don't think that's enough to make the difference either. Jeff suggested that it might be GitHub giving me faster runners coz I only have 1 job to run.

:shrug:

mattwthompson:
I thought we get [two cores](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) these days and I'm pretty sure `pytest-xdist` discovers two cores when I chuck it `-nauto`. Not world-breaking but I remember it dropping toolkit runtime by a few minutes when I added it there.

In general I'm also confused, and don't feel like optimizing this now. But it should be done at some point.

mattwthompson (approved):
Works as advertised - thanks for the fix

https://github.com/openforcefield/openff-interchange/actions/runs/5383089128

Yoshanuikabundi:
> I thought we get [two cores](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) these days

Oh cool when did that happen. Taking another look at the examples run it looks like the different threads might be stepping on each others output because not all the notebooks are there - which might explain why it looks serial.

Issue #765, opened 2023-07-11
gromacs._parse_gmx_energy has hard-coded path to out.edr file


pbuslaev:
Funcition **`_parse_gmx_energy`** in [drivers/gromacs](https://github.com/openforcefield/openff-interchange/blob/v0.3.7/openff/interchange/drivers/gromacs.py) is reading energy data from hard-coded `out.edr` file:

```python
@requires_package("panedr")
def _parse_gmx_energy(edr_path: str) -> dict[str, unit.Quantity]:
    """Parse an `.edr` file written by `gmx energy`."""
    import panedr

    parsed_energies = panedr.edr_to_df("out.edr").to_dict("index")[0.0]
    parsed_energies.pop("Time")
```

It should be solved by passing `*edr_path*` to `**panedr.edr_to_df**`

mattwthompson:
Heh, nice find. Was probably never a problem since that variable is only ever set to `"out.edr"` in my usage. It wouldn't surprise me if there's other similar cases in the codebase.

PR #766, opened 2023-07-11
Update gromacs.py to pass input edr path in _parse_gmx_energies routine

pbuslaev:
This is a fix to #765


mattwthompson (approved):
LGTM - thanks for the catch. CI will not completely pass since we don't expose the OpenEye license on forks, but I'll merge once everything passes as I expect to happen.

PR #803, opened 2023-09-07
Add TIP4P-FB with GROMACS in ligand example

### LINKED ISSUE #587: Export openMM system with tip4pew water to GROMACS top/gro
I am trying to use ParmEd v3.4.3 to convert a system (with a "sage" ligand) solvated in tip4pew water from openMM v. 7.7.0 to GROMACS topology and configuration files (gro). Python lines doing the conversion are reported below (I can provide the full script with generation of openMM system + sage ligand + solvation if needed). Both openMM "export_system" and "pmd_complex_struct" seem to correctly include the tip4pew virtual sites. However, the generated GROMACS topology file does not contain the virtual site and the generation of the "gro" file is interrupted with this error

"RuntimeError: Could not find <Atom O [0]; In HOH 0>"

and no virtual sites are present in the "gro" file written before the error. Any tips?

### ParmEd's GROMACS exporter can't handle constraints from openmm, so we need a variant for export without them
export_system = omm_forcefield.createSystem(
modeller.topology,
nonbondedMethod=openmm.app.PME,
constraints=None,
rigidWater=False,
)

### Combine the topology, system and positions into a ParmEd Structure
pmd_complex_struct = pmd.openmm.load_topology(topology, export_system, positions)

### Export GROMACS files
pmd_complex_struct.save("system.top", overwrite=True)
pmd_complex_struct.save("system.gro", overwrite=True)

mattwthompson:
### Description

Resolves #587

### Checklist

- [ ] Add tests
- [x] Lint
- [ ] Update docstrings


j-wags (approved):
Looks good! I'm a bonehead and forgot to have gromacs installed, but it ran beautifully after that. The OPC water geometry looked funny to me, but I think it was just an illusion - I checked the [source paper](https://pubs.acs.org/doi/10.1021/jz501780a) and the vsite placement looks fine!

Issue #810, opened 2023-09-12
OpenMM v-site ordering does not match ``virtual_site_key_topology_index_map``


SimonBoothroyd:
**Description**

When going from an interchange containing two water molecules, each with a v-site, to OpenMM, the v-sites are placed after each molecule (i.e. with indices of 3, 7). The interchange object (and particularly the `virtual_site_key_topology_index_map` attribute of the v-site handler) state they should have indices 6, and 7.

**Reproduction**

```
molecule_a = openff.toolkit.Molecule.from_smiles("O")
molecule_a.generate_conformers(n_conformers=1)
molecule_b = openff.toolkit.Molecule.from_smiles("O")
molecule_b.generate_conformers(n_conformers=1)

topology = openff.toolkit.Topology()
topology.add_molecule(molecule_a)
topology.add_molecule(molecule_b)

interchange = openff.interchange.Interchange.from_smirnoff(
    openff.toolkit.ForceField("tip4p_fb.offxml"), topology
)
```

**Output**

Please include the output, including full tracebacks of any errors, resulting from the reproduction.

**Software versions**

- Which operating system and version are you using? OSX
- How did you install Interchange? conda
- What is the output of running `conda list`? `openff-interchange-base   0.3.14`

j-wags:
Thanks for the detailed reports, @SimonBoothroyd - Matt's out this week but he'll be able to take a look when he returns!

mattwthompson:
Wow - nice find! It looks like the indexing switched to per-molecule collated at the OpenMM layer. All of the checks were for consistency within the OpenMM layer, so no alarm bells were sounded.

mattwthompson:
I think #816 changed this to be what you expect - here's what I understand that to mean

```python
In [1]: import openff.toolkit
   ...: import openff.interchange

In [2]: molecule_a = openff.toolkit.Molecule.from_smiles("O")
   ...: molecule_a.generate_conformers(n_conformers=1)
   ...: molecule_b = openff.toolkit.Molecule.from_smiles("O")
   ...: molecule_b.generate_conformers(n_conformers=1)
   ...:
   ...: topology = openff.toolkit.Topology()
   ...: topology.add_molecule(molecule_a)
   ...: topology.add_molecule(molecule_b)
   ...:
   ...: interchange = openff.interchange.Interchange.from_smirnoff(
   ...:     openff.toolkit.ForceField("tip4p_fb.offxml"), topology
   ...: )

[...]

PR #816, opened 2023-09-18
Fix virtual site indexing in OpenMM

mattwthompson:
### Description

#810 

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -93,7 +93,8 @@ jobs:
     - name: Install AmberTools
       if: ${{ matrix.openeye == false }}
       run: |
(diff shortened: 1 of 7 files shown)


mattwthompson:
Bummer, GROMACS and OpenMM need different indexing.

mattwthompson:
Might not get to this for a few days, but as a note to myself: I think this can be made to work by having `_build_particle_map` operate in two modes, one that throws all virtual sites at the end of the system (OpenMM style) and another that collates them at the end of each molecule (GROMACS style). I hope this covers what other engines/tools need -  maybe there's another mode that does some funky accounting with residues, but that's a bridge to cross when we get there.

PR #825, opened 2023-10-09
Update host-guest example

mattwthompson:
### Description

Derived from #326

### Checklist

- [ ] Add tests
- [x] Lint
- [x] Update docstrings


mattwthompson:
I could use an edit pass on this if @j-wags has time or if we get @Yoshanuikabundi back - not going to let it diverge in a branch, though

PR #830, opened 2023-10-19
Update versioneer

Yoshanuikabundi:
### Description

Provide a brief description of the PR's purpose here.

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

PR #834, opened 2023-10-21
Update versioneer

mattwthompson:
### Description

#830 with an added detail to properly interact with VCS

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


setup.cfg
```diff
@@ -48,7 +48,7 @@ VCS = git
 style = pep440
 versionfile_source = openff/interchange/_version.py
 versionfile_build = openff/interchange/_version.py
-tag_prefix = ''
+tag_prefix = v
```
mattwthompson, line 51: Not having this was the cause of my problems

mattwthompson: I think, without fully demonstrating, that this is necessary when some tags don't start with `v` but the recent ones do

PR #838, opened 2023-11-03
fixed the zenodo links 

jameseastwood:
...to refer to 0.3.17 and the all-versions doi, respectively.

### Description

The zenodo links in the readme were linking to an old release. This PR fixes them, so that the first one links the current (Nov 3, 2023) release and the second one links the evergreen latest release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Documentation for Interchange, including examples, a brief user guide, release h
 
 ## How to cite
 
-Please cite Interchange using the [Zenodo record](https://zenodo.org/record/8147765) of the [latest release](https://zenodo.org/record/8147765) or the version that was used. The BibTeX reference of the latest release can be found [here](https://zenodo.org/record/8147765/export/hx).
+Please cite Interchange using the [Zenodo record](https://doi.org/10.5281/zenodo.10068101) of the [latest release](https://zenodo.org/doi/10.5281/zenodo.8147764) or the version that was used. The BibTeX reference of the latest release can be found [here](https://zenodo.org/record/8147765/export/hx).
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,7 +14,7 @@ Documentation for Interchange, including examples, a brief user guide, release h
 
 ## How to cite
 
-Please cite Interchange using the [Zenodo record](https://zenodo.org/record/8147765) of the [latest release](https://zenodo.org/record/8147765) or the version that was used. The BibTeX reference of the latest release can be found [here](https://zenodo.org/record/8147765/export/hx).
+Please cite Interchange using the [Zenodo record](https://doi.org/10.5281/zenodo.10068101) of the [latest release](https://zenodo.org/doi/10.5281/zenodo.8147764) or the version that was used. The BibTeX reference of the latest release can be found [here](https://zenodo.org/record/8147765/export/hx).
```
mattwthompson, line 17:
```suggestion
Please cite Interchange using the [Zenodo record](https://doi.org/10.5281/zenodo.10068101) of the [latest release](https://zenodo.org/doi/10.5281/zenodo.8147764) or the version that was used. The BibTeX reference of the latest release can be found [here](https://zenodo.org/record/8147764/export/hx).
```

mattwthompson (approved):
Thanks thanks - I should probably add this to the release process.

I wonder if there's a bot that could do this automatically? Fetch the new record identifier or whatever it's called, `sed` it against the old record, commit and push?

README.md
mattwthompson: Ah - so the `10068101` URL wouldn't need to be updated, but I think the `814....` would?

jameseastwood: That's right, one of them is evergreen -- it's the doi for the entire record, and always resolves to the latest version. 

Maybe the right answer is to ONLY include that link in the README, but the way it was written it had two links so I didn't want to change that without knowing more about why it was written that way.

PR #864, opened 2023-12-21
Update installation instructions

### LINKED ISSUE #862: Change installation instructions to recommend `mamba create` instead of `conda install`
A user recently wrote in that they had ended up with a borked environment, ultimately due to some path-dependence in how they installed Interchange (they created an env with other deps, and then conda-installed interchange). It may be good to update the docs to recommend `mamba create` as the default install pathway.

mattwthompson:
### Description

Resolves #862

The docs default to the `stable` branch so these won't be highly visible until a release is made in early January.

### Checklist

- [x] Update installation instructions
- [x] Recommend Mamba
- [ ] Add tests
- [x] Lint
- [ ] Update docstrings

### DIFF
--- a/docs/developing.md
+++ b/docs/developing.md
@@ -39,7 +39,7 @@ The configuration file (`.pre-commit-config.yaml`) is commited to the repo. This
 First, install `pre-commit`
 
 ```console
-$ conda install pre-commit -c conda-forge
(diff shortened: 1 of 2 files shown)


mattwthompson:
Passes the eye check: 
image

PR #869, opened 2024-01-05
Better support conversion of large topologies

mattwthompson:
### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
I think this is no longer needed after https://github.com/openforcefield/openff-toolkit/pull/1805/files

PR #880, opened 2024-01-31
Stage industry-facing virtual sites example

mattwthompson:
### Description

@lilyminium here's what I promised a while back. I'm not sure this this land in the repo long-term.

Is there a better way to present this than showing what XML files condense out from each SMILES string?


lilyminium (approved):
LGTM -- thanks @mattwthompson, this looks amazing! I especially really like the simulations in the notebook -- being able to watch those gives me a lot of confidence that the OpenFF <-> OpenMM conversion is exactly what we want it to be.

> Is there a better way to present this than showing what XML files condense out from each SMILES string?

If the intention is to give people an exact mapping from SMIRNOFF -> OpenMM, this seems like the safest way -- and I'd also be in favour of adding ITPs and possibly other formats, too, which might be more relevant to people who mostly use those formats.

mattwthompson:
I lost track of the wording, but yesterday @j-wags suggested it's better to have this merged in than not. I'm going to kick the tires on this and aim to merge it more or less as-is, shipping the notebook and XMLs.

mattwthompson:
Oh, it's actually already online in some form 🫠 https://docs.openforcefield.org/en/latest/examples/openforcefield/openff-interchange/virtual_sites/virtual_sites.html

mattwthompson:
A consequence of https://github.com/openforcefield/openff-interchange/pull/947 is that the water molecule with this OFFXML gets _both_ 4- and 5-site parameters. (That makes sense, there isn't otherwise a way to distinguish which water model should be applied.) Previously, only 4-site parameters were applied by nature of them showing up earlier in the force field.

I've removed the 5-site parameters from the notebook, which was already getting around this problem by loading 5-site water externally (`ForceField("tip5p.offxml")`).

PR #883, opened 2024-02-01
Interoperability trunk

mattwthompson:
### Description

Related to https://github.com/openforcefield/openff-toolkit/pull/1818

### Checklist

- [x] Fix #879
- [x] Fix #858
- [x] Preserve atom metadata in OpenMM roundtrip
- [x] Do not write `_tmp_pdb_file.pdb` to user's disk when visualizing
- [x] Fall back to `Topology.visualize` when virtual sites are not included
- [x] Lint
- [ ] Update docstrings

PR #885, opened 2024-02-02
Visualization improvements

mattwthompson:
### Description

- [x] Do not write `_tmp_pdb_file.pdb` to user's disk when visualizing
- [x] Fall back to `Topology.visualize` when virtual sites are not included

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
Now part of #883

PR #916, opened 2024-02-29
Improve performance of LAMMPS export

mattwthompson:
### Description

cc: @timbernat 

### Checklist

- [x] Fix bug in which `get_lammps_energies` crashes when no dihedrals are present
- [x] Remove unnecessary double-nested loops in
  - [x] Bonds section
  - [x] Angles section
  - [x] Propers section
  - [x] Impropers section
- [x] Un-skip some old tests I forgot existed
- [ ] Add tests
- [x] Lint
- [ ] Update docstrings


mattwthompson:
I still need to update the other valence sections, but this knocks down `interchange.to_lammps()` in this example from about 10 seconds to about 3 seconds.

https://docs.openforcefield.org/en/latest/examples/openforcefield/openff-interchange/lammps/lammps.html

Issue #917, opened 2024-03-03
Pass cosmetic attrs through to `Potential` objects


SimonBoothroyd:
**Description**

Currently cosmetic attributes on force field parameters do not propagate through to a `Potential` objects parameters. Currently we need to re-look them up from the original force field.

It would be great if these could be passed through.

SimonBoothroyd:
My current hacky workaround:

```python
def to_interchange(ff, top):
    inter = openff.interchange.Interchange.from_smirnoff(ff, top)

    for col_k, col_v in inter.collections.items():
        for pot_k, pot_v in col_v.potentials.items():
            if " " not in pot_k.id:
                handler = ff[pot_k.associated_handler]
                param = handler[pot_k.id]
            else:
                smirks, lp, match = pot_k.id.split(" ")
                handler = ff["VirtualSites"]
                (param,) = handler.get_parameter(
                    {"smirks": smirks, "name": lp, "match": match}
                )

            if not hasattr(param, "_cosmetic_attribs"):
                continue

            for attr in param._cosmetic_attribs:
                pot_v.parameters[attr] = getattr(param, f"_{attr}")

    return inter
```

mattwthompson:
@SimonBoothroyd Looking at this more closely now - does it make sense to have it on the `Potential` or `PotentialKey` object? Right now `Potential.parameters` really only stores the stuff that describes the physics of the interaction, whereas i.e. stuff like [`parameterize_eval`](https://github.com/openforcefield/openff-toolkit/blob/db5ddfb8d89d7b245c5078ab43ba22dd548e0497/openff/toolkit/_tests/test_forcefield.py#L169) shoved in for ForceBalance's accounting seem like things that better live on pointers to those objects. I'm not sure if such a physics vs. bookkeeping distinction holds for other use cases.

PR #918, opened 2024-03-04
Store cosmetic attributes

mattwthompson:
### Description

One way of doing #917, not sure it's the best approach for all use cases

### Checklist

- [x] Add tests
- [x] Lint
- [ ] Update docstrings


j-wags:
@SimonBoothroyd Just tagging you for visibility - Matt will merge in 5 days, so please take a glance and comment if you have time before then!

Issue #921, opened 2024-03-06
`Interchange.from_openmm` should raise an exception if the topology doesn't match the system


Yoshanuikabundi:
**Description**
I'm not sure how detailed this check /c|sh|w/ould be, but raising an exception if the Topology and System have different numbers of atoms would have saved me an hour or so last night - I had a lot of cryptic errors that all stemmed back to changing a variable name and accidentally still using the old name in one place. I imagine virtual sites might make this a pain.

Passing OpenFF topologies to `from_openmm` is otherwise a great experience! And this change would probably be useful for taking OpenMM topologies as well.

mattwthompson:
> I imagine virtual sites might make this a pain.

Your imagination is correct - but that's implicitly unsupported right now so it's worthwhile to make it explicitly unsupported.

How stringent a check are we talking here? The most naive check that comes to mind is checking that the number of particles in the system matches the number of atoms in the topology. (This could be massaged slightly to only consider massive particles, since virtual sites don't exist in OpenFF's topologies.) I imagine that should cover most cases of accidentally passing in the wrong variables.

I think connectivity checks would be trickier to implement (rigid water often doesn't have bonds in OpenMM models, for example) and risk some false negatives, so I don't think I'd go for that in the first pass here.

Yoshanuikabundi:
> How stringent a check are we talking here? The most naive check that comes to mind is checking that the number of particles in the system matches the number of atoms in the topology. (This could be massaged slightly to only consider massive particles, since virtual sites don't exist in OpenFF's topologies.) I imagine that should cover most cases of accidentally passing in the wrong variables.

That sounds great. `openmm.System` seems to have a `isVirtualSite(index)` method, which might be preferable to checking mass because OpenMM's recommended way to freeze atoms is to set their mass to zero.

I agree connectivity has all sorts of bugbears that are likely to get worse in the future!

PR #923, opened 2024-03-06
Raise errors when `from_openmm` inputs are incompatible or unsupported

mattwthompson:
### Description

#921 

### Checklist

- [x] Add tests
- [x] Lint
- [x] Update docstrings

### DIFF
--- a/openff/interchange/interop/openmm/_import/compat.py
+++ b/openff/interchange/interop/openmm/_import/compat.py
@@ -1,0 +1,35 @@
+from typing import Union
+
+import openmm
+import openmm.app
+from openff.toolkit import Topology
+
+from openff.interchange.exceptions import UnsupportedImportError
+
+
+def _check_compatible_inputs(
+    system: openmm.System,
+    topology: Union[openmm.app.Topology, Topology, None],
+):
+    """Check that inputs are compatible and supported."""
+    for index in range(system.getNumParticles()):
+        if system.isVirtualSite(index):
+            raise UnsupportedImportError(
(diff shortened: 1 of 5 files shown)


mattwthompson:
@Yoshanuikabundi any other compatibility cases I should be on the lookout for?

openff/interchange/interop/openmm/_import/compat.py
```diff
@@ -0,0 +1,35 @@
...
+    """Check that inputs are compatible and supported."""
+    for index in range(system.getNumParticles()):
+        if system.isVirtualSite(index):
+            raise UnsupportedImportError(
+                "A particle is a virtual site, which is not yet supported.",
+            )
+        if system.getParticleMass(index)._value == 0.0:
+            raise UnsupportedImportError(
+                "A particle has a mass of 0.0. Is this intentional?",
+            )
```
Yoshanuikabundi, line 23: I think this is likely to cause false positives if users are setting masses to zero to freeze atoms (which is the way OpenMM recommends). I also don't think it's necessary if we're explicitly declining any system with virtual sites!

Yoshanuikabundi (approved):
Looks good! I would just remove the zero mass check.

openff/interchange/interop/openmm/_import/compat.py
mattwthompson: I didn't know this is how atoms are frozen in OpenMM world ... nowhere else does this code handle frozen atoms but I suppose it's not obvious that it would need to?

PR #936, opened 2024-03-18
Adopt SPEC 0, update other upstreams

mattwthompson:
- [x] Drop Python 3.9
- [x] Drop Pydantic v1
- [x] Update developer docs


mattwthompson:
This got shot down

PR #878, opened 2024-01-30
Partially document current feature support

mattwthompson:
### Description

Provide a brief description of the PR's purpose here.

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

PR #954, opened 2024-04-03
Restore link to Toolkit docs

Yoshanuikabundi:
### Description

This fixes the link mentioned in https://github.com/openforcefield/openff-interchange/pull/878/files#r1471815719

Unfortunately I'm not sure why the original link ever worked. MyST went through a stage of frequently breaking links, but I thought we were still on a version of MyST prior to that.

This PR restores the link, but makes it point to the whole Toolkit developer page instead of the particular subsection. If we want to point to the subsection, we need a PR to the Toolkit to give the subsection a reference anchor so that it appears in the Intersphinx inventory.

### Checklist

- [x] Lint

### DIFF
--- a/docs/using/plugins.md
+++ b/docs/using/plugins.md
@@ -13,7 +13,7 @@ Currently, systems using custom interactions can only be exported to OpenMM. The
 
 ## Creating a custom `ParameterHandler`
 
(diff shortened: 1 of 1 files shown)


mattwthompson (approved):
It'd be easier to follow if it linked more directly to somewhere in the toolkit docs the stored the relevant information. But I don't want to climb that hill, and this fixes what I was originally confused about

PR #957, opened 2024-04-08
Constrain to `openff-nagl ==0.3.6`

mattwthompson:
### Description

Some funky stuff happened with 0.3.7; I don't doubt it'll be fixed soon, this just down-constrains to a slightly older version.

### DIFF
--- a/devtools/conda-envs/beta_env.yaml
+++ b/devtools/conda-envs/beta_env.yaml
@@ -12,7 +12,7 @@ dependencies:
   # OpenFF stack
   - openff-toolkit >=0.15.2
   - openff-models
-  - openff-nagl
+  - openff-nagl ~=0.3.7
   - openff-nagl-models
   # Optional features
   - jax
--- a/devtools/conda-envs/examples_env.yaml
+++ b/devtools/conda-envs/examples_env.yaml
@@ -9,10 +9,10 @@ dependencies:
   - pydantic =2
   - openmm
   # OpenFF stack
-  - openff-toolkit =0.15.2
+  - openff-toolkit ~=0.15.2
   - openff-models
-  - openff-nagl ==0.3
-  - openff-nagl-models ==0.1
+  - openff-nagl ==0.3.6
+  - openff-nagl-models =0.1
   - dgl =1
   # Optional features
   - unyt
@@ -25,7 +25,7 @@ dependencies:
   - pdbfixer
   - nglview
   - openeye-toolkits >=2023.2
-  - pytest =8.0
+  - pytest =8
   - pytest-xdist
   - nbval
   - openmmforcefields >=0.11.2


lilyminium:
Just to check Matt, what funky things were you seeing -- was it the failed pyarrow import? If so I don't think that would be necessarily resolved by downpinning nagl, but hopefully has been fixed upstream now: https://github.com/conda-forge/snappy-feedstock/issues/35#issuecomment-2043486477

Otherwise I can release 0.3.8 as I was planning to yesterday with snappy manually pinned.

mattwthompson:
I was getting a non-integer charge error where previously it was not erroring. That's the only detail I remember, let me find the log

mattwthompson:
Okay, comparing the logs here and #959, the difference is the `snappy` version after all. Still found a couple of improvements, though

lilyminium:
Thanks for looking into this Matt!

I'm surprised that the NAGL failure triggered the `NonIntegralMoleculeChargeError` -- if I'm reading the example correctly, `host` should have no charges assigned (as NAGL failed) and `guest` should have AM1BCC charges assigned. The molecule in question (which I'm guessing is the host) seems to have a formal charge of -24, which might be why the -8 charge raised an error. I'm curious where these charges are coming from at all.


```
Molecule c1c2cc3cc1Oc4cc(cc(c4)Oc5cc6cc(c5)Oc7cc(cc(c7)Oc8cc9cc(c8)Oc1cc(cc(c1)Oc1cc(cc(c1)Oc1cc(cc(c1)O3)[C-]([O-])[O-])C1Oc3cc4c5cc3C(c3cc7c(cc3O1)OC9Oc1cc3c(cc1C7CC[C-]([O-])[O-])C(c1cc(c(cc1OC6O3)OC2O4)C5CC[C-]([O-])[O-])CC[C-]([O-])[O-])CC[C-]([O-])[O-])[C-]([O-])[O-])[C-]([O-])[O-])[C-]([O-])[O-] has a net charge of -8.000004101544619
```

mattwthompson:
I can't tell ... the differences in environments shouldn't cause any changes in AM1-BCC charge assignment, they differ almost literally only in the NAGL version that's installed, which shouldn't affect AmberTool's behavior?

Failing: https://github.com/openforcefield/openff-interchange/actions/runs/8592685125/job/23543063253
Passing: https://github.com/openforcefield/openff-interchange/actions/runs/8606724443/job/23585728482

I have a gut feeling the difference will be difficult or impossible to find, and maybe there's just some quirk associated with using something several patch versions too old. For now I think I'll leave it be, but in the future it may be better to just load the guest charges from disk and say "at this point in time with XYZ installed we assigned AM1-BCC ..."

lilyminium:
Ah, I tracked it down because it was nagging away at me: partial charges are stored in `host.sdf` that the notebook was falling back on when it couldn't import NAGL.

```
>>> host = Molecule.from_file("host.sdf")
>>> host.partial_charges.sum()
-8.000004101544619 elementary_charge
>>> host.total_charge
-24.0 elementary_charge
```

Issue #961, opened 2024-04-09
Interchange.to_gromacs() creates a topology with far too many atomtypes, which influences GROMACS performance


pbuslaev:
**Description**
When dumping the interchange to GROMACS format, too many atom types are created. This is completely fine in terms of compatibility with GROMACS, but it is causing huge memory pressure on GROMACS. The issue is that GROMACS is creating and keeping in memory the matrix of all non-bonded interactions. So the more atom types you have, the larger the table is. The table size (and the memory requirements) has $n^2$ dependence on the number of atom types. For me, a system of around 55,000 atoms led to memory usage over 32 GB at both `grompp` and `mdrun` stages.

To demonstrate the issue, I used the [`102L`](https://www.rcsb.org/structure/102L) structure. I removed all the waters manually with vmd and added hydrogens with `gmx pdb2gmx`:
```shell
gmx pdb2gmx -f 102L_clean.pdb -o 102L_h.pdb
```

When I ran `grompp` with the topology and structure created by GROMACS with `amber99sb-ildn` force field using the following basic `.mdp` file:
```
; minim.mdp - used as input into grompp to generate em.tpr
; Parameters describing what to do, when to stop and what to save
integrator  = steep         ; Algorithm (steep = steepest descent minimization)
emtol       = 1000.0        ; Stop minimization when the maximum force < 1000.0 kJ/mol/nm
emstep      = 0.01          ; Minimization step size
nsteps      = 5000         ; Maximum number of (minimization) steps to perform

; Parameters describing how to find the neighbors of each atom and how to calculate the interactions
nstlist         = 1         ; Frequency to update the neighbor list and long range forces
cutoff-scheme   = Verlet    ; Buffered neighbor searching
ns_type         = grid      ; Method to determine neighbor list (simple, grid)
coulombtype     = PME       ; Treatment of long range electrostatic interactions
rcoulomb        = 1.0       ; Short-range electrostatic cut-off
rvdw            = 1.0       ; Short-range Van der Waals cut-off
pbc             = xyz       ; Periodic Boundary Conditions in all 3 dimensions
```
I got:

```
# Execution command
time gmx grompp -f min.mdp -c 102L_h.gro -p topol.top -o test1.tpr -maxwarn 1

# Output
                      :-) GROMACS - gmx grompp, 2023.1 (-:

Executable:   /software/gromacs_v2023.1/bin/gmx
Data prefix:  /software/gromacs_v2023.1
[...]

PR #962, opened 2024-04-09
draft: Add atom type merging for Interchange.to_gromacs()

pbuslaev:
### Description

This should be solving #961 

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/interop/gromacs/export/_export.py
+++ b/openff/interchange/interop/gromacs/export/_export.py
@@ -80,11 +80,11 @@ class GROMACSWriter(DefaultModel):
             """
             Checks if the atom type is already in list.
             """
-            from openff.toolkit import Quantity
+            from openff.units import Quantity
 
             mass_tolerance = Quantity("1e-5 amu")
             sigma_tolerance = Quantity("1e-5 nanometer")
-            epsilon_tolerance = Quantity("1e-5 kilojoule_per_mol")
+            epsilon_tolerance = Quantity("1e-5 kilojoule_per_mole")
             for _at_name, _atom_type in atom_type_list:
                 if (
                     atom_type.atomic_number == _atom_type.atomic_number
--- a/openff/interchange/interop/gromacs/export/_export.py
+++ b/openff/interchange/interop/gromacs/export/_export.py
@@ -32,7 +32,7 @@ class GROMACSWriter(DefaultModel):
 
         with open(self.top_file, "w") as top:
             self._write_defaults(top)
-            to_reduced_atom_types = self._write_atomtypes(top)
+            reduced_atom_types = self._write_atomtypes(top)
 
             self._write_moleculetypes(top, to_reduced_atom_types)
 
--- a/openff/interchange/interop/gromacs/export/_export.py
+++ b/openff/interchange/interop/gromacs/export/_export.py
@@ -82,7 +82,8 @@ class GROMACSWriter(DefaultModel):
                     atom_type.atomic_number == _atom_type.atomic_number
                     and abs(atom_type.mass - _atom_type.mass).m < mass_tolerance
                     and abs(atom_type.sigma - _atom_type.sigma).m < sigma_tolerance
-                    and abs(atom_type.epsilon - _atom_type.epsilon).m < epsilon_tolerance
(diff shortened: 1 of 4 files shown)


pbuslaev:
This is a draft of the solution, if anyone can suggest me with optimal tests for it, or other considerations to keep in mind, I will be happy to work on those

openff/interchange/interop/gromacs/export/_export.py
```diff
@@ -60,31 +60,71 @@ def _write_defaults(self, top):
...
+        ) -> bool | str:
+            """
+            Checks if the atom type is already in list.
+            """
+            for _at_name, _atom_type in atom_type_list:
+                if (
+                    atom_type.atomic_number == _atom_type.atomic_number
+                    and numpy.abs(atom_type.mass.m - _atom_type.mass.m) < diff
+                    and numpy.abs(atom_type.sigma.m - _atom_type.sigma.m) < diff
+                    and numpy.abs(atom_type.epsilon.m - _atom_type.epsilon.m) < diff
```
mattwthompson, line 83: Since these aren't vectorized, and since in the future they might be stored with different units, I think something like this might be safer and/or faster

```suggestion
                    and abs(atom_type.mass - _atom_type.mass).m < mass_tolerance
                    and abs(atom_type.sigma - _atom_type.sigma).m < sigma_tolerance
                    and abs(atom_type.epsilon - _atom_type.epsilon).m < epsilon_tolerance
```
with something like

```python
mass_tolerance = Quantity("1e-5 amu")
sigma_tolerance = Quantity("1e-5 nanometer")
epsilon_tolerance = Quantity("1e-5 kilojoule_per_mol")
```
defined above (to avoid re-initializing each during every loop)

mattwthompson (commented):
I think this is a good start, it's assuring that no existing tests seem to be passing. My primary worry is something subtly changing with the energies, perhaps in the relationship between nonbonded parameters and bonded parameter housekeeping, but maybe that's not founded.

To keep this moving, we'll definitely want some tests with the flag turned on and off (included in the test suite) for a variety of chemistries and some quick checks that this actually fixes the memory explosion (one-off analysis).

pbuslaev:
So now I added a flag `merge_atom_types` and I drafted some tests (roundtrip, and energy comparison).

openff/interchange/interop/gromacs/export/_export.py
```diff
@@ -60,31 +64,100 @@ def _write_defaults(self, top):
             f"{self.system.coul_14:8.6f}\n\n",
         )
 
-    def _write_atomtypes(self, top):
+    def _write_atomtypes(self, top, merge_atom_types: bool) -> dict[str, str]:
```
j-wags, line 67: It's great to see work toward the "atom type explosion" problem that's plagued us for years. But given my recollection of the severity of the parmed bug that arose from seemingly innocuous atom type merging, it may make sense to have this be `_merge_atom_types`(with the leading underscore) or require `INTERCHANGE_EXPERIMENTAL=1` for a few releases to communicate that it's new/experimental. Once early adopters have run with it for a while we can remove the underscore and make it a public method.

pbuslaev: That is fair, I added underscore to all public methods

mattwthompson:
Thanks @pbuslaev - in case I didn't earlier, I wanted to add a note to make it clear that this is probably in good shape but I need to spend some time tinkering with it in tests outside of the test suite. Some small errors in atom type accounting can lead to huge errors (for example, I've seen papers retracted because non-water parameters overwrote water parameters: https://github.com/ParmEd/ParmEd/issues/1197). Keeping this in the "private" API is definitely the way to do it, but I also want to run it through a bunch of tests cases to ensure nothing funny happens since it might make its way to the public API in the future.

pbuslaev:
Thanks @mattwthompson. Let me know if you have some specific tests in mind and I can run those locally as well.

mattwthompson:
Thanks for the contribution @pbuslaev! I spent a good portion of today shuffling these changes into a new branch and running more tests. #973 is merged and should be in the next release. So far everything is looking great - impressive work!

PR #979, opened 2024-04-27
Added multiplicity to key, when reading topology from_gromacs

pbuslaev:
### Description

This should resolve Solves #978

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

Issue #978, opened 2024-04-26
Dihedrals lost in `Interchange.from_gromacs`


mattwthompson:
```python
import os

from openff.toolkit import ForceField, Molecule

from openff.interchange import Interchange
from openff.interchange.drivers import get_gromacs_energies

os.environ["INTERCHANGE_EXPERIMENTAL"] = "1"


sage_unconstrained = ForceField("openff_unconstrained-2.0.0.offxml")

ethanol = Molecule.from_smiles("CCO")
ethanol.generate_conformers(n_conformers=1)

ethanol._conformers = [ethanol.conformers[0].round(2)]

ethanol_interchange = sage_unconstrained.create_interchange(ethanol.to_topology())
ethanol_interchange.box = [4, 4, 4]

ethanol_interchange.to_gromacs("ethanol")

converted_ethanol = Interchange.from_gromacs(
        "ethanol.top",
        "ethanol.gro",
)
converted_ethanol.to_gromacs("roundtrip")

get_gromacs_energies(ethanol_interchange).compare(get_gromacs_energies(converted_ethanol))

# openff.interchange.exceptions.EnergyError: {'Torsion': <Quantity(1.13361835, 'kilojoule / mole')>, 'Electrostatics': <Quantity(-0.00525069237, 'kilojoule / mole')>}
```

```shell
$ wc -l ethanol.top roundtrip.top                                                                                       14:07:25  ☁  gmx-unique-molecule-names ☂
     108 ethanol.top
     104 roundtrip.top
     212 total
(openff-interchange-dev) [openff-interchange]
$ sed -n '74,92p' ethanol.top && sed -n '74,88p' roundtrip.top                                                          14:07:33  ☁  gmx-unique-molecule-names ☂
[ dihedrals ]
;ai    aj   ak   al   funct phi  k
     4     1     2     8     1      0.000000000000      0.799950138473                 3
     3     2     1     5     1      0.000000000000      0.466735938990                 3
     3     2     1     5     1      0.000000000000      1.436153909162                 1
     6     1     2     7     1      0.000000000000      0.799950138473                 3
     5     1     2     8     1      0.000000000000      0.799950138473                 3
     6     1     2     8     1      0.000000000000      0.799950138473                 3
     3     2     1     4     1      0.000000000000      0.466735938990                 3
     3     2     1     4     1      0.000000000000      1.436153909162                 1
     3     2     1     6     1      0.000000000000      0.466735938990                 3
[...]

PR #980, opened 2024-04-29
Fix parsing GROMACS dihedrals

mattwthompson:
### Description

#979 / #978

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/_tests/unit_tests/interop/gromacs/export/test_export.py
+++ b/openff/interchange/_tests/unit_tests/interop/gromacs/export/test_export.py
@@ -23,11 +23,7 @@ from openff.interchange._tests import (
 from openff.interchange.components.nonbonded import BuckinghamvdWCollection
 from openff.interchange.components.potentials import Potential
 from openff.interchange.drivers import get_gromacs_energies, get_openmm_energies
-from openff.interchange.exceptions import (
-    EnergyError,
-    GMXMdrunError,
-    UnsupportedExportError,
-)
+from openff.interchange.exceptions import GMXMdrunError, UnsupportedExportError
 from openff.interchange.interop.gromacs._import._import import (
     _read_box,
     _read_coordinates,
@@ -486,15 +482,12 @@ class TestGROMACS:
                 "Electrostatics": 0.05 * unit.kilojoule / unit.mol,
             },
         )
-        try:
-            get_gromacs_energies(combined).compare(
-                get_gromacs_energies(combined_from_converted),
-                tolerances={
-                    "Electrostatics": 0.05 * unit.kilojoule / unit.mol,
-                },
-            )
-        except EnergyError:
-            pytest.xfail("Not sure why these energies differ; parameters look okay")
+        get_gromacs_energies(combined).compare(
+            get_gromacs_energies(combined_from_converted),
+            tolerances={
+                "Electrostatics": 0.05 * unit.kilojoule / unit.mol,
+            },
+        )
 
 
 class TestGROMACSMetadata:
--- a/openff/interchange/_tests/unit_tests/interop/gromacs/export/test_export.py
+++ b/openff/interchange/_tests/unit_tests/interop/gromacs/export/test_export.py
@@ -414,6 +414,82 @@ class TestGROMACS:
 
         assert [*parmed.load_file("tmp.top").molecules.keys()] == ["MOL0", "MOL1"]
 
+    @pytest.mark.parametrize("name", ["MOL0", "MOL222", ""])
+    def test_roundtrip_with_combine(
+        self,
+        name,
+        sage_unconstrained,
+        ethanol,
+        cyclohexane,
+        monkeypatch,
+    ):
+        monkeypatch.setenv("INTERCHANGE_EXPERIMENTAL", "1")
+
+        for molecule in [ethanol, cyclohexane]:
+            molecule.name = name
+            molecule.generate_conformers(n_conformers=1)
+
+        # translate one by 1 nm
+        cyclohexane._conformers = [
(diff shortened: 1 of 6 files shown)


openff/interchange/_tests/unit_tests/interop/gromacs/export/test_export.py
```diff
@@ -414,6 +418,84 @@ def test_exisiting_mol0_names_overwritten(self, name, sage, ethanol, cyclohexane
...
+                "Electrostatics": 0.05 * unit.kilojoule / unit.mol,
+            },
+        )
+        try:
+            get_gromacs_energies(combined).compare(
+                get_gromacs_energies(combined_from_converted),
+                tolerances={
+                    "Electrostatics": 0.05 * unit.kilojoule / unit.mol,
+                },
+            )
```
mattwthompson, line 495: This is failing pretty catastrophically, not yet sure why

pbuslaev: Also this is failing

```python
get_openmm_energies(converted_ethanol.combine(converted_cyclohexane)).compare(get_openmm_energies(converted_cyclohexane.combine(converted_ethanol)), tolerances={"Nonbonded": Quantity(0.01, "kilojoule/mole")})
```

pbuslaev:
Ah, ok. I see what is going on. Created `PotentialKey` for different molecules from gromacs are also identical. Thus, only one value is loaded from the [dictionary](https://github.com/openforcefield/openff-interchange/blob/da76d19bbbed3c3d9e8717f5d47a0cd39269076b/openff/interchange/operations/_combine.py#L99). I will work on the potential solution

mattwthompson:
Wow, what a mess. Maybe at import, these should be keyed by something like `f"{molecule_name}-{atomtype1}-{atomtype2}"`?

pbuslaev:
This can be quite a big change, since it has to be introduced into `Bonds`, `Angles`, etc.. I am now testing a different solution, where I am checking if `PotentialKey` is built from indices.

mattwthompson:
Big changes to experimental features are allowed - I consider the state after `Interchange.from_gromacs` to be API- and behavior-unstable and sweeping changes for purposes of fixing bugs or improving performance are okay. By contrast, I'd be apprehensive about changing the results of a vanilla `ForceField.create_interchange` state

Building `PotentialKey` from indices conflicts with the benefit of de-duplicating parameters; this doesn't make much of a difference for these small systems but should for larger systems, proteins, etc.

pbuslaev:
At the moment all my changes are within `_combine` method of `interchange.operations._combine`. What I am trying to do is to check, if the potential key is from indices:

```python
def _check_pot_key_is_by_index(pot_key) -> bool:
        """
        Checks if pot_key is created by index, not by SMARTS.

        Returns
        -------
        True, if generated by index.
        """
        for _atom in pot_key.id.split("-"):
            try:
                int(_atom)
            except ValueError:
                return False
        return True
```

pbuslaev:
But it also won't work with `mol_name`. Currently this scheme is used for `vdW` Handler, and `vdW` are also failing. In the test example, the `mol_name` in the interchanges is the same, so again we fall into same names for different molecules.

pbuslaev:
You are right, but we are dealing with `combine` method. So if I create interchanges like this:
```python
m1 = Molecule("CCC")
m2 = Molecule("CCO")
# skipping conformer generation, etc
i1 = ff.create_interchange(m1.to_topology())
i2 = ff.create_interchange(n2.to_topology())
i1.to_gromacs("m1")
i2.to_gromacs("m2")
i1_c = Interchange.from_gromacs("m1.top", "m1.gro")
i2_c = Interchange.from_gromacs("m2.top", "m2.gro")
```
both `i1_c` and `i2_c` will have handlers associated with `MOL0`. So if I now will try to combine `i1_c.combine(i2_c)` we will have overlap in `PotentialKey` names.

mattwthompson:
I see, that's kinda nasty. This logic might need to get propagated out to `Interchange.combine`: https://github.com/openforcefield/openff-interchange/blob/da76d19bbbed3c3d9e8717f5d47a0cd39269076b/openff/interchange/smirnoff/_gromacs.py#L120-L129

But doing _all_ of the housekeeping for all of the interactions sounds like a pain. I wonder if there's a way to dynamically update all of those keys, or just have them defined as pointers to the molecule name ...

mattwthompson:
There are ways to check if a string is valid SMARTS/SMIRKS, but it'd probably be better to just make a different subclass(es) for these different behaviors ... kinda messy

pbuslaev:
Maybe a simple solution is to add a flag to `PotentialKey` if potentials come from SMIRNOFF or from anywhere else. Then, depending on this flag, different behaviour can be triggered. Maybe `cosmetic_attributes` can be used for it?

mattwthompson:
Yep, we basically do the same thing in a similar situation (storing charges from OpenMM): https://github.com/openforcefield/openff-interchange/blob/da76d19bbbed3c3d9e8717f5d47a0cd39269076b/openff/interchange/interop/openmm/_import/_import.py#L211

openff/interchange/_tests/unit_tests/interop/gromacs/export/test_export.py
mattwthompson: It's probably H-X bonds that aren't being parsed

mattwthompson:
It's a bit tricky since constraints typically need to know about bonds. Currently this is just dealt with by adding extra logic whenever exporting constraint information

pbuslaev (approved):

PR #982, opened 2024-04-29
Set vdW parameters on virtual sites

mattwthompson:
### Description

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/smirnoff/_create.py
+++ b/openff/interchange/smirnoff/_create.py
@@ -380,7 +380,9 @@ def _plugins(
                 collection = collection_class.create(
                     parameter_handler=force_field[handler_class._TAGNAME],
                     topology=topology,
-                    vdw_collection=interchange["DoubleExponential"],
+                    vdw_collection=interchange[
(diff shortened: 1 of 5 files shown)


openff/interchange/smirnoff/_create.py
```diff
@@ -371,10 +371,20 @@ def _plugins(
...
+                    parameter_handler=force_field[handler_class._TAGNAME],
+                    topology=topology,
+                )
+            except TypeError:
+                collection = collection_class.create(
+                    parameter_handler=force_field[handler_class._TAGNAME],
+                    topology=topology,
+                    vdw_collection=interchange[
+                        "DoubleExponential"
+                    ],  # need to hook into .acts_as attribute
```
mattwthompson, line 385: This is going to be a pain, I'm afraid there will need to be logic along the lines of "if any collection is meant to act as a vdW collection, that takes precedence over a vanilla vdW collection"

PR #990, opened 2024-06-12
Collection ergonomics

### LINKED ISSUE #968: Easier indexing into collections
**Description**
```python
In [17]: out = ForceField("openff-2.1.0.offxml").create_interchange(
    ...:     Molecule.from_smiles("CCO").to_topology()
    ...: )
```
Indexing into a collection to get (physics) parameters with known atom indices requires creating a new instance of a key:

```python
In [18]: out["Bonds"].potentials[out["Bonds"].key_map[BondKey(atom_indices=(0, 1))]]
Out[18]: Potential(parameters={'k': <Quantity(419.986927, 'kilocalorie / angstrom ** 2 / mole')>, 'length': <Quantity(1.52794022, 'angstrom')>}, map_key=None)
```

It'd be a little more convenient to, at least in simpler cases, allow lookup just by atom indices. That's really all of the information that's stored in some of those key objects:

```python
In [19]: out["Bonds"].potentials[(0, 1)]
```
I'm not sure if this should return a `PotentialKey` or `Potential`

Yoshanuikabundi:
### Description

Implements easier indexing into collections and dictionaries of potentials. Closes #968.

Works by defining equality between certain `TopologyKey` subclasses and tuples of atom indices when non-atom index attributes are `None`. Also implements `__getitem__` on `Collection` to index directly from atom indices to potential.

```python
>>> from openff.toolkit import ForceField, Molecule
>>> interchange = ForceField("openff-2.2.0.offxml").create_interchange(
...     Molecule.from_mapped_smiles("[H:1][O:2][H:3]").to_topology()
... )
>>> interchange['Bonds'].key_map[1, 2]
PotentialKey associated with handler 'Bonds' with id '[#8:1]-[#1:2]'
>>> interchange['Bonds'][1, 2]
Potential(parameters={'k': <Quantity(1076.70159, 'kilocalorie_per_mole / angstrom ** 2')>, 'length': <Quantity(0.975230101, 'angstrom')>}, map_key=None)
>>> 

```

I had to change the hash implementations to not include `None` in the hash if all non-atom indices attributes were `None` so that the hash matches between the `TopologyKey` and the appropriate tuple. This might occasionally cause collisions... but it should be extremely rare and regardless Python's requirement is that `__eq__` implies hashes are equal, not the converse.

Atom indices for bonds have to be in the correct order. Fixing this for indexing with tuples would require replacing `Collection.key_map` with a custom class. We could make the ergonomics easier by guaranteeing that atom indices in bonds are always ordered a certain way?

### Checklist

- [x] Add tests
- [x] Lint
- [ ] Update docstrings


mattwthompson:
> We could make the ergonomics easier by guaranteeing that atom indices in bonds are always ordered a certain way?

Seems worth it to me - but I'm not sure about other valence terms

mattwthompson:
Okay, I see you've gone about sorting the other terms. I think this breaks angles and dihedrals, though, since they're not symmetric like bonds?

Yoshanuikabundi:
>> We could make the ergonomics easier by guaranteeing that atom indices in bonds are always ordered a certain way?

> Seems worth it to me - but I'm not sure about other valence terms

We'd have to do it everywhere because this PR just indexes in directly. It might be better to just look up all equivalent index orderings?

>Okay, I see you've gone about sorting the other terms.

Welp I didn't mean to, I'll add tests and double check

> I think this breaks angles and dihedrals, though, since they're not symmetric like bonds?

Yeah I agree, you can reverse the order but I think that's the only symmetry op for both?

Yoshanuikabundi:
OK I've just made it so that when indexing into a collection with a tuple, both the forward and reverse of the tuple will be looked up. I think this is a lot less intrusive than guaranteeing everywhere that the tuples are ordered. My understanding is that all valence terms can be indexed in either direction, but if this isn't the case then I can change it to only look up the reverse for 2-tuples.

mattwthompson:
> could you point out where I've done this?

Nope, I think I got confused looking too superficially at which tests you've added

mattwthompson:
Am I meant to be able to look up other valence terms, or is this messed up by all of those extra fields (some of which might not be necessary)?

```python
In [12]: interchange = ForceField("openff-2.2.0.offxml").create_interchange(Molecule.from_smiles("
    ...: CCO").to_topology())

In [13]: next(iter(interchange['ProperTorsions'].key_map))
Out[13]: ProperTorsionKey with atom indices (0, 1, 2, 8), mult 0

In [14]: interchange['ProperTorsions'].key_map[0, 1, 2, 8]
KeyError                                  Traceback (most recent call last)
Cell In[14], line 1
----> 1 interchange['ProperTorsions'].key_map[0, 1, 2, 8]

KeyError: (0, 1, 2, 8)

**Comment by Yoshanuikabundi** [comment]
Yeah that's on purpose - since I implemented this by defining `__eq__` on the `TopologyKey` subclasses, you have to include all the information in the tuple - otherwise you violate `a == b implies hash(a) == hash(b)` and run into issues if you had two entries with the same atom indices. I do this implicitly by only making tuples equal to a `TopologyKey` when the non-indices fields of the `TopologyKey` are the defaults.

I thought about allowing more information to be given in the tuple - for example, to index the above torsion you could do something like:

```python
interchange['ProperTorsions'].key_map[(0, 1, 2, 8), 0]
# Or
interchange['ProperTorsions'].key_map[(0, 1, 2, 8), 0, None, None]
```

But since users would have to remember the correct ordering for everything and there's no way to disambiguate the name of the field you're specifying and all the fields have the same type I thought this provided a relatively small ergonomics boost. If users want to specify the extended fields, they currently need to index with a `ProperTorsionsKey` (or whatever), which provides that documentation (via autocomplete) and disambiguation (via named arguments). I chose the existing behavior as the best ratio of ergonomics improvement to amount of new code, but I made this judgement assuming the default values of all fields would cover the majority of use cases.

Here are some ways I've thought about dealing with this:

1. Leave things as they are

2. Implement equality between extended tuples as above (or some subset)

3. If `mult=0` and `mult=None` are equivalent, change the default value of `mult` (and the value that `__eq__` and `__hash__` assume) to `0` (for bonus points, also change the type from `int | None` to `int`)

4. If `mult=0` is more common than `mult=None` but a breaking change is not desired, change just the value that `__eq__` and `__hash__` assume to `0` (I think this would be more confusing than 3)

5. Implement a more detailed lookup in `Collections.__getitem__` and if the given key is unambiguous wrt the existing keys, use it - the downside of this is that since we don't know the correct value ahead of time (or even the correct type) we'd actually have to iterate over all the existing keys to find possible matches, which changes worst case lookup from O(1) to O(n).

Note also that indexing into `collection` is different to indexing into `collection.key_map`, as I didn't want to change `key_map` away from being a simple dict:

```python
In [10]: next(iter(interchange['Bonds'].key_map))
Out[10]: BondKey with atom indices (0, 1)

In [11]: interchange['Bonds'][0, 1] # Returns the underlying potential
Out[11]: Potential(parameters={'k': <Quantity(430.602781, 'kilocalorie_per_mole / angstrom ** 2')>, 'length': <Quantity(1.5336276, 'angstrom')>}, map_key=None)

In [12]: interchange['Bonds'][1, 0] # Works with reversed keys
Out[12]: Potential(parameters={'k': <Quantity(430.602781, 'kilocalorie_per_mole / angstrom ** 2')>, 'length': <Quantity(1.5336276, 'angstrom')>}, map_key=None)

In [13]: interchange['Bonds'].key_map[0, 1] # Returns the PotentialKey
Out[13]: PotentialKey associated with handler 'Bonds' with id '[#6X4:1]-[#6X4:2]'

In [14]: interchange['Bonds'].key_map[1, 0] # Doesn't work with reversed keys
KeyError                                  Traceback (most recent call last)
Cell In[14], line 1
----> 1 interchange['Bonds'].key_map[1, 0]

KeyError: (1, 0)
```

openff/interchange/_tests/unit_tests/test_models.py
```diff
@@ -110,3 +111,61 @@ def test_reprs():
...
+    assert BondKey(atom_indices=(1, 3)) == (1, 3)
+    assert hash(BondKey(atom_indices=(1, 3))) == hash((1, 3))
+    assert BondKey(atom_indices=(1, 3)) != (1, 4)
+    assert BondKey(atom_indices=(1, 3)) != (3, 1)
+    assert BondKey(atom_indices=(1, 3), bond_order=None) == (1, 3)
+    assert hash(BondKey(atom_indices=(1, 3), bond_order=None)) == hash((1, 3))
+    assert BondKey(atom_indices=(1, 3), bond_order=None) != ((1, 3), None)
+    assert BondKey(atom_indices=(1, 3), bond_order=1.5) != (1, 3)
+    assert BondKey(atom_indices=(1, 3), bond_order=1.5) != ((1, 3), None)
+    assert BondKey(atom_indices=(1, 3), bond_order=1.5) != ((1, 3), 1.5)
```
mattwthompson, line 134: I'm pretty sure these two should be equal?

```python
    def __hash__(self) -> int:
        if self.bond_order is None:
            return hash(tuple(self.atom_indices))
        return hash((tuple(self.atom_indices), self.bond_order))

**Review by mattwthompson — APPROVED** [review]
I think I'm past providing useful feedback on this changelog, which is long past due its green check. I want to point this to the `develop` branch which targets an upcoming 0.4.x line; one test confuses me but otherwise please merge this when you're happy with it (% any confusion I've added).

**Comment by mattwthompson** [comment]
I get why equality across extended tuples is tricky; given that the motivation of this behavior is making our lives easier, I'm fine leaving that door closed.

`mult` in general is ... not something I'm completely happy with but also something I'm not eager to change. `BondKey` basically representing a tuple with some extra (optional!) flavors feels right, but how periodic torsions are handled feels clunkier. Not to be dealt with here, but I think that's a prime candidate for a future refactor.

I looked through the docs to find a good place to show this off, but I didn't find anything. In the future that might be worth a subsection, either in the static docs or a notebook(s) somewhere. I expected there to be tons of these excessiely-difficult-lookup cases, but now I think those are just in tests and the source code. I don't see much of a reason to rewrite what's there, but with any luck I'll remember this trick and future code will be easier to write and a little bit more sightly to look at.

File: openff/interchange/_tests/unit_tests/test_models.py (hunk 1 above)
Comment (Yoshanuikabundi) [reply openff/interchange/_tests/unit_tests/test_models.py]: I believe this test is correct - Lines 131, 133 and 134 are all extended tuples, which has not been implemented (as we've discussed elsewhere), and in line 133 the bond order is not `None` so the tuple doesn't match.

The second return line in that hash function is the existing behaviour - `BondKey.__eq__` is 

```py
def __eq__(self, other) -> bool:
return super().__eq__(other) or (
    self.bond_order is None and other == self.atom_indices
)
```

Yoshanuikabundi: Comparing hashes in `__eq__` can also run into issues when comparing objects of different types.

Yoshanuikabundi:
Ok well I implemented extended tuples as a way to keep the `TopologyKey.__eq__` implementation simple and correct while ensuring it's consistent with `TopologyKey.__hash__`. I've added an example of indexing to the `Interchange` API docs and also to the [part](https://openff-docs--990.org.readthedocs.build/projects/interchange/en/990/using/collections.html) of the user guide that describes the inner workings of `Collection`.

There are still a few topology key classes that are not subclasses of `TopologyKey` because they correspond to a single atom - I think the best way forward there would just be to make the single atom index field a property that points to `self.atom_indices[0]`, but I haven't implemented this.

If you're happy with this change set, I'm happy for you to merge it at your convenience!

PR #1000, opened 2024-07-09
LAMMPS non-unique molecule ID hotfix

timbernat:
### Description

Follow-up to recommendations made in [Issue 999](https://github.com/openforcefield/openff-interchange/pull/999)

### Checklist

- [x] Add molecule_index to CTAB hash to ensure uniqueness
- [x] Add tests
- [x] Ensure tests are compliant with pytest
- [x] Lint
- [x] Update docstrings

### DIFF
--- a/openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
+++ b/openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
@@ -82,7 +82,9 @@ class TestLammps:
 
         # BUILD TOPOLOGY
         ## 1) compute effective radius as the greatest atomic distance from barycenter (avoids collisions when tiling)
-        pilot_mol = MoleculeWithConformer.from_smiles(smiles) # this will serve as a prototype for all other Molecule copies in the Topology
+        pilot_mol = MoleculeWithConformer.from_smiles(
+            smiles
+        )  # this will serve as a prototype for all other Molecule copies in the Topology
 
         conf = pilot_mol.conformers[0]
         COM = conf.mean(axis=0)
--- a/openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
+++ b/openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
@@ -81,7 +81,7 @@ class TestLammps:
         [
             "CCO",
             "N1CCCC1",
-            "c1cccc1c",
+            "c1ccccc1",
         ],
     )
     def test_unique_lammps_mol_ids(
--- a/openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
+++ b/openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
@@ -75,10 +75,11 @@ class TestLammps:
         self,
         smiles: str,
         sage_unconstrained: ForceField,
-        sidelen: int = 0,
+        sidelen: int = 2,
     ) -> bool:
         """Test to see if interop.lammps.export._write_atoms() writes unique ids for each distinct Molecule"""
         temp_lammps_path = Path.cwd() / "temp.lmp"
+        assert(isinstance(sidelen, int) and sidelen > 1)
 
         # BUILD TOPOLOGY
         ## 1) compute effective radius as the greatest atomic distance from barycenter (avoids collisions when tiling)
--- a/openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
+++ b/openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
@@ -72,6 +72,17 @@ class TestLammps:
             },
         )
 
+    @pytest.mark.parametrize("sidelen", [2, 3])
+    @pytest.mark.parametrize(
+        "smiles",
+        [ 
+            "O",   # nothing particularly special about these molecules
+            "CCO", # just testing that unique IDs hold for diverse chemistries
+            "N1CCCC1",
+            "c1cccc1c",
+
+        ]
+    )
     def test_unique_lammps_mol_ids(
         self,
         smiles: str,
--- a/openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
+++ b/openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
@@ -1,6 +1,11 @@
+from typing import Any, Type, Dict
+
 import numpy
 import pytest
-from openff.toolkit import Topology, unit
+
+from pathlib import Path
+from openff.toolkit import Molecule, Topology, unit
+from openff.toolkit.typing.engines.smirnoff import ForceField
 
 from openff.interchange import Interchange
 from openff.interchange._tests import MoleculeWithConformer, needs_lmp
@@ -21,7 +26,7 @@ class TestLammps:
             "C1COC(=O)O1",  # This adds an improper, i2
         ],
     )
-    def test_to_lammps_single_mols(self, mol, sage_unconstrained, n_mols):
+    def test_to_lammps_single_mols(self, mol : str, sage_unconstrained : ForceField, n_mols : int) -> None:
         """
         Test that Interchange.to_openmm Interchange.to_lammps report sufficiently similar energies.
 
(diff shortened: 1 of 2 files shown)


openff/interchange/interop/lammps/export/export.py
```diff
@@ -294,7 +294,10 @@ def _write_atoms(lmp_file: IO, interchange: Interchange, atom_type_map: dict):
         atom_map[atom_index] = atom
 
     for molecule_index, molecule in enumerate(interchange.topology.molecules):
-        molecule_hash = molecule.ordered_connection_table_hash()
+        # molecule_hash = molecule.ordered_connection_table_hash()
+        molecule_hash = f"{molecule_index}_{molecule.ordered_connection_table_hash()}"  # salt with mol ID to allow chemically-identical Molecules to be labelled with distinct IDs)
```
mattwthompson, line 298: IIUC, salting is the process of adding ~random data into hashes, which isn't what you're after here
```suggestion
        # inject mol ID into hash to allow chemically-identical Molecules
        # to be labelled with distinct IDs
        molecule_hash = f"{molecule_index}_{molecule.ordered_connection_table_hash()}"
```

openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
```diff
@@ -62,3 +68,104 @@ def test_to_lammps_single_mols(self, mol, sage_unconstrained, n_mols):
...
+    ) -> bool:
+        """Test to see if interop.lammps.export._write_atoms() writes unique ids for each distinct Molecule"""
+        temp_lammps_path = (
+            Path.cwd() / "temp.lmp"
+        )  # NOTE: lammps writer doesn't currently support IO-like object passing, so need to intercept output with temporary file
+
+        # BUILD TOPOLOGY
+        ## 0) generate pilot Molecule and conformer
+        pilot_mol = Molecule.from_smiles(smiles)
+        pilot_mol.generate_conformers(n_conformers=1)
```
mattwthompson, line 83: I have a `MoleculeWithConformer` class in `openff.interchange._tests` that does this in one pass

openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
```diff
@@ -62,3 +68,104 @@ def test_to_lammps_single_mols(self, mol, sage_unconstrained, n_mols):
...
             },
         )
+
+    def test_unique_lammps_mol_ids(
+        self, smiles: str, sage_unconstrained: ForceField, sidelen: int = 0
+    ) -> bool:
+        """Test to see if interop.lammps.export._write_atoms() writes unique ids for each distinct Molecule"""
+        temp_lammps_path = (
+            Path.cwd() / "temp.lmp"
+        )  # NOTE: lammps writer doesn't currently support IO-like object passing, so need to intercept output with temporary file
```
mattwthompson, line 78:
```suggestion
        )
```
Tracked in https://github.com/openforcefield/openff-interchange/issues/1001

openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
```diff
@@ -1,6 +1,10 @@
+from pathlib import Path
+from typing import Any, Dict, Type
```
mattwthompson, line 2: In general I don't think annotating tests is worthwhile - although annotating the helper functions like `extract_info_from_lammps_atom_lines` is fine

openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
```diff
@@ -1,6 +1,10 @@
...
+
 import numpy
 import pytest
-from openff.toolkit import Topology, unit
+from openff.toolkit import Molecule, Topology, unit
+from openff.toolkit.typing.engines.smirnoff import ForceField
```
mattwthompson, line 7:
```suggestion
from openff.toolkit import Molecule, Topology, unit, ForceField
```

openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
```diff
@@ -62,3 +68,104 @@ def test_to_lammps_single_mols(self, mol, sage_unconstrained, n_mols):
...
+        box_vectors = (
+            numpy.eye(3) * box_dims
+        )  # convert to diagonal matrix to get proper shape
+
+        # EXPORT TO LAMMPS
+        interchange = Interchange.from_smirnoff(
+            sage_unconstrained, tiled_top
+        )  # , charge_from_molecules=[pilot_mol])
+        # UNRELATED QUESTION: why are waters not correctly recognized?
+        # PDB residue name on output is always UNK for Molecule.from_smiles("O"), EVEN when using tip3p.offxml)
```
mattwthompson, line 126: SMIRNOFF force fields don't define residue metadata and our infrastructure doesn't have default residue names for small molecules. As cumbersome as it might seem for simple cases like water, residue metadata needs to be set manually (or programmatically ...) https://github.com/openforcefield/openff-toolkit/issues/1554

openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
```diff
@@ -62,3 +68,104 @@ def test_to_lammps_single_mols(self, mol, sage_unconstrained, n_mols):
...
+        interchange.box = box_vectors
+        interchange.to_lammps(temp_lammps_path)
+
+        # EXTRACT ATOM INFO FROM WRITTEN LAMMPS FILE TO TEST IF MOLEUCLE IDS ARE BEING WRITTEN CORRECTLY
```
mattwthompson, line 130: Could any of the next ~35 lines be streamlines with LAMMPS's in-Python parser? https://docs.lammps.org/Python_module.html

openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
```diff
@@ -62,3 +68,104 @@ def test_to_lammps_single_mols(self, mol, sage_unconstrained, n_mols):
                 "Torsion": 0.02 * unit.kilojoule_per_mole,
             },
         )
+
+    def test_unique_lammps_mol_ids(
+        self, smiles: str, sage_unconstrained: ForceField, sidelen: int = 0
```
mattwthompson, line 73:
```suggestion
        self, sage_unconstrained: ForceField,
```
If `sidelen` is only one value, might as well leave it out of the (test) function definition. It'd make sense to include as an argument into some positions-building helper function if the middle of this test is refactored out.

I think you might also might be getting tripped up about how arguments are passed to `pytest`-style tests in general; most tests either don't need any inputs (save `self` with this class-based organization I like) or fixtures (which you can think of as just constants). Arguments that might take different values only make sense if they're being managed by the fancy magic of a `pytest.mark.parametrize` decorator, so `smiles` here isn't even defined, I don't think. If you're just looking to test one molecule, I'd drop this argument and just directly call `Molecule.from_smiles("CCO")` or whatever down below.

mattwthompson (commented):
Looking good, I have a few comments that might streamline some of the stuff you're doing here. Interoperability work is unfortunately a catastrophic mess of complexity, so "unit" tests can't be the beautiful <10 LOC tests they teach in textbooks.

openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
```diff
@@ -62,3 +68,104 @@ def test_to_lammps_single_mols(self, mol, sage_unconstrained, n_mols):
...
+
+        written_mol_ids: set[int] = set()
+        for atom_line in atom_lines:
+            atom_info = extract_info_from_lammps_atom_lines(atom_line)
+            written_mol_ids.add(atom_info["molecule_index"])
+        expected_mol_ids = {
+            i + 1 for i in range(interchange.topology.n_molecules)
+        }  # we'd like for each of the N molecules to have ids from [1...N]
+
+        return expected_mol_ids == written_mol_ids
```
mattwthompson, line 171:
```suggestion
        assert expected_mol_ids == written_mol_ids
```

I think my useless ` -> None:` annotation in a different test suggested that this should return a boolean of the pass/fail state, but `pytest`-style tests are meant to not return anything and just run without errors (read: fail on various `assert` statements) synonymously with passing

openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
timbernat: ForceField import from `...smirnoff` was because the direct import from openff.toolkit raised an import error on my end when attempting to import TestLammps. I'll consolidate these and see if the issue persists, or if the test remains runable

openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
timbernat: I had superficially based the structure of the inputs to my test on the [parameters for `test_to_lammps_single_mols()` above](https://github.com/openforcefield/openff-interchange/blob/lammps-hotfix/openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py#L16-L26) without having a great sense of how these parameters interact, thanks for clarifying this to me.

`sidelen` can be any arbitrary positive int (except 1, since you can't test for uniqueness when there's only 1 molecule present) and I left it as a free parameter accordingly; incidentally, I now notice the default value in this commit of `sidelen=0` is wrong for this reason. Would it be better to hard-code an arbitrary value within the function block (e.g. 3) or pass in values via `pytest.mark.parametrize`? Same goes for `smiles`, I wanted to be able to test any molecule to make there's nothing chemistry-specific to the test (namely for water or other "special" molecules). Should I hard-code a smiles string or just parameterize as with `test_to_lammps_single_mols()`?

openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
timbernat: I've removed the Dict and Type imports in [this commit](https://github.com/openforcefield/openff-interchange/commit/424de3ba75509b5b290cad53d741deacfe6d0ab6).The pathlib import is necessary to fetch the current working directly (and because I prefer Path objects over strings :P), and the typing.Any import is to annotate `extract_info_from_lammps_atom_lines()`.

openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
timbernat: Adopted and resolved in [this commit](https://github.com/openforcefield/openff-interchange/commit/4c50cd337dbe432b916be50644354ae7abcf41b6)

openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
timbernat: They totally can! See [this commit](https://github.com/openforcefield/openff-interchange/commit/8b72acde8e0c70cded0a81ad5d54b4e68be3660e). One needs to specify atom/bond/pair etc styles to get the lammps interface to read the data file. I did this using MDConfig, as I thought it the most internally-consistent solution; the extra file this requires is easy to clean up in post and hopefully an acceptable tradeoff for the more concise test.

openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
timbernat: I've parameterized these fields using `pytest.mark.parameterize()` in [this commit](https://github.com/openforcefield/openff-interchange/commit/652ad4359a48bd03aed9f9a06bbdfb05cfc5a25a), but I'd be more than happy to nix these changes if they're errant.

timbernat:
Alright @mattwthompson I've incorporated the streamlining suggestions you made in one form or another and verified that the test still behaves as expected. Could you have a final look through to see if we're ready to merge?

openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
mattwthompson: Great

openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
mattwthompson: Good - highlighting the `pathlib` import was incidental on my part

openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
mattwthompson: The quick imports should be available in any recent toolkit versions, starting somewhere around 0.11 or 0.12

mattwthompson: If you can't `from openff.toolkit import ForceField` that's an issue to me because it's a huge surprise, not because of the style

openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
mattwthompson: `parameterize` is a good solution here

mattwthompson:
@timbernat Thanks for your patience on this; since the actual change is fairly small I'm just going to pull the trigger

If there are other changes that would make your work run smoother, please feel free to coordinate with me on them or open PRs directly. Stuff in `lammps-improvements` was probably well-intentioned but lost attention in the mess of things

mattwthompson:
Let me try to revive it and see if it needs a rebase or something funkier

mattwthompson:
It looks like that branch wasn't so out of date it couldn't have `main` merged back in. Have a look and let's go from there: https://github.com/openforcefield/openff-interchange/pull/1011

Issue #1001, opened 2024-07-09
Support writing to file-like objects


mattwthompson:
@timbernat noted in #1000 that `to_lammps` doesn't support writing to file-like objects. It's probably not the only method that could use that functionality.

Issue #955, opened 2024-04-03
Improved LAMMPS support.


mrshirts:
**Description**
Please describe the behavior you would like added to Interchange.

Various things we've discovered about lammps files that are requiring/using some post-generation modification we would like to improve on.

* One HOGHLY annoying thing with the hybrid functional form:  Currently in the lammps data file output, hybrid functions (i.e. "bond_style                      hybrid harmonic" are specified in the data file as:

```
Bond Coeffs

1 harmonic      582.698766451   1.225198386222
2 harmonic      329.96998057905 1.427343958716
3 harmonic      301.15400723785 1.345741232799
4 harmonic      264.62148576755 1.52190126495
5 harmonic      370.04670688625 1.093899492634
6 harmonic      366.7408841747  1.425895053732

```
This is a totally valid simulation input. However, with the harmonic specified explicitly in the Bond Coeffs section,  APPARENTLY, when running from a restart file (and then converting that restart file to a new data file), this section is just plain skipped.  Not clear why - maybe the restart file can't store strings, so just ignores the lines instead (seems bad).  Thus, if one wants to continue a run by using the restart file, hybrid style terms have to be specified in the input file instead of the data file, i.e one would put in lines like: 

```
bond_coeff                      1 harmonic      582.698766451   1.225198386222
bond_coeff                      2 harmonic     329.96998057905 1.427343958716
bond_coeff                      3 harmonic      301.15400723785 1.345741232799
bond_coeff                      4 harmonic      264.62148576755 1.52190126495
bond_coeff                      5 harmonic      370.04670688625 1.093899492634
bond_coeff                      6 harmonic      366.7408841747  1.425895053732
```

* In the data file, "Atoms" should be "Atoms # full" to match the "full" in the atom style.  The comment triggers lammps to check if the data is indeed of atom full style.
 
* The "atom/bond/etc count" section format has changed.  LAMMPS reads both, but some tools/libraries do not use the old format.  Something like: 

```
9375 atoms
9250 bonds
16625 angles
28500 dihedrals
2250 impropers
 
8 atom types
9 bond types
5 angle types
25 dihedral types
1 improper types

``` 
Should now be written like: 

```
9375 atoms
8 atom types
9250 bonds
9 bond types
16625 angles
5 angle types
28500 dihedrals
25 dihedral types
2250 impropers
1 improper types
```

[...]

PR #903, opened 2024-02-14
Use LAMMPS's Python API in energy driver

mattwthompson:
### Description

Thanks to @timbernat for the suggestion

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

PR #897, opened 2024-02-09
Lammps constraints

mattwthompson:
### Description

Provide a brief description of the PR's purpose here.

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

PR #1011, opened 2024-07-16
LAMMPS improvements trunk

mattwthompson:
### Description

For now, just seeing how out of date this branch is. This may serve as a trunk for developing smaller LAMMPS-related features.


mattwthompson:
Here are the open LAMMPS issues: https://github.com/openforcefield/openff-interchange/issues?q=is%3Aissue+label%3Alammps

The diff looks like it handles #806 

https://github.com/openforcefield/openff-interchange/pull/897 might already be done?

https://github.com/openforcefield/openff-interchange/issues/955 could use some TLC; all/mostly good ideas that we haven't gotten to yet

I think #998 is done

This is just what's obviously tracked - anything else we should add @timbernat?

timbernat:
Regarding the tracked changes:
- #806  remains hard-coded AFAIK (see [here](https://github.com/openforcefield/openff-interchange/blob/main/openff/interchange/components/mdconfig.py#L359)), would be good to parameterize name of LAMMPS datafile
- #897 seems plausible, though it may be worth revisiting how constraints could be applied vs [how they currently are](https://github.com/openforcefield/openff-interchange/blob/main/openff/interchange/components/mdconfig.py#L364-L379). I'll consult with our collaborating LAMMPS expert and circle back to this...
-  #955 is a really good one, I have [a directory of reference files](https://github.com/timbernat/NREL_polymers/tree/main/lammps_file_ref) which show desired formats as written by a LAMMPS expert and their Interchange equivalents. This would be a good place to start for these changes, but I haven't gotten the chance to parse these files in details yet
- #998 is now resolved and closed, yes

Some other changes not mentioned above:
- #1001 we've already marked this, but it would be very nice to support passing of IO-like objects in place of strings/Paths for LAMMPS files in [`to_lammps()`](https://github.com/openforcefield/openff-interchange/blob/main/openff/interchange/interop/lammps/export/export.py#L14-L19), particularly to streamline unit testing
- #903 I wonder if there are other places which can be rewritten to use the Python API for lammps (I don't know off the top of my head)
- A very low-priority change, but one which I think would be nice to have is explicit specification of [LAMMPS units styles](https://docs.lammps.org/units.html) beyond [just "real"](https://github.com/openforcefield/openff-interchange/blob/main/openff/interchange/components/mdconfig.py#L284). I have [my own reference implementation](https://github.com/timbernat/polymerist/blob/main/polymerist/mdtools/lammpstools/unitstyles.py) of this which could be modified to used pint (rather than OpenMM) units, if the will to include this is there and a good home for this functionality can be found. I suspect the desire for such a change is dependent on peoples' standard use case for LAMMPS.

I'll add others as I think of them!

mattwthompson:
Looks like I botched the merge and overwrote part of https://github.com/openforcefield/openff-interchange/pull/1011/commits/c662e44fd4e1bc2627f26535ddf4f581878b8ead

mattwthompson:
I think the data file in `write_lammps_input` is fine on this branch: https://github.com/openforcefield/openff-interchange/blob/lammps-improvements/openff/interchange/components/mdconfig.py#L364

The Python API seems good (a little funky since it's not written in Python) but there are some ambiguities in whether or not importing it makes this package GPL. It'd be wonderful to use [LGPL releases](https://docs.lammps.org/Intro_opensource.html#lgpl-version-of-lammps), but I suspect they don't actually want to make those.

Are any of these other changes particularly useful for you sooner than later, and are any of them ones you'd like to tackle?

timbernat:
#955 and #897 are the most important for my purposes, I'm happy to take the helm on those issues once I circle back to them

j-wags:
@timbernat - Just checking on your plans/status here. This isn't holding up anyone else's work, I've just seen this on our trello board for a few weeks and want to make sure you have what you'd need to move forward here. Happy to offer guidance/schedule a quick screenshare if that'd help!

timbernat:
@j-wags Sorry for the long turnaround on this! It has been pushed down my TO-DO stack over the last few weeks as other projects have bubbled up. 

It'd be really useful for me to get an update on the status/relevance of some of these items, particularly as there's been a lot of related activity on #1027 (and other places that I'm probably not aware of); I want to make sure I can preferentially make contributions for which there is a) higher demand, and b) lower bandwidth from OpenFF core devs. Also, seems like there were some lingering loose ends with #1013 which were similar (but different) to the [LAMMPS ID writer issue I PR'ed](https://github.com/openforcefield/openff-interchange/pull/1000) earlier, I'd be happy to pitch in there as well, if it's warranted.

I'll be out for part of next week, but can set aside time the week of 09/02 to diff between the current Interchange LAMMPS outputs and the examples I have and lay out exactly where/what changes to constraint handling and bond/angle styles are needed. I'd be happy to set up a meeting if I get stuck there, and in either case, please keep me looped in!

PR #1013, opened 2024-07-17
Fixes residue ID increment in GROMACS GRO files

PEFrankel:
### Description

Addresses failure to increment the residue ID of identical molecule types in .gro files. Similar to #1000. 

### Checklist

- [x] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/interop/gromacs/export/_export.py
+++ b/openff/interchange/interop/gromacs/export/_export.py
@@ -445,6 +445,7 @@ class GROMACSWriter(DefaultModel):
         gro.write(f"{n_particles}\n")
 
         count = 0
+        residue_id = 1
         for molecule_name, molecule in self.system.molecule_types.items():
             n_copies = self.system.molecules[molecule_name]
 
@@ -456,7 +457,8 @@ class GROMACSWriter(DefaultModel):
                         f"%{decimal + 5}.{decimal}f"
                         f"%{decimal + 5}.{decimal}f\n"
                         % (
-                            atom.residue_index,  # This needs to be looked up from a different data structure
+                            residue_id = 1,
+                            # (hidden: 1013) atom.residue_index,  # This needs to be looked up from a different data structure
                             atom.residue_name[:5],
                             atom.name[:5],
                             (count + 1) % 100000,
@@ -467,6 +469,7 @@ class GROMACSWriter(DefaultModel):
                     )
 
                     count += 1
+                residue_id += 1
 
         if self.system.box is None:
             warnings.warn(
--- a/openff/interchange/_tests/unit_tests/interop/gromacs/export/test_export.py
+++ b/openff/interchange/_tests/unit_tests/interop/gromacs/export/test_export.py
@@ -192,6 +192,27 @@ class TestGROMACSGROFile(_NeedsGROMACS):
         openmm_atom_names = openmm.app.GromacsGroFile("atom_names.gro").atomNames
 
         assert openmm_atom_names == pdb_atom_names
+        
+    def test_residue_id_increment(self):
+        """Test that residue IDs increment properly for multiple molecules."""
+
+        mol1 = MoleculeWithConformer.from_smiles("CCO")
+        mol2 = MoleculeWithConformer.from_smiles("CCO")
+
+        for atom in mol1.atoms:
+            atom.metadata["residue_name"] = "MOL1"
+        for atom in mol2.atoms:
+            atom.metadata["residue_name"] = "MOL2"
+
+        interchange = Interchange.from_smirnoff(ForceField("openff-2.0.0.offxml"), [mol1, mol2])
+        interchange.to_gro("resid.gro")
+
+        with open("resid.gro", "r") as gro_file:
+            lines = gro_file.readlines()
+            residue_ids = {int(line[:5].strip()) for line in lines[2:-2]}
+
+            # expecting residue IDs 1 and 2 for MOL1 and MOL2 respectively
+            assert residue_ids == {1, 2}
 
 
 class TestGROMACS(_NeedsGROMACS):
--- a/openff/interchange/_tests/unit_tests/interop/gromacs/export/test_export.py
+++ b/openff/interchange/_tests/unit_tests/interop/gromacs/export/test_export.py
@@ -192,7 +192,7 @@ class TestGROMACSGROFile(_NeedsGROMACS):
         openmm_atom_names = openmm.app.GromacsGroFile("atom_names.gro").atomNames
 
         assert openmm_atom_names == pdb_atom_names
-        
+
     def test_residue_id_increment(self):
         """Test that residue IDs increment properly for multiple molecules."""
 
@@ -204,10 +204,12 @@ class TestGROMACSGROFile(_NeedsGROMACS):
         for atom in mol2.atoms:
             atom.metadata["residue_name"] = "MOL2"
 
-        interchange = Interchange.from_smirnoff(ForceField("openff-2.0.0.offxml"), [mol1, mol2])
+        interchange = Interchange.from_smirnoff(
+            ForceField("openff-2.0.0.offxml"), [mol1, mol2]
+        )
         interchange.to_gro("resid.gro")
 
-        with open("resid.gro", "r") as gro_file:
+        with open("resid.gro") as gro_file:
             lines = gro_file.readlines()
             residue_ids = {int(line[:5].strip()) for line in lines[2:-2]}
 


openff/interchange/interop/gromacs/export/_export.py
```diff
@@ -456,7 +457,8 @@ def _write_gro(self, gro, decimal: int):
                         f"%{decimal + 5}.{decimal}f"
                         f"%{decimal + 5}.{decimal}f\n"
                         % (
-                            atom.residue_index,  # This needs to be looked up from a different data structure
+                            residue_id = 1,
+                            # (hidden: 1013) atom.residue_index,  # This needs to be looked up from a different data structure
```
mattwthompson, line 461: Could you help me understand why the residue number should be hardcoded instead of looked up from the topology? Sometimes residue information is meaningless but it's often useful, and modifying it at export time is something I'd like to avoid

timbernat:
A few observations after having a poke around, based on the [new unit test provided](https://github.com/openforcefield/openff-interchange/blob/fix-1013/openff/interchange/_tests/unit_tests/interop/gromacs/export/test_export.py#L196-L217):
- The non-uniqueness issue is already present in the [molecule_types](https://github.com/openforcefield/openff-interchange/blob/fix-1013/openff/interchange/interop/gromacs/models/models.py#L328-L331) field of a [GROMACSSystem](https://github.com/openforcefield/openff-interchange/blob/fix-1013/openff/interchange/interop/gromacs/models/models.py#L296) created from an Interchange via [`interchange.smirnoff._gromacs._convert()`](https://github.com/openforcefield/openff-interchange/blob/fix-1013/openff/interchange/smirnoff/_gromacs.py#L49-L313), which happens [*prior to* any to_gro()/to_top() writer calls](https://github.com/openforcefield/openff-interchange/blob/fix-1013/openff/interchange/components/interchange.py#L453-L460); in other words, **the issue is upstream of the writers**, and the changes in [this commit](https://github.com/openforcefield/openff-interchange/commit/357aa315b15bea8102a4a67f50fa8b9c684477d2) from @PEFrankel are likely unnecessary
- Introducing a second pair of chemically-identical molecules which are distinct to the pair in the unit test writes only the first of this new pair and the original pair, respectively, to the .gro file produced by the writer
- Looking through how molecule_types is constructed, one finds an iteration over the [*allegedly* unique molecules in the interchange topology](https://github.com/openforcefield/openff-interchange/blob/fix-1013/openff/interchange/smirnoff/_gromacs.py#L117-L118); in fact, this [happens twice](https://github.com/openforcefield/openff-interchange/blob/fix-1013/openff/interchange/smirnoff/_gromacs.py#L195-L196)
- However, closer inspection of the [unique_molecule_map](https://github.com/openforcefield/openff-interchange/blob/fix-1013/openff/interchange/smirnoff/_gromacs.py#L86-L89), derived from [toolkit.Topology.identical_molecule_groups](https://github.com/openforcefield/openff-toolkit/blob/main/openff/toolkit/topology/topology.py#L1097-L1157), reveals that the molecule indices being iterated over here are only for the first occurence of each unique mol. In other words, **the wrong molecule index is being read from the identical_molecule_groups**; one should be reading the first field of the value (NOT the key) to index each individual occurrence of a molecule uniquely.

I tried to hack together a fix based on this, but encountered an [`AssertionError` in gromacs.export._export ](https://github.com/openforcefield/openff-interchange/blob/fix-1013/openff/interchange/interop/gromacs/export/_export.py#L436-L439) which indicated the number of molecules was being *overspecified* exactly by the amount of duplicates. My best guess would be that parts of the GROMACSSystem definition *do* need to be done on just the prototype molecule for a given chemistry (e.g. defining atom types), while others needs to be done for each instance (e.g. initializing GROMACSMolecule and GROMACSAtom instances), but it's unclear which parts of the duplicated unique_molecule_types iterations need to be done where, perhaps @mattwthompson can shed some light on this.

Hopefully this has been useful in pinning down the issue! I'm happy to look some more otherwise

mattwthompson:
Just a quick thought before I dig into this later - topology files need unique molecules and can't store information about copies, but coordinate files need all copies and don't care particularly about any de-duplication. I think the `.gro` file writer reaching into information stored for the `.top` file isn't the best way to go, and that might be what I did

mattwthompson:
Okay, there's a lot going on here and I need to be extra careful to only fix what needs fixing. I've opened a new issue to track requirements for various use cases. I am asking for input at the linked issue, specifically
* if I'm factually incorrect about anything (it wouldn't be the first time)
* if I'm missing any important use cases
* if there are other edge cases that would be useful to add as tests

https://github.com/openforcefield/openff-interchange/issues/1016

openff/interchange/interop/gromacs/export/_export.py
```diff
@@ -445,6 +445,7 @@ def _write_gro(self, gro, decimal: int):
         gro.write(f"{n_particles}\n")
 
         count = 0
+        residue_id = 1
```
mattwthompson, line 448: We cant' always start residue ID at 1, this is often a different value in user inputs and ought to be respected, i.e. this protein which starts a 3 for whatever reason https://github.com/openforcefield/protein-ligand-benchmark/blob/a0b67b93c4f12f63a8bb417a1f6e4df4dc10c6c5/data/shp2/01_protein/crd/protein.pdb#L591-L621

PEFrankel: Believed you had moved it to #1016 or #1024 / planned on encountering it there. I'll review this again then. 

It doesn't matter, but I would want to know "whatever reason" this example starts at 3 if this were the fix.

mattwthompson: I don't know why it's the case, but it's common enough in biophysics that we need to support it, i.e.

* https://github.com/openforcefield/protein-ligand-benchmark/blob/a0b67b93c4f12f63a8bb417a1f6e4df4dc10c6c5/data/tyk2/01_protein/crd/protein.pdb#L332-L337
* https://github.com/openforcefield/protein-ligand-benchmark/blob/a0b67b93c4f12f63a8bb417a1f6e4df4dc10c6c5/data/cdk2/01_protein/crd/protein.pdb#L623-L628

PLB is a relatively high-quality dataset that's valued, if not endorsed in some capacity, by consortium alumni and industry partners. It confuses me, like plenty of PDB files, but it's at least a concrete starting point to work on.

Note that this applies to residue IDs defined in biopolymers, not necessarily applicable to non-biological polymers and certainly not applicable to cases in which Interchange applies "residue IDs" on the fly just to make GROMACS happy. One of the details here that's been a thorn in my side for a decade or so is that these engines inconsistently handle residue and molecule indices, almost none handling them both at the same time.

PEFrankel: @mattwthompson 
An additional loop appears to be necessary within the [existing one](https://github.com/openforcefield/openff-interchange/blob/main/openff/interchange/smirnoff/_gromacs.py#L195) which covers unique_molecules in order to loop through the topology_mol_idx in the lists:
example: 0: [. (0, {0: 0, 1: 1, 2: 2}),  <--the existing loop only accesses the unique_molecule but we need the list int.
                        (1, {0: 1, 1: 0, 2: 2})]...
This can be achieved via:
```
        list_of_mol_tuples = unique_molecule_map[unique_molecule_index]
        for (topology_mol_idx, atom_map) in list_of_mol_tuples:
            unique_molecule_index = topology_mol_idx    # lazy syntax assignment
```
...then in [GROMACSAtom](https://github.com/openforcefield/openff-interchange/blob/main/openff/interchange/smirnoff/_gromacs.py#L228):
        `residue_index=unique_molecule_index + 1`

This change prints the maximum molecule int in the system (originally all the minimum) which could indicate a few things: 

1. Additional loop logic is needed.
2. The "writer" either only checks the final unique_molecule or is overwritten by each new identical molecule.
3. GROMACSMolecule contains one GROMACSAtom set/molecule instance that overwrites itself before sending to the system.

I checked the last possibility by reviewing the contents of GROMACSMolecule _after_ each append below GROMACSAtom (I checked this a number of ways, but this seems to be the most telling. Printing the GROMACSMolecule ID is a little confusing):
            print(f"After append: unique_molecule.name = {unique_molecule.name}")
            for a in molecule.atoms:
                print(a)

This result interestingly does increment the residue_index correctly (testing with the original loop will only throw one instance of course) which suggests it may be the writer or a process nearby upstream. Regardless, I think the loop here does need to be changed, possibly in addition to a writer fix.

Also, both loops in _gromacs.py may need to be changed ([line 113](https://github.com/openforcefield/openff-interchange/blob/main/openff/interchange/smirnoff/_gromacs.py#L113)), but the errors I got from not doing so were inconsistent.

I think I'm quite close and I could try to track down the writer in comp/[interchange](https://github.com/openforcefield/openff-interchange/blob/main/openff/interchange/components/interchange.py#L494) which my debugger accesses as the next path, but you may be able to solve this far faster than me (especially if the loop is the issue still). Let me know your thoughts. I spent a while on this and to little avail since I accidently tracked `deepcopy` and `add_mol_keep_cache` for molecules before returning to this file thinking it would help me understand the background.

mattwthompson: To be frank, there's too much going on here for me to follow, and I don't fully understand the behavior you're trying to fix. If you can open an issue with a minimally reproducing example demonstrating a difference between expectation and behavior, that would greatly help me understand the objective here.

I think #1024 there are changes you might be after? In particular I think this test looks at residue IDs in `.gro` files when multiple copies of a unique molecule are present

https://github.com/openforcefield/openff-interchange/pull/1024/files#diff-796291360ccaaab03d27aab793674a9c3b447a64c3f6fd2402b600611f1e36f5R200-R214

PR #1023, opened 2024-07-30
Fix indexing in vdW lookup of atoms in GROMACS export

### LINKED ISSUE #1022: GROMACS export crashes due to non-bonded lookup failures when virtual sites are present in some topologies
**Description**

Trying to use Interchange and packmol to write out a solvated gromacs topology. 
Using the same code as the minimally reproducible example below but changing out the FF for sage 2.0, 2.1 or 2.2 I have no issues.
I also have no issues if you replace `solv_top = solvate_topology(mol.to_topology())` with just `solv_top = mol.to_topology()`.
I believe the issue has to do with the combination of the virutal sites forcefield and packmol.

**Reproduction**
[example-sigma-hole-bromine.offxml.zip](https://github.com/user-attachments/files/16432814/example-sigma-hole-bromine.offxml.zip)

```

# create molecule with virtual site
mol = Molecule.from_smiles("CBr")
mol.generate_conformers()

# assign parameters using openff toolkit -> openff interchange

ff = ForceField("openff-2.0.0.offxml", "example-sigma-hole-bromine.offxml")
solv_top = solvate_topology(mol.to_topology())
ic = ff.create_interchange(topology=solv_top)

ic.to_top("cbr.top")
ic.to_gro("cbr.gro")
```

**Output**

```
KeyError                                  Traceback (most recent call last)
Cell In[34], line 3
      1 # write out to gromacs with default 5 nm box
----> 3 ic.to_top("cbr.top")
      4 ic.to_gro("cbr.gro")

File ~/mambaforge/envs/openff-toolkit/lib/python3.12/site-packages/openff/interchange/components/interchange.py:493, in Interchange.to_top(self, file_path, hydrogen_mass, _merge_atom_types)
    489 from openff.interchange.interop.gromacs.export._export import GROMACSWriter
    490 from openff.interchange.smirnoff._gromacs import _convert
    492 GROMACSWriter(
--> 493     system=_convert(self, hydrogen_mass=hydrogen_mass),
    494     top_file=file_path,
    495 ).to_top(_merge_atom_types=_merge_atom_types)

File ~/mambaforge/envs/openff-toolkit/lib/python3.12/site-packages/openff/interchange/smirnoff/_gromacs.py:143, in _convert(interchange, hydrogen_mass)
    140 topology_index = particle_map[interchange.topology.atom_index(atom)]
    141 key = TopologyKey(atom_indices=(topology_index,))
    142 vdw_parameters = vdw_collection.potentials[
--> 143     vdw_collection.key_map[key]
    144 ].parameters
    145 charge = electrostatics_collection.charges[key]
    147 # Build atom types

KeyError: TopologyKey with atom indices (1300,)
```

**Software versions**

     - Sonoma 14.2.1 on M1 Mac
     - `mamba create -n openff-toolkit -c conda-forge openff-toolkit`
```
openeye-toolkits          2024.1.0                py312_0    openeye
openff-amber-ff-ports     0.0.4              pyhca7485f_0    conda-forge
openff-forcefields        2024.07.0          pyhff2d567_0    conda-forge
openff-interchange        0.3.28             pyhd8ed1ab_0    conda-forge
openff-interchange-base   0.3.28             pyhd8ed1ab_0    conda-forge
openff-models             0.1.2              pyhca7485f_0    conda-forge
openff-toolkit            0.16.2             pyhd8ed1ab_0    conda-forge
openff-toolkit-base       0.16.2             pyhd8ed1ab_0    conda-forge
openff-units              0.2.2              pyhca7485f_0    conda-forge
openff-utilities          0.1.12             pyhd8ed1ab_0    conda-forge
openmm                    8.1.2           py312hacce00b_2_khronos    conda-forge
```

mattwthompson:
### Description

Resolves #1022

I'm confused how the ligand-in-water example was working previously, since this may cause the incorrect non-bonded (both vdW and electrostatics) parameters to be looked up on molecules placed in a topology after a virtual site-bearing molecule.

### Checklist

- [x] Add tests
- [x] Lint
- [x] Update docstrings

### DIFF
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -11,9 +11,10 @@ Dates are given in YYYY-MM-DD format.
 
 Please note that all releases prior to a version 1.0.0 are considered pre-releases and many API changes will come before a stable release.
 
-## Current development
+## 0.3.29 - 2024-08-01
 
 * #1023 Fixes a bug in which non-bonded parameter lookup sometimes crashed when virtual sites were present.
+* #1018 #1019 Improve performance of `Interchange.to_top`, particularly after importing from OpenMM objects.
 
 ## 0.3.28 - 2024-07-17
 
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -11,6 +11,10 @@ Dates are given in YYYY-MM-DD format.
 
 Please note that all releases prior to a version 1.0.0 are considered pre-releases and many API changes will come before a stable release.
 
+## Current development
+
+* #1023 Fixes a bug in which non-bonded parameter lookup sometimes crashed when virtual sites were present.
+
 ## 0.3.28 - 2024-07-17
 
 * #991 Fixes the virtual site example for upstream changes in virtual site parameter lookup.
--- a/openff/interchange/_tests/__init__.py
+++ b/openff/interchange/_tests/__init__.py
@@ -1,11 +1,12 @@
 """Assorted utilities used in tests."""
 
 import pathlib
+import random
 from importlib import resources
 
 import numpy
 import pytest
-from openff.toolkit import Molecule
+from openff.toolkit import ForceField, Molecule, Topology
 from openff.toolkit.utils import (
     AmberToolsToolkitWrapper,
     OpenEyeToolkitWrapper,
@@ -13,6 +14,7 @@ from openff.toolkit.utils import (
 )
 from openff.utilities.utilities import has_executable, has_package
 
+from openff.interchange import Interchange
 from openff.interchange.drivers.gromacs import _find_gromacs_executable
 
 if has_package("openmm"):
@@ -110,3 +112,38 @@ needs_not_sander = pytest.mark.skipif(
     HAS_SANDER,
     reason="sander needs to NOT be installed",
 )
+
+
+def shuffle_topology(
+    interchange: Interchange,
+    force_field: ForceField,
+    seed: int = 123455,
+) -> Interchange:
+    """Shuffle the atom order of the topology and return a new Interchange object."""
+    assert interchange.positions is not None, "Written to also manage positions"
+
+    rng = random.Random(seed)
+
+    shuffled_topology = Topology()
+
+    # can't easily shuffle a generator ...
+    for molecule in interchange.topology.molecules:
+        topology_indices = [
+            interchange.topology.atom_index(atom) for atom in molecule.atoms
+        ]
+
+        molecule._conformers = [
+            interchange.positions[topology_indices, :].to("angstrom"),
+        ]
+
+        # so just randomize order at insertion time
+        shuffled_topology._molecules.insert(
+            rng.randint(0, len(shuffled_topology._molecules)),
+            molecule,
+        )
+
+    shuffled_topology._invalidate_cached_properties()
+
+    shuffled_topology.box_vectors = interchange.topology.box_vectors
+
+    return force_field.create_interchange(shuffled_topology)
--- a/openff/interchange/_tests/data/example-sigma-hole-bromine.offxml
+++ b/openff/interchange/_tests/data/example-sigma-hole-bromine.offxml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<SMIRNOFF version="0.3" aromaticity_model="OEAroModel_MDL">
+    <VirtualSites version="0.3" exclusion_policy="parents">
+        <!-- A chlorine bound to (any) carbon, for fixing sigma holes -->
+        <VirtualSite
+            smirks="[#6:2]-[#35X1:1]"
+            epsilon="0.0 * kilojoules_per_mole ** 1"
+            type="BondCharge"
+            match="all_permutations"
+            distance="1.45 * angstrom ** 1"
+            outOfPlaneAngle="None"
+            inPlaneAngle="None"
+            charge_increment1="-0.063 * elementary_charge ** 1"
+            charge_increment2="0.0 * elementary_charge ** 1"
+            sigma="0.0 * angstrom ** 1"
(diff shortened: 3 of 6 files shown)


mattwthompson:
Let's see if I can explain this with a little more context. This is mostly for my own purposes, but I welcome feedback on either the design decisions I made along the way or (see the end) test cases that will protect us moving forward.

---

Internally, Interchange uses the OpenFF Toolkit's representation for all molecules, whether or not they have virtual sites applied to them by the force field (or HMR, or SETTLES, or any number of things). At export time, virtual sites need to be added, and there are two ways this may happen.

* Put all virtual sites at the end of the topology. This is how the OpenMM export is built; though OpenMM should be able to handle it and the following case, it made more sense to us to do this given the way OpenMM tracks _all_ particles in a system with no de-duplication. The objects/XML lines in OpenMM objects/files manage the bookkeeping of which virtual sites are attached to which (massive) particles.
* Put the virtual sites and the end _of each molecule_. This is how the GROMACS export is built, since GROMACS de-duplicates systems into their constituent molecules, it doesn't make any sense to put all virtual sites at the end.

In either case, their needs to be a mapping between the "topology" indices (the index in the `Topology` object, which does not store virtual sites) and some more ~system-level representation of all particles, which include virtual sites. This is handled by [this function](https://github.com/openforcefield/openff-interchange/blob/v0.3.28/openff/interchange/interop/common.py), which takes in a bunch of arguments and returns a mapping between atoms indices (ints representing an atom's index in the topology) or virtual site keys to "particle" indices (ints representing the richer "system-level" index, which includes both atoms and virtual sites). So the return value is typed as `dict[int | VirtualSiteKey, int]`. For systems without virtual sites, this is a silly-looking mapping between ints of identical values, but once any virtual site is introduced into the system (even on waters) it becomes crucial for tracking the relationship between virtual sites, their parent atom, their orientation atoms, and the non-bonded parameters of all involved particles.

Here's what that mapping looks like for a system like reported in #1022 (sorry this is a bit long ... the payoff is right at the end):

```python
from collections import defaultdict
from openff.toolkit import Molecule, ForceField, Topology

from openff.interchange.models import VirtualSiteKey
from openff.interchange.interop.common import _build_particle_map
from openff.interchange.interop._virtual_sites import (
    _virtual_site_parent_molecule_mapping,
)


def get_particle_map(interchange) -> dict[int | VirtualSiteKey, int]:
    virtual_site_molecule_map: dict[
        VirtualSiteKey,
        int,
    ] = _virtual_site_parent_molecule_mapping(interchange)

    molecule_virtual_site_map: dict[int, list[VirtualSiteKey]] = defaultdict(list)

    for virtual_site_key, molecule_index in virtual_site_molecule_map.items():
        molecule_virtual_site_map[molecule_index].append(virtual_site_key)

    return _build_particle_map(
        interchange,
        molecule_virtual_site_map,
        collate=True,
    )


topology = Topology.from_molecules(
    [
        Molecule.from_smiles(smi)
        for smi in [
            "CBr",
            "O",
            "[Na+]",
            "[Br-]",
        ],
    ],
)

interchange = ForceField(
    "openff-2.0.0.offxml",
    "example-sigma-hole-bromine.offxml",
).create_interchange(topology)

particle_map = get_particle_map(interchange)

for key, val in particle_map.items():
    if isinstance(key, int):
        print((key, val))
    if isinstance(key, VirtualSiteKey):
        print((key.orientation_atom_indices, val))
"""
(0, 0)  # C
(1, 1)  # Br
(2, 2)  # H
(3, 3)  # H
(4, 4)  # H
((1, 0), 5)   #  <-- here's the virtual site, between atoms (C, Br) and a virtual site with (system-level) index 5
(5, 6)  # O in water, notice the "system" index is now the "topology" index +1, because of that virtual site coming earlier in the topology
(6, 7)  # H
(7, 8)  # H
(8, 9)
(9, 10)
"""
```

The quirk that @megosato stumbled across happens only when the _first_ copies of some unique molecule(s) appear at the _end_ of the topology [^1], one that I'm a little surprised hasn't caused problems so far. It's possible this combination of inputs just hasn't been tested before.

An intersecting question here is how the vdW data should be keyed - should the atoms use the "topology" or atom index, or the "system" or particle index? `Interchanges` (well, their vdW collections) use the "topology" index and this makes the most sense to me. The GROMACS code used the "particle" index - the only substantive change in this PR is to switch that - effectively using the second value in each tuple above instead of the first. (The other changes are a test and comments clarifying some of this too-long comment.) This causes problems at the _end_ of a topology like above, which has 10 atoms and 11 particles; there's no way to look up the parameters associated with the 11th (0-index 10) atom when there isn't one.

This did _not_ affect a system like a ligand (without virtual sites) in OPC/TIP4P water (without ions) because GROMACS-style atom-deduplication means the last few particles in the topology aren't ever looked at.

This change appears to fix the linked MWE, but has me super spooked. Something so simple as parameter lookup shouldn't be failing at this stage in the project for a system as relatively simple as a ligand in solvent. I've been thinking about how this could have been prevented. Short of the literal case that was reported, I think a general strategy that could catch some of these cases is to randomly shuffle the order of molecule(s) in some simple systems. There are some conventions around atom ordering (protein, ligand, other fun stuff, and then solvent) but, in principle, energy evaluations should be completely independent. Given how integral molecule de-duplication is to GROMACS bookkeeping, I think it'd also be useful to have a mix of single and many copies of chemically unique molecules. (Again, it shouldn't make a difference if there's one cation or a dozen, but bugs can make these mundane details matter. And some bugs are hard to spot without being tested.)

[^1]: Though I'm pretty sure this bug should affect any system in which virtual site-bearing ligands aren't in the last unique molecule. So a topology beginning with a ligand with virtual site parameters **should** be affected, but a ligand without virtual sites listed before 4-site water **should not** be affected. I haven't tested this specifically.

PR #1034, opened 2024-08-15
Do not write vacuum simulations to GROMACS files

### LINKED ISSUE #1029: Status of vacuum simulations in GROMACS
**Description**
Recent (~2020 onwards) versions of GROMACS do not properly support vacuum simulations. The developers [are aware of this](https://gitlab.com/gromacs/gromacs/-/issues/3526) and how it has been a limitation for its use in force field development. Users [generally work around this by setting the box](https://gromacs.bioexcel.eu/t/vacuum-simulations-in-gromacs-2020/7249) to be some arbitrarily large value, but this is a hack and has downsides.

This is a little downstream of Interchange[^1] but we run into it in tests and can't get around it as long as we want a `.to_gromacs` method that also includes the `.mdp` file. Currently, these files are written out but the user is warned about this issue and [arbitrarily sets the box vectors](https://github.com/openforcefield/openff-interchange/blob/v0.3.29/openff/interchange/interop/gromacs/export/_export.py#L471-L479) to be a 5 nanometer cube. I'd like to move to a cleaner solution with the behavior changes afforded to us in the 0.4 release, of which we have thought of a few options in this case. **In the case of a non-periodic `Interchange`, i.e. one without box vectors defined, we could:

1. Leave as-is
2. Do not write (any) GROMACS files if the system is periodic
3. Do not write some (`.mdp` and `.gro`?) files, but write other (`.top` ?) files
4. Write any files, expecting the simulations themselves to crash, and just document this quirk

Josh @Yoshanuikabundi [brought in several useful details](https://github.com/openforcefield/openff-interchange/pull/1027#issuecomment-2276872082) for context:

* Users could install old (2019 or prior) versions which have support
* GROMACS itself may add support back in a future release

These generally support the case for (4), which I think is their preference. I think I like that option as well, but I'd be fine with (2) or (3). As long as there's some error raised in this ~unsupported use case, I'm happier.

[^1]: Strictly speaking, it's fine to write out [`pbc = no`](https://manual.gromacs.org/documentation/2018/user-guide/mdp-options.html#mdp-value-pbc=no), but then trying to use _use_ it requires `cutoff-scheme = group` which is unsupported ... but confusingly a [allowed setting](https://manual.gromacs.org/documentation/2020/user-guide/mdp-options.html#mdp-value-pbc=no).

mattwthompson:
### Description

This PR removes support for writing files to GROMACS simulations if no box vectors are defined.

Resolves #1029 

### Checklist

- [x] Add tests
- [x] Lint
- [ ] Update docstrings

### DIFF
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -11,6 +11,14 @@ Dates are given in YYYY-MM-DD format.
 
 Please note that all releases prior to a version 1.0.0 are considered pre-releases and many API changes will come before a stable release.
 
+## 0.4.0 - 2024
+
+* Pydantic v2 is now used at runtime. As a consequence, models containing `Interchange`s cannot also use models from the v1 API.
+* `Interchange.to_gromacs` and similar methods now raise an error if no box is defined.
+  * Previously, this was a warning.
+  * GROMACS dropped support for (proper) vacuum simulations in version 2020 and there are no immediate plans to re-introduce it.
+  * Users freqently approximate vacuum simulation with periodic boundary conditions by applying a large box. This has some performance issues and some non-bonded terms likely differ numerically compared to similar implementations in other engines.
+
 ## 0.3.29 - 2024-08-01
 
 * #1023 Fixes a bug in which non-bonded parameter lookup sometimes crashed when virtual sites were present.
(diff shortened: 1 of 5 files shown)


mattwthompson:
I still need to document this (in each branch)

PR #1035, opened 2024-08-16
Fix access to `_NonbondedData` NamedTuple

JMorado:
### Description

I'm not certain if `_NonbondedData` was implemented as a dictionary in earlier versions of the code, but it is now a `NamedTuple`. Therefore, it should be accessed using either integer indices or attribute names. This PR addresses this issue. The error I encountered was the following:

```python                                    │
│ /home/joaomorado/opt/micromamba/envs/fes-ml-dev/lib/python3.10/site-packages/openff/interchange/ │
│ interop/openmm/_nonbonded.py:345 in _create_single_nonbonded_force                               │
│                                                                                                  │
│    342 │   │   │   non_bonded_force.setNonbondedMethod(openmm.NonbondedForce.LJPME)              │
│    343 │   │   │   non_bonded_force.setEwaldErrorTolerance(ewald_tolerance)                      │
│    344 │   │                                                                                     │
│ ❱  345 │   │   elif data["vdw_method"] == data["electrostatics_method"] == "cutoff":             │
│    346 │   │   │   if data["vdw_cutoff"] != data["electrostatics_collection"].cutoff:            │
│    347 │   │   │   │   raise UnsupportedExportError(                                             │
│    348 │   │   │   │   │   "If using cutoff vdW and electrostatics, cutoffs must match.",        │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
TypeError: tuple indices must be integers or slices, not str
```

### DIFF
--- a/openff/interchange/interop/openmm/_nonbonded.py
+++ b/openff/interchange/interop/openmm/_nonbonded.py
@@ -342,15 +342,15 @@ def _create_single_nonbonded_force(
             non_bonded_force.setNonbondedMethod(openmm.NonbondedForce.LJPME)
             non_bonded_force.setEwaldErrorTolerance(ewald_tolerance)
 
-        elif data["vdw_method"] == data["electrostatics_method"] == "cutoff":
-            if data["vdw_cutoff"] != data["electrostatics_collection"].cutoff:
+        elif data.vdw_method == data.electrostatics_method == "cutoff":
+            if data.vdw_cutoff != data.electrostatics_collection.cutoff:
                 raise UnsupportedExportError(
                     "If using cutoff vdW and electrostatics, cutoffs must match.",
                 )
 
             non_bonded_force.setNonbondedMethod(openmm.NonbondedForce.CutoffPeriodic)
             non_bonded_force.setCutoffDistance(
-                to_openmm_quantity(data["vdw_cutoff"]),
+                to_openmm_quantity(data.vdw_cutoff),
             )
 
         else:
@@ -708,7 +708,7 @@ def _create_multiple_nonbonded_forces(
                             eps_14 = (eps1 * eps2) ** 0.5 * vdw_14
                         else:
                             raise UnsupportedExportError(
-                                f"Unsupported mixing rule: {data['mixing_rule']}",
+                                f"Unsupported mixing rule: {data.mixing_rule}",
                             )
 
                         # ... and set the 1-4 interactions


mattwthompson:
Thanks!

You guessed correctly - this was previously implemented as a dict.

I'm curious how you encountered this error, though? I expect that line would be covered by a test somewhere, but I could be wrong. Cut-off vdW _and_ electrostatics interactions is something that I don't see used much but try to test.

JMorado:
I created an `MDConfig` object, which I applied to an `interchange` one. The error occurred whenever I set the Coulomb method to 'cutoff.' You're right, I don't think it's as commonly used as PME, though RF can provide pretty accurate results and be cheaper as well. I was combining RF with VdW + dispersion correction.

mattwthompson:
That makes sense - looking at the coverage report, this slipped through the cracks because this case happens to be un-tested. This would be a good time to add such a test - I might not get to this until Monday but I should be able to take care of it.

mattwthompson:
Okay, I worked on writing up a test that hits this case. This is some murky territory with some sharp, pokey things hiding in the shadows. Some things I'm confused/hung up on

* I haven't (explicitly) implemented reaction-field electrostatics
  * OpenMM appears to always apply the reaction-field correction when "cutoff" electrostatics are used [^1]
  * These are distinct options in GROMACS [^2], [^3]
* Reaction-field electrostatics are called out in the SMIRNOFF spec as something that might be supported in the future, but I don't think there's any inertia there. Interchange doesn't support it, but (generally) allows "cutoff" electrostatics for testing purposes ... although that's also not in the spec yet [^4]

I bring this up only because the current implementation allows cut-off electrostatics when everything is grouped up into a single `NonbondedForce` but not when split out into different forces [^5]. This appears to be an inconsistency since electrostatics are handled by the same object type in either case; splitting out the forces just moves vdW and (all) 1-4 interactions into separate forces.

So ... this is the background context for me asking
1. Are you using `combine_nonbonded_forces=False`, `=True` (the default), or both? I loosely sense you're using the default, but maybe you have some funky systems.
3. Are your systems periodic? I figure yes, since you called out PME earlier.
4. What non-bonded method(s) do you want your force(s) to end up with? (`.getNonbondedMethod()`, etc.) I think `CutoffPeriodic` but I'm not sure.

Sorry for the landmine you've stumbled across - your changes are fine and stand on their own - but I do want to make sure that the test I add in makes some sense, and also that I don't turn this around only for you to find that you're hitting an unsupported case, or I'm wrong about how to set these non-bonded methods, or something else that causes extra headache. 

[^1]: https://github.com/openmm/openmm/issues/397
[^2]: https://manual.gromacs.org/2024-beta/user-guide/mdp-options.html#mdp-value-coulombtype-Cut-off
[^3]: https://manual.gromacs.org/2024-beta/user-guide/mdp-options.html#mdp-value-coulombtype-Reaction-Field
[^4]: https://github.com/openforcefield/standards/issues/6
[^5]: https://github.com/openforcefield/openff-interchange/blob/48359c42c9f4190901f166015498fa79a6cb34f5/openff/interchange/interop/openmm/_nonbonded.py#L899-L902

JMorado:
>  OpenMM appears to always apply the reaction-field correction when "cutoff" electrostatics are used [1](#user-content-fn-1-e09d468376f15956745f2bf92af04c83)

Correct. That's my use case.

>   These are distinct options in GROMACS [2](#user-content-fn-2-e09d468376f15956745f2bf92af04c83), [3](#user-content-fn-3-e09d468376f15956745f2bf92af04c83)

I personally think that if I were writing Gromacs files for an interchange where the Coulomb method is set to cutoff, I would not expect the use of RF. OpenMM works differently here: a hard periodic cutoff without RF requires the definition of a `CustomNonbondedForce`, I believe. 

Answering your questions:

> Are you using `combine_nonbonded_forces=False`, `=True` (the default), or both? I loosely sense you're using the default, but maybe you have some funky systems.

I'm using the default option.

> Are your systems periodic? I figure yes, since you called out PME earlier.

Yes, exactly.

> What non-bonded method(s) do you want your force(s) to end up with? (`.getNonbondedMethod()`, etc.) I think `CutoffPeriodic` but I'm not sure.

Yes, I would like to end up with `CutoffPeriodic`.

To be honest, the main reason I'm using this combination at the moment is for testing purposes (and in this regard, it would be amazing to have a hard cutoff working with OpenMM). I'm not really attempting to run heavy production runs. Let me know if I can be of further help.

JMorado:
> This feedback is super helpful! I'm going to try out the following changes
> 
>     * `periodic_method="reaction-field"` produces `NonbondedForce` with its method set to `CutoffPeriodic` whether or not non-bonded forces are split out
> 
>     * `periodic_method="cutoff"` raises an error
> 
>     * No changes to GROMACS implementation (for now)

Makes total sense to me!

> I'm generally curious if you get a hard cut-off working, given that it's caused some confusion with OpenMM users in the past (linked thread) and that's how it'd be implemented here if we want to go with it. I understand it's scientifically dubious but it's absolutely necessary to get tight energy comparisons between engines (https://link.springer.com/article/10.1007/s10822-016-9977-1) which is something that this project cares about. But we might not need to implement that for a while, and no pressure at all to continue down this rabbit hole 🙃

I think it's definitely a useful feature to have for testing purposes. There have already been discussions about how to construct a nonbonded electrostatic interaction with a hard (sharp) cut-off. For reference: https://github.com/openmm/openmm/issues/3281

mattwthompson:
I've included these changes in #1039, it'll be in the next release (this week or early next).

Thanks for your report, feedback, and rapid iteration!

Issue #1031, opened 2024-08-13
Loosing Gromacs atom name details when using an OpenMM topology


hannahbaumann:
When I create an interchange using `from_openmm` and I provide an OpenMM topology for the `topology=` option, and then write out the Gromacs .top and .gro file, the atom names of protein atoms are, e.g. C and not CA. When I instead provide an OpenFF topology in the `from_openmm` call, the protein atom names have the same names as in the .pdb file used, e.g. CA. Is this information lost in the OpenMM topology or would it be possible to also transfer the atom names through that one?

mattwthompson:
This is legit and just an oversight; most of my testing has ensured that energy evaluations are reproduced in these code paths and atom names aren't needed for that.

The short of it, I think is - 

* The toolkit has `Topology.from_openmm` / `Topology.from_pdb` that Jeff and others have invested a lot of effort into and probably does what users expect w/r/t atom names, so if you used the toolkit to load a PDB file the atom names are probably accurately defined
* Interchange has a thinner function that does the same thing when an OpenMM topology is passed to `from_openmm`, but does not route through the toolkit's solution and probably failed to process atom names from OpenMM for no reason other than oversight

I think I'm able to isolate what you're running into:

```python
"""
$ INTERCHANGE_EXPERIMENTAL=1 python issue_1031.py
"""

import openmm.app
from openff.toolkit import Topology
from openff.interchange import Interchange


def get_atom_names(topology: Topology) -> set[str]:
    return {atom.name for atom in topology.atoms}


pdb_file = openmm.app.PDBFile("protein.pdb")

interchange = Interchange.from_openmm(
    system=openmm.app.ForceField(
        "amber99sb.xml",
        "tip3p.xml",
    ).createSystem(
        pdb_file.topology,
        nonbondedMethod=openmm.app.PME,
    ),
    topology=pdb_file.topology,
)

# These are more likely to accurately represent the intended data in the PDB;
# JW/JM and friends put a lot of work into this
toolkit_topology = Topology.from_pdb("protein.pdb")

print(get_atom_names(toolkit_topology))
# {'HD13', 'NZ', 'CG1', 'HD21', 'HG21', 'HB1', 'HG2', 'HG23', 'HD12', 'HE2', 'OH', 'HA', 'H1', 'HA3', 'H3', 'HD11',
[...]

PR #1037, opened 2024-08-16
Test that atom names are processed in `Interchange.from_openmm`

mattwthompson:
### Description

This PR adds a sort of regression test for the behavior observed in #1031. The fix is actually in the toolkit (though in a pinch it would have been possible to add some bubblegum-and-paperclips here) and is expected to land in release 0.16.4 early next week (8/19 or so).

### Checklist

- [x] Add tests
- [x] Lint
- [ ] Update docstrings

### DIFF
--- a/devtools/conda-envs/beta_env.yaml
+++ b/devtools/conda-envs/beta_env.yaml
@@ -10,7 +10,7 @@ dependencies:
   - pydantic >=1.10.17,<3
   - openmm >=7.6
   # OpenFF stack
-  - openff-toolkit >=0.15.2
+  - openff-toolkit ~=0.16.4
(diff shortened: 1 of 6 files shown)


mattwthompson:
@hannahbaumann if you have some time, could you double-check that this test covers the behavior we're looking for?

This is the PDB file in the toolkit I (mostly arbitrarily) grabbed: https://github.com/openforcefield/openff-toolkit/blob/0.16.3/openff/toolkit/data/proteins/MainChain_HIE.pdb

mattwthompson:
I think this works now - again, this requires the toolkit release (0.16.4) that's already live and no changes in Interchange

PR #1051, opened 2024-09-17
Update topology/system particle bookkeeping in OpenMM

### LINKED ISSUE #1049: Created OpenMM Topology and System virtual site atom indices don't match
**Expectation**

The atom indices of an OpenMM Topology match those of its associated System, this includes virtual sites.

**Current status**

Creating an OpenMM Topology with virtual sites puts the virtual site at the end of each residue whilst the System has them completely at the end of the entire System. This means that if you have more than one residue, the atom indices don't match.

**Reproduction**

```python

ligand = Molecule.from_smiles('CCCCC')
ligand.generate_conformers()

off_top = Topology.from_molecules(ligand)
solvated_off_top = solvate_topology(topology=off_top)

inter = Interchange.from_smirnoff(topology=solvated_off_top, force_field=ff)
omm_top = inter.to_openmm_topology()
omm_sys = inter.to_openmm_system()

print(list(omm_top.atoms())[20].name == 'EP')
print(omm_sys.isVirtualSite(20) == True)
```

**Output**

```
True
False
```

**Software versions**

Interchange v0.3.27

mattwthompson:
### Description

Resolves #1049

### Checklist

- [x] If not collating, add each virtual site to its own residue at the end
- [x] Add tests
  - [x] Reproduce issue #1049
  - [x] Test topology indexing and without collating
- [x] Lint
- [x] Update documentation
  - [x] Update docstrings
  - [x] Add section to user guide on user expectations


openff/interchange/interop/openmm/_topology.py
```diff
@@ -173,6 +186,21 @@ def to_openmm_topology(
                 order=bond_order,
             )
 
+    if has_virtual_sites and not collate:
+        virtual_site_chain = openmm_topology.addChain(atom_chain_id)
+        virtual_site_residue = openmm_topology.addResidue("VS", virtual_site_chain)
```
IAlibay, line 191: From a practical standpoint, a reference to the original residue would possibly be useful - i.e. when I'm doing "solvent within N of Y" I'm looking for a residue name that matches SOL | WAT | HOH not VS.

openff/interchange/interop/openmm/_topology.py
```diff
@@ -173,6 +186,21 @@ def to_openmm_topology(
...
+
+        for molecule_index, molecule in enumerate(topology.molecules):
+            virtual_sites_in_this_molecule: list[VirtualSiteKey] = molecule_virtual_site_map[molecule_index]
```
IAlibay, line 194: Only briefly looking at things, but it looks to me like that we're really needing is this map. If we could map things to/from the collated form then yes we have to plan for this map being optionally around, but at the same time it's not crazy extra amounts of work to work around it.

openff/interchange/interop/openmm/_topology.py
```diff
@@ -15,9 +15,22 @@
...
-    """Create an OpenMM Topology containing some virtual site information (if appropriate)."""
+    """
+    Create an OpenMM Topology containing some virtual site information (if appropriate).
+
+    Parameters
+    ----------
+    interchange
+        The Interchange object to convert to an OpenMM Topology.
+    collate
+        If False, the default, all virtual sites will be added to a single residue at the end of the topology.
```
IAlibay, line 29: > single residue

I'm sure I'm missing something important here - is there a practical advantage to doing a single residue rather than a residue per molecule's worth of virtual sites?

IAlibay (commented):
Couple of very brief (i.e. had 10 mins to read the code) and probably way too specifically opiniated to OpenFE workfllows views.

openff/interchange/interop/openmm/_topology.py
mattwthompson: I'd like to expose a map like this to the user - I have to re-create it a number of times while exporting, might as well cache it on the `Interchange` itself or do something like that - but want to be careful and deliberate when making that change

openff/interchange/interop/openmm/_topology.py
mattwthompson: I think this would be solved by your suggestion later - that each "virtual site residue" be named according to the molecule/residue it's from?

openff/interchange/interop/openmm/_topology.py
IAlibay: > upper bound on residue numbers

I could see AMBER and GROMACS being issues here :/ (at least iirc AMBER is still technically fixed width in parm7).

mattwthompson: Updated with this idea in mind:

```python
In [30]: x = (
    ...:     ForceField("opc.offxml")
    ...:     .create_interchange(
    ...:         Topology.from_molecules(
    ...:             [
    ...:                 Molecule.from_smiles("O"),
    ...:                 Molecule.from_smiles("O"),
    ...:             ],
    ...:         ),
    ...:     )
    ...:     .to_openmm_topology()
    ...: )

In [31]: for residue in x.residues():
    ...:     print(residue.index, [*residue.atoms()])
0 [<Atom 0 (O1x) of chain 0 residue 0 (UNK)>, <Atom 1 (H1x) of chain 0 residue 0 (UNK)>, <Atom 2 (H2x) of chain 0 residue 0 (UNK)>]
1 [<Atom 3 (O1x) of chain 1 residue 1 (UNK)>, <Atom 4 (H1x) of chain 1 residue 1 (UNK)>, <Atom 5 (H2x) of chain 1 residue 1 (UNK)>]
2 [<Atom 6 (EP) of chain 2 residue 2 (VS)>]
3 [<Atom 7 (EP) of chain 2 residue 3 (VS)>]
```

The only thing to note being that each virtual site has its own residue (which would have multiple virtual sites for a molecule that has more than one virtual site - for example, two TIP5P waters would have 3, 3, 2, 2 "atoms" in each "residue")

IAlibay: I think that breakdown is great! That way you can quickly iterate through your residues and find the matching virtual sites.

openff/interchange/interop/openmm/_topology.py
```diff
@@ -173,6 +186,26 @@ def to_openmm_topology(
...
+    if has_virtual_sites and not collate:
+        # TODO: This structure is imperfect, no clue how to handle chains here
+        virtual_site_chain = openmm_topology.addChain(atom_chain_id)
+
+        for molecule_index, molecule in enumerate(topology.molecules):
+            virtual_sites_in_this_molecule: list[VirtualSiteKey] = molecule_virtual_site_map[molecule_index]
+
+            # make a new "residue" for each molecule which has virtual sites
+            if len(virtual_sites_in_this_molecule) > 0:
+                this_virtual_site_residue = openmm_topology.addResidue("VS", virtual_site_chain)
```
IAlibay, line 198: Would it make sense to name these residues by their parent residue name? The use case I'm thinking of here is let's say you want to select all your waters, you'd go "resname WAT HOH SOL" and know you're picking up all the waters and their virtual sites. Similarly if you knew you had a ligand named LIG, then your user experience (especially when you output to a PDB from an openmm topology via something like mdtraj) is a lot cleaner.

IAlibay, line 198: If we're not picking up the resid from anywhere, I believe that doesn't need to be unique - so it might make sense to set the resid to of the virtual site residue to the same value of its parent residue?

openff/interchange/interop/openmm/_topology.py
mattwthompson: #1076

openff/interchange/interop/openmm/_topology.py
mattwthompson: Sure - with a21f2ceeceab08e7587c4ef0ea0be8e4656005cd,

```python
In [1]: from openff.toolkit import *

In [2]: mol = Molecule.from_smiles("O")

In [3]: for atom in mol.atoms:
   ...:     atom.metadata["residue_name"] = "SOL"

In [4]: In [30]: x = (
   ...:     ...:     ForceField("opc.offxml")
   ...:     ...:     .create_interchange(
   ...:     ...:         Topology.from_molecules(
   ...:     ...:             [
   ...:     ...:                 mol,#Molecule.from_smiles("O"),
   ...:     ...:                 mol,#Molecule.from_smiles("O"),
   ...:                          mol,
   ...:     ...:             ],
   ...:     ...:         ),
   ...:     ...:     )
   ...:     ...:     .to_openmm_topology()
   ...:     ...: )
/Users/mattthompson/micromamba/envs/new-models/lib/python3.11/site-packages/smirnoff99frosst/smirnoff99frosst.py:11: UserWarning: Module openff was already imported from None, but /Users/mattthompson/software/openff-interchange is being added to sys.path
  from pkg_resources import resource_filename
/Users/mattthompson/micromamba/envs/new-models/lib/python3.11/site-packages/mdtraj/formats/__init__.py:13: DeprecationWarning: 'xdrlib' is deprecated and slated for removal in Python 3.13
  from mdtraj.formats.trr import TRRTrajectoryFile


In [5]: for residue in x.residues():
   ...:     print(residue.index, [*residue.atoms()])
0 [<Atom 0 (O1x) of chain 0 residue 0 (SOL)>, <Atom 1 (H1x) of chain 0 residue 0 (SOL)>, <Atom 2 (H2x) of chain 0 residue 0 (SOL)>]
1 [<Atom 3 (O1x) of chain 1 residue 1 (SOL)>, <Atom 4 (H1x) of chain 1 residue 1 (SOL)>, <Atom 5 (H2x) of chain 1 residue 1 (SOL)>]
2 [<Atom 6 (O1x) of chain 2 residue 2 (SOL)>, <Atom 7 (H1x) of chain 2 residue 2 (SOL)>, <Atom 8 (H2x) of chain 2 residue 2 (SOL)>]
3 [<Atom 9 (EP) of chain 3 residue 3 (SOL)>]
4 [<Atom 10 (EP) of chain 3 residue 4 (SOL)>]
5 [<Atom 11 (EP) of chain 3 residue 5 (SOL)>]

Comment (IAlibay) [reply openff/interchange/interop/openmm/_topology.py]: Is that for my other comment 😅?

File: openff/interchange/interop/openmm/_topology.py [hunk 5]
```diff
@@ -173,6 +174,30 @@ def to_openmm_topology(
...
+
+            for this_virtual_site in virtual_sites_in_this_molecule:
+                virtual_site_name = this_virtual_site.name
+
+                openmm_topology.addAtom(
+                    virtual_site_name,
+                    virtual_site_element,
+                    this_virtual_site_residue,
+                )
+
```
Yoshanuikabundi, line 200: I like Irfan's suggestions of having residue names and numbers match - that way the only thing that varies with the `collate` argument is the atom ordering, not the metadata of the virtual sites. To that end, I'd suggest an implementation along the lines of the following, which just uses the fact that atom indices for parent atoms are kept constant to index into the chain and residue metadata that we just generated:

```suggestion
    if has_virtual_sites and not collate:
        chain = None
        residue = None
        for virtual_site_key in interchange["VirtualSites"].key_map:
            assert isinstance(virtual_site_key, VirtualSiteKey)
            parent_atom_index = virtual_site_key.orientation_atom_indices[0]
            parent_atom = omm_atoms[parent_atom_index]

            if chain is None or chain.id != parent_atom.residue.chain.id:
                chain = openmm_topology.addChain(id=parent_atom.residue.chain.id)

            if residue is None or not (
                residue.name == parent_atom.residue.name
                and residue.id == parent_atom.residue.id
                and residue.insertionCode == parent_atom.residue.insertionCode
                and residue.chain.id == parent_atom.residue.chain.id
            ):
                residue = openmm_topology.addResidue(
                    name=parent_atom.residue.name,
                    chain=chain,
                    id=parent_atom.residue.id,
                    insertionCode=parent_atom.residue.id,
                )

            virtual_site_name = virtual_site_key.name

            openmm_topology.addAtom(
                name=virtual_site_name,
                element=virtual_site_element,
                residue=residue,
            )
```

docs/using/output.md
```diff
@@ -69,6 +69,8 @@ openmm_positions: openmm.unit.Quantity = interchange.positions.to_openmm()
 openmm_box: openmm.unit.Quantity = interchange.box.to_openmm()
 ```
 
+If virtual sites are present in the system, they will all be placed at the end of the system and the topology. In this case, all virtual sites are placed in a new "VS" residue at the end of the topology. Optionally, virtual sites can collated within molecules in the topology, associated with the last residue in each molecule. In this case, all virtual sites are still placed at the end of the system. To do this, use `collate=True` as an argument to `Interchange.to_openmm_topology`. For discussion, see [issue #1049](https://github.com/openforcefield/openff-interchange/issues/1049).
```
Yoshanuikabundi, line 72:
```suggestion
If virtual sites are present in the system, they will all be placed at the end of the system and the topology. Optionally, virtual sites can collated within molecules in the topology, associated with the last residue in each molecule. In this case, all virtual sites are still placed at the end of the system. To do this, use `collate=True` as an argument to `Interchange.to_openmm_topology`. For discussion, see [issue #1049](https://github.com/openforcefield/openff-interchange/issues/1049).
```

Update to reflect changes after Irfan's code review.

Yoshanuikabundi (commented):
Nice.

- Having `to_openmm_topology` match `to_openmm_system` seems pretty essential
- Maintaining the guarantee that we'll keep atom indices constant no matter what you do is extremely valuable
- This is a great solution

I have one blocking bug fix in `to_pdb`, a suggestion of how to accomplish Irfan's suggestion of preserving parent atom metadata (which I think is very valuable for analysis), and a few nits.

openff/interchange/interop/openmm/_topology.py
mattwthompson: Ah, yes, IDs and names are different nouns, after all ...

openff/interchange/components/interchange.py
```diff
@@ -680,6 +697,7 @@ def to_pdb(self, file_path: Path | str, include_virtual_sites: bool = False):
             )
 
             openmm_topology = self.to_openmm_topology(
+                collate=True,
```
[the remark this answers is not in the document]
mattwthompson: I'm aggressively indifferent on this (or at least I can't remember why I chose this - and I think the question of whether or not to include them, by default, at all is substantially more important) so let's go with this

mattwthompson, line 700:
```suggestion
                collate=False,
```

PR #1065, opened 2024-09-30
Error if HMR results in negative particle mass

### LINKED ISSUE #1061: `_apply_hmr` can silently lead to negative values
**Description**

Appying HMR has no validation / checks in place to avoid zero / negative mass setting.

Generally I would say this is rare, but I managed to accidentally create this with a 4 amu style test on a methane.

I haven't gone all the way through to a simulation, but whilst I suspect negative mass will fail, zero mass is a special case for openmm which will just freeze the atom (you'd have to be really lucky to set things to zero though).

Possibly just checking for the negative case and throwing an error / warning could be useful?

**Reproduction**

```python
from openff.interchange.components._packmol import solvate_topology_nonwater
import pytest


ligand = Molecule.from_smiles('C')
ligand.generate_conformers()
ligand.assign_partial_charges(partial_charge_method='am1bcc')

inter = Interchange.from_smirnoff(topology=ligand.to_topology(), force_field=ff)

system = inter.to_openmm_system(hydrogen_mass=4.0*1.008)
print(system.getParticleMass(0))
```

**Output**

```
-0.08543200000000084 Da
```

**Software versions**

Interchange v0.3.29

mattwthompson:
### Description

Resolves #1061

### Checklist

- [x] Add tests
- [x] Lint
- [x] Update docstrings

### DIFF
--- a/openff/interchange/interop/openmm/__init__.py
+++ b/openff/interchange/interop/openmm/__init__.py
@@ -219,7 +219,7 @@ def _apply_hmr(
             # when virtual sites + HMR are supported
             mass_to_transfer = _hydrogen_mass - system.getParticleMass(hydrogen_index)
 
-            if mass_to_transfer > heavy_mass:
+            if mass_to_transfer >= heavy_mass:
                 raise NegativeMassError(
                     f"Particle with index {heavy_index} would have a negative mass after hydrogen "
                     "mass repartitioning. Consider transferring a smaller mass than "
--- a/openff/interchange/interop/openmm/__init__.py
+++ b/openff/interchange/interop/openmm/__init__.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING, TextIO
 
 from openff.utilities.utilities import has_package, requires_package
 
+from openff.interchange._annotations import PositiveFloat
 from openff.interchange.exceptions import (
     NegativeMassError,
     PluginCompatibilityError,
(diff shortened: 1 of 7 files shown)


openff/interchange/interop/openmm/__init__.py
```diff
@@ -36,7 +38,7 @@ def to_openmm_system(
     combine_nonbonded_forces: bool = False,
     add_constrained_forces: bool = False,
     ewald_tolerance: float = 1e-4,
-    hydrogen_mass: float = 1.007947,
+    hydrogen_mass: PositiveFloat = 1.007947,
```
IAlibay, line 41: I know we have the annotation and all, but what if I passed a negative value here? Is there anything at runtime that would actually stop me from creating negatively massed hydrogens with super heavy atoms?

mattwthompson: Yep - slightly buried in the logic so it doesn't stand out in the diff: https://github.com/openforcefield/openff-interchange/pull/1065/files#diff-80d8de9a0e22631e9558b73e53bb5e90a13dd20d0e8cb9fa26f25c95622d4745R222-R228

ntBre (approved):
Overall LGTM. One minor point, is it worth mentioning in the release notes (and maybe the type annotation) the behavior around zero? From the actual [condition](https://github.com/openforcefield/openff-interchange/pull/1065/files#diff-80d8de9a0e22631e9558b73e53bb5e90a13dd20d0e8cb9fa26f25c95622d4745R222-R228), I think exactly zero is okay, right? This feels really pedantic but that would make `PositiveFloat` a `NonNegativeFloat` or something and the `Annotated` field `Ge(0)`.

mattwthompson:
Good catch - I'll make it easy on myself and forbid zero mass. If somebody wants it ... well, they can make the case for it

Issue #1062, opened 2024-09-25
Packmol interface: can't pack mnsol benchmark inputs with assumed evaluator defaults


IAlibay:
**Description**

I'm trying to test if we can pack mnsol (from the 2.0 sage benchmark) with Interchange.
My rough understanding (from the paper and some limited looking at the evaluator v3.11 code) is that it was possible to pack all the systems with 2k molecules, a cubic box and a 0.95 * unit.gram/unit.milliliter target density.

However, I can't seem to be able to make this work using Interchange.

The main questions here are:

1. Is this a correct assumption for how the sage MNSOL benchmarks were run, or is there more to it that I'm missing?
2. Should we be able to reproduce this with Interchange?
3. If so, how?

**Reproduction**

Here's a good example case:

```python
from openff.units import unit
from openff.units.openmm import from_openmm
from openmm import NonbondedForce
from openff.toolkit import Molecule, Topology, ForceField
from openff.interchange.components._packmol import pack_box, UNIT_CUBE
from openff.interchange import Interchange
import pytest

solvent = Molecule.from_smiles('CCCCO')
solvent.generate_conformers()
solvent.assign_partial_charges(partial_charge_method='am1bcc')

ligand = Molecule.from_smiles('CCCCCCCC')
ligand.generate_conformers()
ligand.assign_partial_charges(partial_charge_method='am1bcc')

off_top = Topology.from_molecules(ligand)
solvated_off_top = pack_box(
    molecules=[solvent],
    number_of_copies=[2000],
    solute=off_top,
    tolerance=2.0*unit.angstrom,
    mass_density=0.95 * unit.gram/unit.milliliter,
    box_shape=UNIT_CUBE,
    center_solute=True,
    working_directory='.',
    retain_working_files=False,
)
```

**Output**

```
---------------------------------------------------------------------------
CalledProcessError                        Traceback (most recent call last)
File ~/software/mambaforge/install/envs/openfe/lib/python3.12/site-packages/openff/interchange/components/_packmol.py:718, in pack_box(molecules, number_of_copies, solute, tolerance, box_vectors, mass_density, box_shape, center_solute, working_directory, retain_working_files)
    717 try:
--> 718     result = subprocess.check_output(
    719         packmol_path,
    720         stdin=file_handle,
    721         stderr=subprocess.STDOUT,
    722     ).decode("utf-8")
    723 except subprocess.CalledProcessError as error:

[...]

PR #1068, opened 2024-10-02
Decrease default density in Packmol wrappers

mattwthompson:
### Checklist

- [x] Set default density to 0.9 for water-based solvents
- [x] Make density a required argument for non-water solvents
- [ ] Buffer box size by 10% (across the board)
- [x] Update argument name for consistency
- [x] Communicate changes to stakeholders
- [x] Add tests
- [x] Lint
- [x] Update release history


IAlibay:
Is 0.6 actually tested on a simulation (i.e. do we have a sense for how well it equilibrates out?).

mattwthompson:
Not really - that'd be a good integration test to add. Maybe 0.7-0.8 works better, my intuition is that going a little lower than 0.8 is better but the tradeoffs could deteriorate there

mattwthompson:
Current intention is to get this into 0.4.0, since it's strictly a significant set of API breaks, and kick more rigorous testing down the road.

davidlmobley:
As a quick note on this:
> Is 0.6 actually tested on a simulation (i.e. do we have a sense for how well it equilibrates out?).

My previous finding was that, for many solvents, densities equilibrate relatively quickly (pathological mixtures like octanol in water of course being somewhat of an exception, and even that isn't necessarily too bad on the many nanoseconds timescale). I once did a deletion of a whole protein from a box of water and re-equilibrated the water and it wasn't that painful. I think long range order (like if your box was initially packed as a crystal and you need it to thaw) is a much worse problem than "dramatically wrong density". 

I haven't explored that exhaustively, and as noted there will certainly be pathological solvent mixtures, or things which gel or form glasses or... but for many liquids I think this should often be fine. As a user I'm certainly comfortable with someone giving me whatever density they need to get it to pack and then equilibrating from there.

IAlibay:
> densities equilibrate relatively quickly

I generally agree with this, however there are two considerations here:

1. The main reason for doing some kind of "real simulation" test here is because we know that Packmol (at least in the way it's being interfaced here) can often pack a box but in such a way that it will just randomly explode. My suspicion is that this is the kind of bug that will happen when you push things to the extreme. Since this density is reasonably low, it would be good to at least minimally check that, for the "solvents that OpenFF wants to use" things seem reasonably stable.
2. On the note of "solvents that OpenFF wants to use", the original issue #1062 is related to trying to get systems to run for OpenFF's SFE purposes. Who's responsibility is it to validate these solvent mixtures for standard (non-alchemical) MD? Interchange or Pontibus or something else?

docs/releasehistory.md
```diff
@@ -22,6 +22,8 @@ Please note that all releases prior to a version 1.0.0 are considered pre-releas
 * Several classes and methods which were deprecated in the 0.3 line of releases are now removed.
 * Previously-deprecated examples are removed.
 * `ProperTorsionKey` no longer accepts an empty tuple as atom indices.
+* Default densities for Packmol wrapper functions have been lowered to 0.9 g/cc and 0.6 g/cc for water and non-water solvent, respectively.
+* Argument `mass_density` to some packing functions has been renamed to `target_density` for consistency and to better reflect its usage.
```
Yoshanuikabundi, line 26: Yeah I agree, `target_density` is a better name. Who thinks "density" needs to be disambiguated to "mass density", some weirdo I reckon (reader, it was me)

openff/interchange/components/_packmol.py
```diff
@@ -801,7 +801,7 @@ def solvate_topology(
     nacl_conc: Quantity = 0.1 * unit.mole / unit.liter,
     padding: Quantity = 1.2 * unit.nanometer,
     box_shape: NDArray = RHOMBIC_DODECAHEDRON,
-    target_density: Quantity = 1.0 * unit.gram / unit.milliliter,
+    target_density: Quantity = 0.9 * unit.gram / unit.milliliter,
```
Yoshanuikabundi, line 804: :+1:

openff/interchange/components/_packmol.py
```diff
@@ -889,7 +889,7 @@ def solvate_topology_nonwater(
     solvent: Molecule,
     padding: Quantity = 1.2 * unit.nanometer,
     box_shape: NDArray = RHOMBIC_DODECAHEDRON,
-    target_density: Quantity = 1.0 * unit.gram / unit.milliliter,
+    target_density: Quantity = 0.6 * unit.gram / unit.milliliter,
```
Yoshanuikabundi, line 892: Unlike `solvate_topology`, `solvate_topology_nonwater` could have a lot of different solvents with a lot of different densities. I agree 0.6 g/mL is a better default for organic solvents than 1.0 g/mL, but I think making this a required argument is a better option. 

I designed the API of the `solvate_topology` function to allow the user not to think about the box, because a very reasonable default does exist, and overwriting any single part of that default makes sense for some special case (and I got sick of writing the same calculation in every prep workflow).

With a generalized solvent, that default doesn't really exist - the padding distance might need to be different if finite size effects are more or less important, and the density is straightforwardly unknown. In the case of padding I think the "images of the solute don't see each other's VDWs" heuristic is still at least well grounded, but the density is really a mystery. The best case is that the system blows up and a user thinks "why is that the default". The worst case is we get the density only slightly wrong, the box equilibrates, the padding distance is just made meaningless and some finite size effect is discovered after publication.

So I'd suggest

```suggestion
    target_density: Quantity,
```

Yoshanuikabundi (commented):
Love the rename and agree that 0.9 g/mL is probably better than 1.0. I have an argument below that `target_density` should just be a required argument for `solvate_topology_nonwater`.

openff/interchange/components/_packmol.py
mattwthompson: You make a good case

mattwthompson:
Thanks for the review @Yoshanuikabundi!

PR #1066, opened 2024-09-30
Fix default `.charges`

mattwthompson:
### Description

Resolves #1052 ... mostly

### Checklist

- [x] Add tests
- [x] Lint
- [ ] Update docstrings

PR #1069, opened 2024-10-02
Use update instead of assignment to set _charges

ntBre:
### Description

Minor update to #1066 to include `_charges` in the model definition. I still don't know why this is required, but it seems to work in my local tests. I just realized that it might be necessary to do `self._charges.clear()` before `self._charges.update()` to avoid mixing charges. I read the `if` condition as `len == 0 and ...` instead of `or`, which could possibly allow `update` to be called on a dict already containing some charges.

### Checklist

- [x] Add tests - should be covered by existing tests
- [ ] Lint
- [x] Update docstrings - n/a

### DIFF
--- a/openff/interchange/common/_nonbonded.py
+++ b/openff/interchange/common/_nonbonded.py
@@ -1,6 +1,6 @@
 import abc
 from collections.abc import Iterable
-from typing import Literal
+from typing import Any, Literal
 
 from openff.toolkit import Quantity, unit
 from pydantic import Field, PrivateAttr, computed_field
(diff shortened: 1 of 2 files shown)


openff/interchange/common/_nonbonded.py
```diff
@@ -112,7 +111,7 @@ def charges(
     ) -> dict[TopologyKey | LibraryChargeTopologyKey | VirtualSiteKey, _ElementaryChargeQuantity]:
         """Get the total partial charge on each atom, including virtual sites."""
         if len(self._charges) == 0 or self._charges_cached is False:  # type: ignore[has-type]
-            self._charges = self._get_charges(include_virtual_sites=False)
+            self._charges.update(self._get_charges(include_virtual_sites=False))
```
mattwthompson, line 114: This ... makes a fair amount of sense to me, actually. I think in non-Pydantic world it's best practice to update a dict instead of re-assign it, but I'm also pretty sure something in Pydantic's magic machinery means just passing the return value of another function to an instance attribute gummed something up something.

mattwthompson (approved):
Thanks!

mattwthompson:
>  I just realized that it might be necessary to do self._charges.clear() before self._charges.update() to avoid mixing charges.

For posterity, this was a Cassandra level of prescient. For some reason, no tests that existed before this PR were failing, but some tests added elsewhere (#1070) did break without clearing the cache at that point.

Issue #1059, opened 2024-09-20
`from_smirnoff` _without_ `charge_from_molecules` silently assigns all isormophic molecules the same partial charges


IAlibay:
**Description**

User case:

I construct an Interchange object from a Topology with two or more ismorphic Molecules with different partial charges.
In this case the first Molecule has am1bcc charges whilst all the others have gasteiger charges, which differ significantly.

Outcome: 

All the Molecules silently get assigned the partial charges from the first Molecule.

**Reproduction**

```python
from openff.units.openmm import from_openmm
from openmm import NonbondedForce
from openff.toolkit import Molecule, Topology, ForceField
from openff.interchange.components._packmol import solvate_topology_nonwater
from openff.interchange import Interchange
import pytest

solvent = Molecule.from_smiles('c1ccccc1')
solvent.assign_partial_charges(partial_charge_method='gasteiger')
print("solvent charges: ", solvent.partial_charges)

ligand = Molecule.from_smiles('c1ccccc1')
ligand.generate_conformers()
ligand.assign_partial_charges(partial_charge_method='am1bcc')
print("ligand charges: ", ligand.partial_charges)

print("molecules are isomorphic: ", ligand.is_isomorphic_with(solvent))

off_top = Topology.from_molecules(ligand)
solvated_off_top = solvate_topology_nonwater(topology=off_top, solvent=solvent)

ff = ForceField('openff-2.2.0.offxml', 'opc.offxml')
inter = Interchange.from_smirnoff(topology=solvated_off_top, force_field=ff)

for _, c in inter.collections["Electrostatics"].charges.items():
    assert abs(c.m) == pytest.approx(0.13)
    
# Double check this isn't just the collection, and it goes all the way to an openmm System
system = inter.to_openmm_system()
nonbond = [f for f in system.getForces() if isinstance(f, NonbondedForce)][0]

for idx in range(system.getNumParticles()):
    c, _, _ = nonbond.getParticleParameters(idx)
    assert abs(from_openmm(c).m) == 0.13

print(all(solvated_off_top.molecule(0).partial_charges == ligand.partial_charges))
print(all(solvated_off_top.molecule(1).partial_charges == ligand.partial_charges))
print(all(solvated_off_top.molecule(1).partial_charges == solvent.partial_charges))
```

**Output**

```
[...]

Issue #1058, opened 2024-09-20
`charge_from_molecules` silently ignores isomorphic molecules in list, only using the first entry


IAlibay:
**Description**

Passing two isomorphic Molecules with different partial charges to `from_smirnoff`'s `charge_from_molecules` flag will lead to partial charges only being assigned from the first Molecule.

**Reproduction**

```python
from openff.toolkit import Molecule, Topology, ForceField
from openff.interchange.components._packmol import solvate_topology_nonwater
from openff.interchange import Interchange
import pytest

solvent = Molecule.from_smiles('c1ccccc1')
solvent.assign_partial_charges(partial_charge_method='gasteiger')
print("solvent charges: ", solvent.partial_charges)

ligand = Molecule.from_smiles('c1ccccc1')
ligand.generate_conformers()
ligand.assign_partial_charges(partial_charge_method='am1bcc')
print("ligand charges: ", ligand.partial_charges)

print("molecules are isomorphic: ", ligand.is_isomorphic_with(solvent))

off_top = Topology.from_molecules(ligand)
solvated_off_top = solvate_topology_nonwater(topology=off_top, solvent=solvent)

ff = ForceField('openff-2.2.0.offxml', 'opc.offxml')
inter = Interchange.from_smirnoff(topology=solvated_off_top, force_field=ff, charge_from_molecules=[ligand, solvent])

for _, c in inter.collections["Electrostatics"].charges.items():
    assert abs(c.m) == pytest.approx(0.13)
    
inter = Interchange.from_smirnoff(topology=solvated_off_top, force_field=ff, charge_from_molecules=[solvent, ligand])

for _, c in inter.collections["Electrostatics"].charges.items():
    assert abs(c.m) == pytest.approx(0.06226857)
```

**Output**

```
solvent charges:  [-0.062268570782092456 -0.062268570782092456 -0.062268570782092456 -0.062268570782092456 -0.062268570782092456 -0.062268570782092456 0.062268570782092456 0.062268570782092456 0.062268570782092456 0.062268570782092456 0.062268570782092456 0.062268570782092456] elementary_charge
ligand charges:  [-0.13 -0.13 -0.13 -0.13 -0.13 -0.13 0.13 0.13 0.13 0.13 0.13 0.13] elementary_charge
molecules are isomorphic:  True
```

**Software versions**

Interchange v0.3.29

mattwthompson:
That doesn't seem good!

SMIRNOFF doesn't help in this case:

[...]

Issue #1050, opened 2024-09-16
Partial charge assignment using `charge_from_molecules` on virtual sites / charge increments can be unexpected.


IAlibay:
**Description**

Related to #1048 

When building a system with OPC water, if a user passed along a pre-charged water, they might end up with an unexpected set of partial charges in the resultant 4 point molecule due to charge increments.

The behaviour isn't incorrect according to the spec, but you then also have to expect users to a) know the charge assignment order in the spec, b) remember that there are charge increments affecting the molecule they passed charges for, c) do the charge increment math ahead of time to make sure they passed along the right charges.

I'm not 100% sure, but I think the ask would be something along the lines of:

1. Some kind of tooling to inspect the partial charges you are likely to get after all the increments, etc.. are applied for each molecule (rather than having to manually inspect the OpenMM system to see what happened). Is there a better way to assign partial charges for virtual site containing molecules that I'm missing?
2. Some kind of warning that further processing of the partial charges will happen.
3. Documentation that a) yes `charge_from_molecules` will ignore library charges, but b) charge increments will be applied (this is not clear from the existing docs).

**Reproduction**

```python
from openff.toolkit import Molecule, Topology, ForceField
from openff.interchange.components._packmol import solvate_topology
from openff.interchange import Interchange
from openmm import NonbondedForce

water = Molecule.from_smiles('O')
water.assign_partial_charges(partial_charge_method='gasteiger')
print(water.partial_charges)

ligand = Molecule.from_smiles('CCCCC')
ligand.generate_conformers()

off_top = Topology.from_molecules(ligand)
solvated_off_top = solvate_topology(topology=off_top)

ff = ForceField('openff-2.2.0.offxml', 'opc.offxml')
inter = Interchange.from_smirnoff(topology=solvated_off_top, force_field=ff, charge_from_molecules=[water])
omm_system = inter.to_openmm_system()

nonbond = [f for f in system.getForces() if isinstance(f, NonbondedForce)][0]

# Get the parameters for the first water oxygen
nonbond.getParticleParameters(17)
```

**Output**

partial charges
```
[-0.4115095219374226 0.2057547609687113 0.2057547609687113] elementary_charge
```
particle parameters for oxygen (in OPC this is a zero charged atom since all the charge is on the virtual site)
```
[Quantity(value=-0.4115095219374226, unit=elementary charge),
 Quantity(value=0.31506999999999996, unit=nanometer),
[...]

PR #1053, opened 2024-09-19
Log charge assignment

mattwthompson:
### Description

Resolves #1048

### Checklist

- [x] Log, for most cases, which atom (with some topology index) got what charge assigned by which method
  - [x] Library charges
  - [x] Preset charges
  - [x] `ChargeIncrementHandler`
  - [x] AM1-BCC
  - [x] Virtual sites
- [x] Add tests
  - [x] Sage with a ligand in vacuum
  - [x] Sage with a ligand in water/ions
    * Water gets library charges
    * Ions get library charges
    * Non-water solvent gets AM1-BCC
    * Ligands get AM1-BCC
  - [x] Sage with a ligand in non-water solvent
    * Non-water solvent gets AM1-BCC
    * Ligands get AM1-BCC
  - [x] Sage with a ligand in mixed solvent
    * Water gets library charges
    * Ions get library charges
    * Non-water solvent gets AM1-BCC
    * Ligands get AM1-BCC
  - [x] ff14SB with Sage
    * Protein gets library charges
    * Water gets library charges
    * Ions get library charges
    * Non-water solvent gets AM1-BCC
    * Ligands get AM1-BCC
  - [x] ff14SB with Sage and preset charges on Protein A
    * Protein A gets preset charges
    * Other proteins get library charges
    * Water gets library charges
    * Ions get library charges
    * Non-water solvent gets AM1-BCC
    * Ligands get AM1-BCC
  - [x] Sage with ligand and OPC water
    * Water gets library charges
    * Water virtual sites get ??
    * Ions get library charges
    * Non-water solvent gets AM1-BCC
    * Ligands get AM1-BCC
  - [x] Sage with preset charges on ligand A
    * Water gets library charges
    * Ions get library charges
    * Non-water solvent gets AM1-BCC
    * Ligand A gets preset charges
    * Other ligands get AM1-BCC
  - [x] Sage with preset charges on water
    * Water gets preset charges
    * Ions get library charges
    * Non-water solvent gets AM1-BCC
    * Ligands get AM1-BCC
  - [x] Sage with (ligand) virtual site parameters
    * Water gets preset charges
    * Ions get library charges
    * Non-water solvent gets AM1-BCC
    * Ligand heavy atoms get AM1-BCC and charge increments
    * Virtual sites get charge increments
  - [x] AM1-with-custom-BCCs Sage with ligand and ions water
    * Water gets library charges
    * Ions get library charges
    * Ligand gets charge increments
[...]

PR #1070, opened 2024-10-03
Handle some edge cases with preset charges

### LINKED ISSUE #1057: Passing a Molecule without charges to `from_smirnoff`'s `charge_from_molecules` fails ungracefully
Definitely not a big issue, mostly opening it because @mattwthompson told me to raise these issues 😅 

**Description**

Passing an uncharged Molecule through `charge_from_molecule` leads to a TypeError when trying to iterate over None.
The traceback is clear, but the error message could be better for users.

**Reproduction**

```python

water = Molecule.from_smiles('O')

ligand = Molecule.from_smiles('CCCCC')
ligand.generate_conformers()

off_top = Topology.from_molecules(ligand)
solvated_off_top = solvate_topology(topology=off_top)

inter = Interchange.from_smirnoff(topology=solvated_off_top, force_field=ff, charge_from_molecules=[water])
```

**Output**

TypeError: 'NoneType' object is not iterable

**Software versions**

Interchange v0.3.29

mattwthompson:
### Description

This PR better handles user inputs around the `charge_from_molecules` argument from the toolkit. Resolves #1057 #1058 #1059

### Checklist

- [x] Error if `charge_from_molecules` contains any molecule without partial charges
- [x] Error if `charge_from_molecules` contains any duplicate molecules (defined by isomorphism, approximated by SMILES without hydrogens)
- [x] Test behavior discussed in #1050
- [x] Add tests
- [x] Lint
- [x] Update docstrings
  - [x] Update `from_smirnoff` docstring
  - [x] Add user guide
    - [x] Describe charge assignment hierarchy
    - [x] Describe limitations with preset charges
    - [ ] ~~Add FAQ entry~~ I guess we don't have an FAQ here, maybe should check `openff-docs`


openff/interchange/smirnoff/_create.py
```diff
@@ -93,17 +97,61 @@ def validate_topology(value):
...
 ) -> Interchange:
+    molecules_with_preset_charges = _preprocess_preset_charges(molecules_with_preset_charges)
+
     _check_supported_handlers(force_field)
 
+    if molecules_with_preset_charges is not None and "VirtualSites" in force_field.registered_parameter_handlers:
+        warnings.warn(
+            "Preset charges were provided (via `charge_from_molecules`) alongside a force field that includes "
+            "virtual site parameters. Note that virtual sites will be applied charges from the force field and "
+            "cannot be given preset charges.",
```
Yoshanuikabundi, line 151:
```suggestion
            "Preset charges were provided (via `charge_from_molecules`) alongside a force field that includes "
            "virtual site parameters. Note that virtual sites will be applied charges from the force field and "
            "cannot be given preset charges. Virtual sites may also affect the charges of their orientation "
            "atoms, even if those atoms are given preset charges.",
```

Do you think adding something like this to the warning will be useful for people? It seems like the sort of thing that would surprise me but maybe it's just noise.

openff/interchange/smirnoff/_create.py
```diff
@@ -93,17 +97,61 @@ def validate_topology(value):
...
+
+    1. The input is a list of Molecules
+    2. Each molecule has assign partial charges
+    3. Ensure no molecules are isomorphic with another in the list
+
+    """
+    if molecules_with_preset_charges is None:
+        return None
+
+    molecule_set = {molecule.to_smiles() for molecule in molecules_with_preset_charges}
```
Yoshanuikabundi, line 118:
```suggestion
    molecule_set = {molecule for molecule in molecules_with_preset_charges}
```

`Molecule.__eq__()` exists, so you should be able to make a set out of molecules directly. It definitely seems to work for me. Is matching SMILES faster for large topologies or something?

openff/interchange/_tests/unit_tests/smirnoff/test_create.py
```diff
@@ -231,6 +234,98 @@ def test_charges_from_molecule_reordered(
 
         assert numpy.allclose(expected_charges, found_charges)
 
+    def test_error_when_any_missing_partial_charges(self, sage):
+        topology = Topology.from_molecules(
+            [
+                Molecule.from_smiles("C"),
+                Molecule.from_smiles("CCO"),
+            ],
+        )
```
Yoshanuikabundi, line 243: Should this topology include a molecule *with* partial charges, to ensure that it's testing if there are *any* missing charges?

Yoshanuikabundi (commented):
These are great improvements I think. I think the documentation changes from #1048 could be merged with this so that the charge hierarchy is only documented once, which might improve findability - I think if there is similar documentation in two places, people sometimes read one place carefully, then later find the second place and think they're already familiar with it coz it looks the same.

My notes are mostly just adding more detail to what you've written, but this is great!

openff/interchange/smirnoff/_create.py
mattwthompson: I figure making SMILES is probably slower because of the short-circuits in `are_isomorphic`. I think relying on built-in equality would work fine here, but I also want to encode the intentionality of uniqueness being defined here in a way that's different than how the toolkit does it, which may change. (I'm not sure `.to_smiles()` is the best way to do this

I'm pretty neutral on this change, but your suggestion is simpler. I'll go with it and note the context in a comment

openff/interchange/smirnoff/_create.py
mattwthompson: I think this is a great example of something that seems repetitive / doesn't need to be made explicit to me (since I'm so steeped in the details) but is worth repeating / making explicit to a user

openff/interchange/components/interchange.py
```diff
@@ -109,7 +109,9 @@ def from_smirnoff(
             positions are taken from the molecules in topology, if present on all molecules.
         charge_from_molecules : `List[openff.toolkit.molecule.Molecule]`, optional
             If specified, partial charges will be taken from the given molecules
-            instead of being determined by the force field.
+            instead of being determined by the force field. All molecules in this list
+            must have partial charges assigned and must not be isomorphic with any other
+            molecules in the list.
```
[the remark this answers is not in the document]
mattwthompson: (I had to grab this change manually because ... GitHub detected this involved a deleted line, and for some reason that precludes multi-line suggestions from being applied. Boo!)

mattwthompson:
Aiming to sync this up and re-visit, probably to finish it up, after #1053 is merged in

mattwthompson:
Thank you @Yoshanuikabundi!

PR #1073, opened 2024-10-07
Fix regression tests

mattwthompson:
### Description

Regression tests haven't been running for a while, not sure why


mattwthompson:
There are other regression tests in that repo, but this just gets what is currently failing back to passing.

Issue #1054, opened 2024-09-19
`to_openmm_positions` includes virutal sites but `Interchange.positions.to_openmm()` does not


IAlibay:
**Expectation**

1. Positions obtained from an interchange object should include all the particles in the system, including virtual sites.
2. Various API points for getting positions should all lead to the same outcomes

**Description**

1. Calling `Interchange.positions.to_openmm()` yields positions **without** virtual sites.
2. Calling `to_openmm_positions` yields positions **with** virtual sites.

**Reproduction**

```python
from openff.toolkit import Molecule, Topology, ForceField
from openff.interchange.components._packmol import solvate_topology
from openff.interchange.interop.openmm import to_openmm_positions
from openff.interchange import Interchange

water = Molecule.from_smiles('O')
water.assign_partial_charges(partial_charge_method='gasteiger')

ligand = Molecule.from_smiles('CCCCC')
ligand.generate_conformers()

off_top = Topology.from_molecules(ligand)
solvated_off_top = solvate_topology(topology=off_top)

ff = ForceField('openff-2.2.0.offxml', 'opc.offxml')
inter = Interchange.from_smirnoff(topology=solvated_off_top, force_field=ff)
print("Number of particles with .positions: ", len(inter.positions.to_openmm()))
print("Number of particles with to_openmm_positions: ", len(to_openmm_positions(inter)))
```

**Output**

```
Number of particles with .positions:  2035
Number of particles with to_openmm_positions:  2707
```

**Software versions**

Interchange v0.3.29

mattwthompson:
Hmm, yeah that's a bit of a surprises. Each of these things have some reason to behave the way they do but the inconsistency isn't great.

> Positions obtained from an interchange object should include all the particles in the system, including virtual sites.

Unfortunately, this is the heart of the matter. This is a consequence of the first-class view mirroring the toolkit's behavior of not knowing about virtual sites but other API points knowing about them. Right now, `Interchange.topology` can't store virtual sites[^1], and I'm not sure how I feel about `Interchange.positions` having a different shape than the number of atoms.[^2]

> Various API points for getting positions should all lead to the same outcomes

Lots of API points have arguments like `includle_virtual_sites` wired through them since there are some use cases that only want the atomic particles; I'm not sure how this would implemented on something that's a `pydantic.Field`, but maybe there's a way to wrap something up that handles both cases?

At very least we should document all of this better, but I'm not sure how best to do that since this touches so many parts of the codebase.
 
[...]

PR #1074, opened 2024-10-07
Add high-level positions getter, exclude virtual sites by default

mattwthompson:
### Description

While working on a short-term solution to #1054, I realized there's no clean high-level getter in the public API for accessing positions with virtual sites. This adds that, via refactoring `to_openmm_positions`, and documents some of the quirks in the existing design around positions and other topology objects.

### Checklist

- [x] Add `Interchange.get_positions` with (optionally, off by default) virtual sites
- [x] Refactor `to_openmm_positions` to use the non-OpenMM positions getter
- [x] Document quirks
- [ ] Add tests
- [x] Lint
- [ ] Update docstrings


ntBre (approved):
LGTM! The code looks good, and I think this makes a lot of sense for the API too.

mattwthompson:
Thanks!

PR #1079, opened 2024-10-17
Update Sphinx to 6.1.2

Yoshanuikabundi:
### Description

This PR updates Sphinx to a more recent version, as well as fixing a few rendering issues in the API docs. The main beneficial change here is that the API docs now have a right hand margin navigation menu, so users can quickly navigate to a field or function. Previously, this worked for modules but not classes.

api_docs_side_nav

I investigated updating further, but 6.2+ made changes to how privacy detection works that breaks our lazy loading subpackages. Curse Python for making modules and packages different!

I also investigated using a pre-commit hook to line wrap docstrings. I tried two formatters, [docformatter](https://github.com/PyCQA/docformatter) and [pydocstringformatter](https://github.com/DanielNoord/pydocstringformatter). Neither seemed to be able to reliably get docstrings to wrap to the correct width, and both introduced other changes I didn't like, so I haven't included them.

### DIFF
--- a/devtools/conda-envs/docs_env.yaml
+++ b/devtools/conda-envs/docs_env.yaml
@@ -20,8 +20,8 @@ dependencies:
   - myst-parser
   - numpydoc
   - autodoc-pydantic =2
-  - sphinx ~=4.4
+  - sphinx ~=5.3
   - sphinxcontrib-mermaid
   - sphinx-notfound-page
   - pip:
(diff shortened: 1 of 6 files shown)


mattwthompson (approved):
LGTM, I briefly poked through a couple of pages in the rendered docs and everything looks good to me

> I also investigated using a pre-commit hook to line wrap docstrings.

Would be happy to include something along these lines if you find a tool you're happy with, also fine not doing that here if you didn't find something you like

PR #1081, opened 2024-10-21
Support importing (some) virtual sites in `Interchange.from_openmm`

### LINKED ISSUE #1063: Support of virtual sites for creating an Interchange `from_openmm`
**Description**
I'm trying to create an Interchange object from an OpenMM system that contains TIP4P water and got this Error: `UnsupportedImportError: A particle is a virtual site, which is not yet supported.` 
Are there any plans to support virtual sites in the `from_openmm` call or would you recommend a different procedure in a case like this?

mattwthompson:
### Description

Resolves #1063

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/_tests/unit_tests/interop/openmm/_import/test_virtual_sites.py
+++ b/openff/interchange/_tests/unit_tests/interop/openmm/_import/test_virtual_sites.py
@@ -1,0 +1,56 @@
+import pytest
+from openff.toolkit import Topology
+
+from openff.interchange import Interchange
+from openff.interchange.components._packmol import solvate_topology
+from openff.interchange.drivers.openmm import _get_openmm_energies, get_openmm_energies
+
+
+class TestTIP4PVirtualSites:
+    def test_dimer_energy_equals(self, tip4p, water_dimer):
+        out: Interchange = tip4p.create_interchange(water_dimer)
+
+        roundtripped = Interchange.from_openmm(
+            system=out.to_openmm_system(),
+            topology=out.to_openmm_topology(collate=False),
+            positions=out.get_positions(include_virtual_sites=True).to_openmm(),
+            box_vectors=out.box.to_openmm(),
+        )
+
+        assert get_openmm_energies(out) == _get_openmm_energies(roundtripped)
+
+    def test_minimize_solvated_ligand(self, sage_with_tip4p, ethanol, default_integrator):
+        topology = solvate_topology(ethanol.to_topology())
+
+        simulation = sage_with_tip4p.create_simulation(
+            topology,
+        ).to_openmm_simulation(
+            integrator=default_integrator,
+        )
+
+        roundtripped = Interchange.from_openmm(
+            system=simulation.system,
+            topology=simulation.topology,
+            positions=simulation.context.getState(getPositions=True).getPositions(),
+            box_vectors=simulation.system.getDefaultPeriodicBoxVectors(),
+        )
+
+        original_energy = get_openmm_energies(simulation)
+
+        # TODO: Much more validation could be done here, but if a simulation
+        #       can start and minimize at all, that should catch most problems
+        roundtripped.minimize()
+
+        assert get_openmm_energies(roundtripped) < original_energy
+
+    def test_error_index_mismatch(self, tip4p, water):
+        out: Interchange = tip4p.create_interchange(Topology.from_molecules([water, water]))
+
+        with pytest.raises(
+            ValueError,  # TODO: Make a different error
+            match="The number of particles in the system and the number of atoms in the topology do not match.",
+        ):
+            Interchange.from_openmm(
+                system=out.to_openmm_system(),
+                topology=out.to_openmm_topology(collate=True),
+            )
--- a/openff/interchange/_tests/conftest.py
+++ b/openff/interchange/_tests/conftest.py
@@ -13,19 +13,26 @@ from openff.interchange._tests import MoleculeWithConformer, _rng, get_test_file
 
 
 @pytest.fixture
-def sage():
+def sage() -> ForceField:
     return ForceField("openff-2.0.0.offxml")
 
 
 @pytest.fixture
-def sage_unconstrained():
+def sage_unconstrained() -> ForceField:
     return ForceField("openff_unconstrained-2.0.0.offxml")
 
 
 @pytest.fixture
-def sage_no_switch(sage):
+def sage_no_switch(sage) -> ForceField:
     sage["vdW"].switch_width = Quantity(0.0, "angstrom")
     return sage
+
+
+@pytest.fixture
+def sage_with_tip4p() -> ForceField:
+    # re-build off of existing fixtures if this gets implemented
+    # https://github.com/openforcefield/openff-toolkit/issues/1948
+    return ForceField("openff-2.0.0.offxml", "tip4p.offxml")
 
 
 @pytest.fixture
--- a/openff/interchange/_tests/unit_tests/interop/openmm/_import/test_compat.py
+++ b/openff/interchange/_tests/unit_tests/interop/openmm/_import/test_compat.py
@@ -12,9 +12,7 @@ from openff.interchange.warnings import MissingPositionsWarning
 
 class TestUnsupportedCases:
     @pytest.mark.filterwarnings("ignore:.*are you sure you don't want to pass positions")
-    def test_error_topology_mismatch(self, monkeypatch, sage_unconstrained, ethanol):
-        monkeypatch.setenv("INTERCHANGE_EXPERIMENTAL", "1")
-
+    def test_error_topology_mismatch(self, sage_unconstrained, ethanol):
         topology = ethanol.to_topology()
         topology.box_vectors = Quantity([4, 4, 4], "nanometer")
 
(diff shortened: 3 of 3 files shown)


openff/interchange/interop/openmm/_import/compat.py
```diff
@@ -4,6 +4,11 @@
 
 from openff.interchange.exceptions import UnsupportedImportError
 
+_UNSUPPORTED_VIRTUAL_SITE_TYPES: set[type] = {
+    openmm.TwoParticleAverageSite,
+    openmm.OutOfPlaneSite,
+}
+
```
Yoshanuikabundi, line 11: My instinct would be to flip this and explicitly list the supported virtual site types, so that if OpenMM adds a new virtual site type this won't need to be updated.

If there's a nice place/name for it, making this constant public would allow the supported virtual sites to be documented automatically directly from the code.

openff/interchange/components/toolkit.py
```diff
@@ -105,35 +108,98 @@ def _check_electrostatics_handlers(force_field: "ForceField") -> bool:
...
     graph = networkx.Graph()
 
+    virtual_sites: list[ImportedVirtualSiteKey] = list()
+
+    # map indices of OpenMM system with virtual site to topology indices, or virtual site keys,
+    # in associated OpenFF topology, since stripping out virtual sites offsets particle indices
+    openmm_openff_particle_map: dict[int, int | ImportedVirtualSiteKey] = dict()
+
     # TODO: This is nearly identical to Topology._openmm_topology_to_networkx.
     #  Should this method be replaced with a direct call to that?
```
Yoshanuikabundi, line 136: Is this comment still up-to-date? :upside_down_face:

openff/interchange/components/toolkit.py
```diff
@@ -105,35 +108,98 @@ def _check_electrostatics_handlers(force_field: "ForceField") -> bool:
...
+
+    # TODO: This iteration strategy scales horribly with system size - need to refactor -
+    #       since looking up topology atom indices is slow. It's probably repetitive to
+    #       look up the molecule index over and over again
+    for particle in virtual_sites:
+        molecule_index = topology.molecule_index(topology.atom(particle.orientation_atom_indices[0]).molecule)
+
+        topology._molecule_virtual_site_map[molecule_index].append(particle)
+
+    topology._particle_map = openmm_openff_particle_map
```
Yoshanuikabundi, line 200: Would it be reasonable to squish `topology._particle_map` and `topology._molecule_virtual_site_map` into atom metadata rather than monkey patching? I think the answer is probably "no", and I see this information is moved to the `Interchange`, and I've tested that `Interchange.topology` seems to be able to be swapped out for a true topology (with full molecular graphs etc).

Yoshanuikabundi (commented):
I tested this on a hand-built OpenMM Simulation featuring a single water molecule with made up parameters, and also on a single Tip4P molecule. Things mostly worked, and errors were mostly clear. I do have a few notes:

- `export INTERCHANGE_EXPERIMENTAL=1` seems to be required, but this does not seem to be documented and the error message doesn't make this clear. This was the only time I was confused by an error message.
- At the moment, the number of atoms in the imported topology must match the number of particles in the system. I think we should consider if topologies matching the number of *real atoms* in the system should be accepted as well. This would allow OpenFF `Topology` objects to work when importing systems with supported virtual sites, as they currently do for systems without virtual sites. This makes particular sense given that it appears that virtual sites are stripped out of the OpenFF topology that's available from `interchange.topology`.
- The exported OpenMM topology via `to_openmm_simulation()` had its virtual sites in a separate chain and residue - I think it'd be better if they were in the same chain and residue as the parent atom.
- This is unrelated to the PR, but I tried a hand-built system without bonds (single water molecule with constraints, but no bonds force), and Interchange complained:
```py
    136 # TODO: Does this run through the Interchange.box validator?
    137 interchange.box = _box_vectors
--> 139 if interchange.topology.n_bonds > len(interchange.collections["Bonds"].key_map):
    140     # There are probably missing (physics) bonds from rigid waters. The topological
    141     # bonds are probably processed correctly.
    142     _fill_in_rigid_water_bonds(interchange)
    144 return interchange

KeyError: 'Bonds'
```
Since this code currently attempts to account for bond parameters missing from the bond parameters, maybe we should account for this possibility?

A full review for such a big patch is gonna take me some time to work through, especially since I'm gonna need to get my head around how everything works, but those are my notes from some independent testing.

But yeah this is great! Things seem to be working pretty robustly once you come up with a system and topology that works - I was able to swap the topology of simple molecules out for a "real" openff topology after the Interchange was constructed and things kept working. Round trips work, energy minimization works from Interchange land, and visualizations work perfectly.

mattwthompson:
Thanks for the review

> export INTERCHANGE_EXPERIMENTAL=1 seems to be required

This shouldn't be the case in the 0.4.x line, possibly your local setup is out of date. I've removed some obsolete stuff in tests now that `Interchange.from_openmm` is not "experimental."

> I think we should consider if topologies matching the number of real atoms in the system should be accepted as well

Agree, but something for future work. I want to get this functionality over the finish line in its current scope as soon as I can

> virtual sites in a separate chain and residue

Right, this was the result of #1051

> we should account for this possibility

Probably yeah, topological edges being defined different ways is pretty frustrating to work with

openff/interchange/components/toolkit.py
mattwthompson: Nope, nice catch

openff/interchange/components/toolkit.py
mattwthompson: Possibly; I don't love the current design but it has the benefit of not interacting with anything the user might do in the public API

openff/interchange/interop/openmm/_import/compat.py
mattwthompson: Looks like I did this at some point

openff/interchange/components/toolkit.py
mattwthompson: https://github.com/openforcefield/openff-interchange/issues/1201

mattwthompson:
I think this is ready for another pass;
* #1150 should have fixed some errant need for using `INTERCHANGE_EXPERIMENTAL`
* Based on #1197, the first pass of this feature will require positions (if non-`None`) to be of length equal to the number of _particles_. I think supporting the case of being equal to the number of _atoms_ is valid, if less common, and I can't think of a reason why it wouldn't work. But I want to push that off into a future changeset.

My hope is that this is more or less good to go in its current state, but if there are more changes needed before it finds a way into a release that's fine. OpenFE is a key stakeholder for this feature but I don't think there's significant time pressure from them.

I think everything else is either out of the original scope or something I forgot!

Yoshanuikabundi (commented):
This is an enormous amount of code that lays the basis for a lot of future work and the fact that it works reflects a pretty amazing effort!

I checked this with a notebook I wrote. First, it creates a single "water" molecule topology and system by hand and imports that, then round trip exports it. Then I created a small box of OPC water (because it uses a 3 particle average virtual site) with Modeller, imported that, and roundtripped it. Notebook here - just rename to `.ipynb`: [openmm_vsite_import.txt](https://github.com/user-attachments/files/19972190/openmm_vsite_import.txt)
 Here's what I found:

- The modeller-created system worked perfectly and was easy to load, visualize, and export. This is probably the most fair and relevant test, and I imagine should represent the vast majority of OpenMM system use.
- My hand-built system imported and visualized fine, but when I roundtripped it back to OpenMM the resulting system didn't have any particles. I wonder if this had something to do with the fact that it didn't have any forces. This was really not a fair test or something anyone would ever do, but hey if that's worth fixing then there it is.
- The `topology` argument of `Interchange.from_openmm` is optional (has a default value, has `None` as a possible type, and is described as optional in the docstring), but a value of `None` unconditionally raises an error - this should just be a required argument. Not a blocker.
- An OpenFF topology can't be used, because it can't contain vsites and therefore has the wrong number of atoms - I mentioned this in my previous review and I think it got discussed as if I was talking about positions, but specifically I think that a topology that matches the number of real atoms should be accepted. This allows users to include complete chemical information instead of leaning on `_SimpleMolecule`s etc. Not a blocker here but definitely a nice to have.

Notes from looking at the code: 

- `LocalCoordinatesSite` is in `_SUPPORTED_VIRTUAL_SITE_TYPES`, but is not supported - I've added a note on the code below
- I have some notes on the docs page
- Users can currently provide OpenFF topologies when importing OpenMM topologies - I've used this in vignettes to demonstrate how users can take advantage of having full chemical information even when constructing an interchange from OpenMM. We should not regress this by having OpenFF topologies raise an `AttributeError`, even if we do not support this use case for systems with virtual sites. More details below.

All that said, I think this is just about ready to go and most of the above feedback can be addressed in future PRs. The three blockers I've identified so far are the suggestions I made to `edges.md`, removing `LocalCoordinatesSite` from `_SUPPORTED_VIRTUAL_SITE_TYPES`, and addressing the regression with OpenFF topologies.

I haven't looked in great detail at the guts of the collection or export code, and I'm happy to do that tomorrow - my eyes are just bouncing off it right now and its the end of my day. So I'll finish this off tomorrow - I expect to approve it, but such a detailed PR deserves a full review.

openff/interchange/_tests/unit_tests/interop/openmm/_import/test_virtual_sites.py
```diff
@@ -0,0 +1,102 @@
...
+        )
+
+        imported = Interchange.from_openmm(
+            topology=modeller.topology,
+            system=system,
+            positions=modeller.getPositions(),
+            box_vectors=modeller.getTopology().getPeriodicBoxVectors(),
+        )
+
+        get_openmm_energies(imported)
```
Yoshanuikabundi, line 47: This test could perhaps be improved by comparing the energies of `system` to `imported.to_openmm_system()` to ensure that they're within some bound? That would be a stopgap for the next test until `LocalCoordinatesSite` is supported.

Yoshanuikabundi (approved):
Internal code all looks good! I have suggested revisions, mostly to user-facing code and docs, but only the three points I raised in my previous comments are what I'd consider blocking, and I don't think any of them should need another round of review. Thanks for this feature, it opens the way for exciting things!

openff/interchange/interop/openmm/_import/_virtual_sites.py
```diff
@@ -0,0 +1,83 @@
...
+        return None
+
+    collection = _VirtualSiteCollection()
+
+    for particle_index in range(system.getNumParticles()):
+        if system.getParticleMass(particle_index)._value != 0.0:
+            continue
+
+        virtual_site = system.getVirtualSite(particle_index)
+
```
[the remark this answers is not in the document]
mattwthompson: Great find - textbook case of review uncovering uses I haven't thought of! - but in this case I think we should keep it as an error and think through the intended behavior in a separate PR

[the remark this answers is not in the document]
mattwthompson: Captured here: #1217

openff/interchange/_tests/unit_tests/interop/openmm/_import/test_virtual_sites.py
mattwthompson: Good idea, 8100ad8514b471b26993f235306c51cab8879b9d

openff/interchange/_tests/unit_tests/interop/openmm/_import/test_import.py
```diff
@@ -72,6 +73,8 @@ def test_different_ways_to_process_box_vectors(
         simple_system,
     ):
         topology = Molecule.from_smiles("C").to_topology()
+        topology._molecule_virtual_site_map = defaultdict(list)
+        topology._particle_map = {index: index for index in range(topology.n_atoms)}
```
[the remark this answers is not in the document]
mattwthompson: I've decided to make this change - even though it's probably a regression, this behavior was never tested. I'd like to come back to it and add tests / fix anything that this breaks. And I can hardly add more to this PR in its current state so I must do it later

#1218

PR #1085, opened 2024-10-29
Defer some third-party imports for quicker high-level imports

### LINKED ISSUE #239: Cleanup: Slow import times
While nearing a 0.1.0 release, I am taking some time to look at import times. The first pass of checks is, quite frankly, awful.

The quickest way I know how to evaluate import times is with [`tuna`](https://github.com/nschloe/tuna) and a one-liner like

```shell
python -X importtime -c "from openff.interchange.components.interchange import Interchange" 2> tuna.log && tuna tuna.log
```

The result against the most recent commit is [here](https://gist.github.com/mattwthompson/cc4b7fbb04be75db012eba8cb09fa90b):

image

A ***lot*** of this time is importing `unyt` and its dependencies. This is a [known issue](https://github.com/yt-project/unyt/issues/27#issuecomment-775503341), though there hasn't been much movement on it lately. Frustratingly, `unyt` is the only dependency pulling in `sympy`:

```shell
$ conda-tree whoneeds sympy
['gmso', 'unyt']
```

Dropping those (and by extension `foyer`) the results are still [bad, but better](https://gist.github.com/mattwthompson/91c91f94d464c75de193125ec874c350):

image

`pint` is somewhat unfairly represented as a culprit here; `pint-pandas` is responsible for about 0.25 seconds of import time by nature of pulling `pandas` in (which `pint` does not do out of the box).

That's all the analysis for now; I'm going to look into lazy-loading MDTraj and possibly  a few other modules in an effort to trim off some fraction of a second more.

mattwthompson:
### Description

Resolves #239, for now?

I mean, we could defer loading `openff.toolkit.ForceField` but that seems obtuse

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
image

openff/interchange/components/potentials.py
```diff
@@ -25,15 +26,13 @@
...
-
-from numpy.typing import ArrayLike
-
-if has_package("jax"):
-    from jax import Array
+if TYPE_CHECKING:
+    if has_package("jax"):
+        from jax import Array
+    else:
+        Array = Any  # type: ignore
```
Yoshanuikabundi, line 33: This [seems](https://docs.python.org/3/library/typing.html#type-aliases) to be the preferred syntax for type aliases prior to 3.12 - does it resolve the `type:ignore`?

```suggestion
        Array: TypeAlias = Any
```

openff/interchange/components/potentials.py
```diff
@@ -2,9 +2,10 @@
 
 import ast
 import json
-from typing import Annotated, Any
+from typing import TYPE_CHECKING, Annotated, Any
```
Yoshanuikabundi, line 5:
```suggestion
from typing import TYPE_CHECKING, Annotated, Any, TypeAlias
```

Needed for some suggestions below.

openff/interchange/components/potentials.py
```diff
@@ -25,15 +26,13 @@
...
-if has_package("jax"):
-    from jax import Array
+if TYPE_CHECKING:
+    if has_package("jax"):
+        from jax import Array
+    else:
+        Array = Any  # type: ignore
 else:
-    Array = Any  # type: ignore
+    Array = ArrayLike
```
Yoshanuikabundi, line 35:
```suggestion
    Array: TypeAlias = ArrayLike
```

Yoshanuikabundi (approved):
> I mean, we could defer loading `openff.toolkit.ForceField` but that seems obtuse

A little, but my inclination would be to do it:

- it's almost all of the remaining import time according to your graphic (if it ends up being imported anyway I wouldn't bother)
- there are some workflows in important growth areas that I don't believe need `ForceField`:
   - workflows importing from OpenMM or Foyer and not using any SMIRNOFF force field
   - workflows loading a previously created Interchange directly from disk (this may be important for determinism)
- the API points that need it all take it as an argument and call methods on it, so it's going to be imported by the user in most cases, which means complaints about import times will probably go to the right place

Other than that, this LGTM! I have a suggestion about indicating type aliases more explicitly that may or may not make the type checker happier but it's entirely optional.

mattwthompson:
> my inclination would be to do it

Good points, I'm convinced

Unfortunately, most of the time is spent importing the toolkit's wrapper machinery (I'd rather RDKit and OEChem not be needed until they are - separate push for a separate time)

image

PR #1095, opened 2024-11-06
Added triclinic boxes for lammps

wutobias:
### Description

This PR adds functionality to use triclinc boxes in simulations with LAMMPS. The simulation box vectors are transformed to LAMMPS's [restricted triclinix box format](https://docs.lammps.org/Howto_triclinic.html). A warning is raised when the box vectors before and after the transformation come out to be different. Users want to be aware of that in case they start with a pre-existing set of velocities. See also the [LAMMPS documentation](https://docs.lammps.org/Howto_triclinic.html#general-triclinic-simulation-boxes-in-lammps).

### Checklist

- [x] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
+++ b/openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
@@ -27,6 +27,69 @@ class TestLammps:
             "C1COC(=O)O1",  # This adds an improper, i2
         ],
     )
+    def test_to_lammps_single_mols_triclinic(
+        self,
+        mol: str,
+        sage_unconstrained: ForceField,
+        n_mols: int,
+    ) -> None:
+        """
+        Test that Interchange.to_openmm Interchange.to_lammps report sufficiently similar energies
+        in triclinic simulation boxes.
+        """
+        mol = MoleculeWithConformer.from_smiles(mol)
+        mol.conformers[0] -= numpy.min(mol.conformers[0], axis=0)
+        top = Topology.from_molecules(n_mols * [mol])
+
+        box = numpy.zeros((3,3), dtype=float) * unit.angstrom
+
+        box[0] = [51.34903463831951, 0, 0] * unit.angstrom
+        box[1] = [-0.03849979989403723, 50.9134404144338, 0] * unit.angstrom
+        box[2] = [-2.5907782992729538, 0.3720740833800747, 49.80705567557188] * unit.angstrom
+
+        top.box_vectors = box
+
+        if n_mols == 1:
+            positions = mol.conformers[0]
+        elif n_mols == 2:
+            positions = numpy.concatenate(
+                [mol.conformers[0], mol.conformers[0] + 1.5 * unit.nanometer],
+            )
+
+        interchange = Interchange.from_smirnoff(sage_unconstrained, top)
+        interchange.positions = positions
+        interchange.box = top.box_vectors
+
+        reference = get_openmm_energies(
+            interchange=interchange,
+            round_positions=3,
+        )
+
+        lmp_energies = get_lammps_energies(
+            interchange=interchange,
+            round_positions=3,
+        )
+
+        lmp_energies.compare(
+            reference,
+            tolerances={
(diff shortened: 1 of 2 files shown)


openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
```diff
@@ -15,6 +15,69 @@
...
+            interchange=interchange,
+            round_positions=3,
+        )
+
+        lmp_energies.compare(
+            reference,
+            tolerances={
+                "Nonbonded": 1 * unit.kilojoule_per_mole,
+                "Torsion": 0.02 * unit.kilojoule_per_mole,
+            },
```
mattwthompson, line 78: I figure this is copy-pasted code from my own technical debt, but what happens when this is set to a lower threshold? I only care about the non-bonded energies here, since those would be what are wrong if the box vectors aren't written correctly
```suggestion
            tolerances={
                "Nonbonded": 0.01 * unit.kilojoule_per_mole,
                "Torsion": 0.02 * unit.kilojoule_per_mole,
            },
```

wutobias: Right, this is copy-paste. I agree it would make sense to have tighter tolerance. I have tested this localy with 0.01 kJ/mol energy tolerance. Some of the triclinic box tests are failing, but so are they in the regular test `test_to_lammps_single_mols` with default rectangular boxes when using this low energy tolerance... Something else is going on I think.

mattwthompson: Right, there are some funky things going on with how OpenFF's non-bonded settings get translated differently to LAMMPS than OpenMM, and that's out of scope here. Still, it'd be nice to have this value dropped a bit (i.e. even 0.1 would be great, if that passes tests) and a comment encoding what you just said.

mattwthompson (commented):
Are there any tools that will parse LAMMPS data files and read these box vectors? I know MDTraj can read some LAMMPS trajectories and InterMol can convert LAMMPS to GROMACS and maybe some other formats. Maybe the LAMMPS API itself is the easiest way to get this done.

What I'm looking for, because of the (upstream) non-bonded quirks, is a test that does the first half of what you've already done, but also writes out an `openmm.System` and then reads them back in and compares the box vectors.

If there isn't a quick way to do that, I'll just write a few files locally and see if I can visualize as a spot-check. But it'd be great to have it in CI right off the bat.

openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
wutobias: 0.4 kJ/mol is the lowest I can drop this tolerance before the `test_to_lammps_single_mols` (and hence the triclinic one) fails.

mattwthompson: Works for me, thanks

wutobias:
> Are there any tools that will parse LAMMPS data files and read these box vectors? I know MDTraj can read some LAMMPS trajectories and InterMol can convert LAMMPS to GROMACS and maybe some other formats. Maybe the LAMMPS API itself is the easiest way to get this done.

I do not know. I am relatively new to LAMMPS, but I guess the box vectors should be stored in the trajectory (at least in NPT runs), which can be read by MDTraj.


> What I'm looking for, because of the (upstream) non-bonded quirks, is a test that does the first half of what you've already done, but also writes out an openmm.System and then reads them back in and compares the box vectors.

Not 100% answering the question, but the box vectors should be identical to OpenMM box vectors. I used [this code from openmm](https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/internal/unitcell.py) as a starting point.

openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
wutobias: I modified the test so that it tests whether we get a higher energy deviation with the triclinic test than the orthogonal one. Is that maybe more to the point of what we actually want to test?

PR #1100, opened 2024-11-12
Constrain `openmm =8.1.2` in tests

mattwthompson:
### Description

Something changed in regression tests, but not in a way I can reproduce locally


mattwthompson:
Tag #1101

PR #1103, opened 2024-11-13
Fix RTD build

Yoshanuikabundi:
### Description

RTD builds are succeeding, but with warnings that suggest that API docs are not being generated. I can't reproduce the issue locally so I'm going to poke around here.

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -14,3 +14,8 @@ sphinx:
 
 conda:
   environment: devtools/conda-envs/docs_env.yaml
+
+python:
+  install:
+    - method: pip
+      path: .
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -10,7 +10,7 @@ build:
 
 sphinx:
   configuration: docs/conf.py
-  fail_on_warning: false
+  fail_on_warning: true
 
 conda:
   environment: devtools/conda-envs/docs_env.yaml
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -10,6 +10,8 @@ build:
 
 sphinx:
   configuration: docs/conf.py
+  # Interchange's docs should not produce warnings; if they start,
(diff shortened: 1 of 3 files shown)


Yoshanuikabundi:
OK this was unhinged.

Just a note - Sphinx warnings in Interchange mean the documentation is not building correctly. I know Sphinx produces a lot of noise most of the time, but any warnings produced in this repo are real. In this case, the entire API documentation was not building and was absent, so the docs really were badly broken - missing references really should be a hard error.

The error was relating to Autosummary not being able to import the `openff.interchange` module, so it couldn't produce API docs. The `sys.path.insert(0, os.path.abspath("../"))` line in `docs/conf.py` is supposed to tell Sphinx where to find this as by default RTD doesn't install the package in the repo it's building for. I've now configured RTD to install Interchange, and the `sys.path.insert` line is no longer needed. However, this wasn't the problem.

The problem was that the call to `version("openff.interchange")` was failing because of a mismatch between the path to the module and the name of the package. VersionInGit was loading the package name from the `project.name` entry in `pyproject.toml`, which was `Interchange`. Importlib then couldn't match that up with the module path `openff.interchange`. Changing `project.name` from `Interchange` to `openff-interchange` fixed this problem. This also means that Interchange will now show up with the same name whether its installed via pip (eg, in dev installs) or conda. I don't know why things were working everywhere except the docs - I'd guess it has something to do with the fact that conf.py is `exec()`ed rather than imported.

I've also made some changes to conf.py for a bit more robustness in the face of some changes RTD is making to default configurations.

mattwthompson:
> any warnings produced in this repo are real

Yup, turning it off was intentional since these partial build failures were blocking other development

> VersionInGit was loading the package name from

This makes sense! I'm trying to keep human-facing branding as "Interchange" but didn't think of this connection

mattwthompson (approved):

PR #1104, opened 2024-11-13
Consolidate Python config files

mattwthompson:
### Description

I don't like TOML


mattwthompson:
Unfortunately, <stuff> is looking at `openff/interchange/foyer/*.py` and not the proper location when doing `from foyer import ...`


This is probably the cause?
```toml
[tool.setuptools.packages.find]
where = ["openff/interchange"]

File: openff/interchange/_tests/unit_tests/components/test_packmol.py [hunk 1]
```diff
@@ -388,24 +388,6 @@ def test_packmol_add_negative_solvent_mass(self):
...
-            match="could not be parsed by RDKit",
-        ):
-            pack_box(
-                molecules=[Molecule.from_smiles("CCO")],
-                number_of_copies=[11112],
-                box_shape=UNIT_CUBE,
-                tolerance=1.0 * unit.angstrom,
-                target_density=0.1 * unit.grams / unit.milliliters,
-            )
-
```
j-wags, line 408: (not blocking) This may have slipped in from another changeset - I'm not sure what removing this test has to do with changing config files/how OpenMM is handled in the test matrix. Not blocking, just wanted to make sure you saw this was here.

j-wags (approved):

openff/interchange/_tests/unit_tests/components/test_packmol.py
mattwthompson: Nice find, not sure how that snuck in there. Some tests are duplicated but this isn't such a case

mattwthompson:
Well that was (not) fun. Some things I baked in here

* Setting code coverage option at the project level conflicts with my desire to not track coverage in examples. Hence `-c /dev/null`
* Having a module named `foyer/__init__.py` should be fine but I ran into some issues which were hard to pinpoint or absolve away. I know not to have a `foyer.py` file, but I thought modules were okay.
  * This can cause a sort of false positive of `has_package("foyer")` which picks up something from the wrong path and returns `True`, so at a few points I look to see if I can grab `from foyer.forcefield import Forcefield` or something like that and store it as a boolean.
* I'm still not sure what `tool.setuptools.packages.find` does
* `tool.mypy.python_version` only sets what Python version is used when `mypy` is called and can be updated whenever

PR #1105, opened 2024-11-14
Document Pydantic/Pint helpers

mattwthompson:
### Description

Ref https://github.com/openforcefield/status/issues/96

https://openff-docs--1105.org.readthedocs.build/projects/interchange/en/1105/annotations.html

### Checklist

- [ ] Add tests
- [x] Lint
- [ ] Update docstrings

### DIFF
--- a/docs/annotations.md
+++ b/docs/annotations.md
@@ -1,0 +1,102 @@
+# Annotations
+
+This document was written in November 2024 assuming
+
+* Python 3.10+
+* Pydantic V2 (not the V1 API backdoor)
+* Mypy 1.10+ or similarly-functional Python type-checker
+* Interchange 0.4+
+
+Consider a class to store information about atoms. It needs to store an atomic mass, a partial charge, and a position in 3-D space. The following class would work fine:
+
+```python
+class Atom:
+    mass: float
+    charge: float
+    position: list
+```
+
+but this has some shortcomings:
+
+* What units are assumed for each attribute?
+* What if the position is a `numpy.ndarray` (or `openmm.Vec3`)?
+* Can instances of this class be serialized to disk? And then de-serialized?
+* Are any fields optional?
+
+OpenFF's combined use of [Pint](https://pint.readthedocs.io/en/stable/) and [Pydantic](https://docs.pydantic.dev/latest/) solves these problems. To leverage these tools,
+
+1. Classes should subclass out of `openff.interchange.pydantic._BaseModel`, which is a thin wrapper around `pydantic.BaseModel`
+2. Annotations should use `openff.interchange._annotations._Quantity`, which is a thin wrapper around `openff.units.Quantity`, itself having the same behavior as `pint.Quantity`, or annotations derived thereof.
+
+The same class can be rewritten:
+
+```python
+from openff.interchange.pydantic import _BaseModel
+from openff.interchange._annotations import _Quantity
+from openff.units import Quantity
+
+
+class Atom(_BaseModel):
+    mass: _Quantity
+    charge: _Quantity
+    position: _Quantity | None = None
+
+
+carbon = Atom(
+    mass=_Quantity("12.011 amu"),
+    charge=_Quantity("-0.1 elementary_charge"),
+    position=_Quantity([0, 0, 0], "nanometer"),
+)
+
+print(carbon.model_dump())
+# {'mass': {'val': 12.011, 'unit': 'unified_atomic_mass_unit'}, 'charge': {'val': -0.1, 'unit': 'elementary_charge'}, 'position': {'val': [0, 0, 0], 'unit': 'nanometer'}}
+
+print(carbon.model_dump_json())
+# {"mass":{"val":12.011,"unit":"unified_atomic_mass_unit"},"charge":{"val":-0.1,"unit":"elementary_charge"},"position":{"val":[0,0,0],"unit":"nanometer"}}
+
+print(Atom.model_validate_json(carbon.model_dump_json()))
+# Atom(mass=<Quantity(12.011, 'unified_atomic_mass_unit')>, charge=<Quantity(-0.1, 'elementary_charge')>, position=<Quantity([0 0 0], 'nanometer')>)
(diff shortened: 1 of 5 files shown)


openff/interchange/_annotations.py
```diff
@@ -151,6 +155,20 @@ def quantity_json_serializer(
     WrapSerializer(quantity_json_serializer),
 ]
 
+_MassQuantity = Annotated[
+    Quantity,
+    WrapValidator(quantity_validator),
+    AfterValidator(_is_mass),
+    WrapSerializer(quantity_json_serializer),
+]
+
```
mattwthompson, line 164: @Yoshanuikabundi these things (I don't know the right noun ...) appear to be re-usable as classes, not just annotations

1. Am I ... correct?
2. Should this magic live elsewhere, like in `openff-units`?
3. Should these things be "public" in some way?

Yoshanuikabundi: This is extremely cool. 

1. I think so - It looks like `Annotated[T]` is just a type alias for `T` that maintains a particular runtime value of the `__metadata__` attribute: https://github.com/python/cpython/blob/3.13/Lib/typing.py#L2210

2. I think for the time being putting this in `openff-models` would make the most sense. That way it can be used by all our Pydantical projects without being confused for a Pint thing. Since the validators would only kick in when the `Annotated[Quantity, ...]` is in a Pydantic model, I'd be wary of putting it in `openff-units`

3. I think having this be public in `openff-models` would be fine. I would love something like this to be public in `openff-units`, but the API there should be used something like:

```python
def foo(length: Quantity[NDArray, "angstrom"]):
   ...
```

I've taken a quick stab at this just now and it's a bit hairy but I think should be possible. In the interim, having this available for Pydantic projects would be great.

docs/annotations.md
```diff
@@ -0,0 +1,106 @@
+# Annotations
+
+This document was written in November 2024 assuming
+
+* Python 3.10+
+* Pydantic V2 (not the V1 API backdoor)
+* Mypy 1.10+ or similarly-functional Python type-checker
+* Interchange 0.4+
```
Yoshanuikabundi, line 8:
```suggestion
:::{note}
  This document was written in November 2024 assuming
  
  * Python 3.10+
  * Pydantic V2 (not the V1 API backdoor)
  * Mypy 1.10+ or similarly-functional Python type-checker
  * Interchange 0.4+
:::
```

Bit clearer what's going on if this is in its own box

docs/annotations.md
```diff
@@ -0,0 +1,106 @@
...
+
+* What units are assumed for each attribute?
+* What if the position is a `numpy.ndarray` (or `openmm.Vec3`)?
+* Can instances of this class be serialized to disk? And then de-serialized?
+* Are any fields optional?
+
+OpenFF's combined use of [Pint](https://pint.readthedocs.io/en/stable/) and [Pydantic](https://docs.pydantic.dev/latest/) solves these problems. To leverage these tools,
+
+1. Classes should subclass out of `openff.interchange.pydantic._BaseModel`, which is a thin wrapper around `pydantic.BaseModel`
+2. Annotations should use `openff.interchange._annotations._Quantity`, which is a thin wrapper around `openff.units.Quantity`, itself having the same behavior as `pint.Quantity`, or annotations derived thereof.
```
Yoshanuikabundi, line 29:
```suggestion
2. Annotations should use the [`openff.interchange._annotations._*Quantity`] types, which are thin wrappers around `openff.units.Quantity`, itself having the same behavior as `pint.Quantity`, or annotations derived thereof.

[`openff.interchange._annotations._*Quantity`]: https://github.com/openforcefield/openff-interchange/blob/v0.4.0/openff/interchange/_annotations.py#L127-L191
```

Just to give us an easy way to see what's available. If this were a public API, this would obvs be a link to the appropriate module's API docs.

docs/annotations.md
```diff
@@ -0,0 +1,106 @@
...
+print(
+    Atom(
+        mass=Quantity(12.011, "amu"),
+        charge=Quantity("-0.1 / 6.241509074460763e+18 coulomb"),
+    ).charge
+)
+# -0.10000000000000002 elementary_charge
+```
+
+This can be useful if i.e. it is important that positions are universally stored with Angstroms and not nanometers.
```
Yoshanuikabundi, line 106:
```suggestion
This can be useful if i.e. it is important that positions are universally stored with Angstroms and not nanometers.

Note that this `Annotated` trick is specifically used by Pydantic classes, and will cause an error at runtime when violated - static type checkers like MyPy and PyRight will not be able to validate that units and dimensionality are correct.
```

docs/annotations.md
```diff
@@ -0,0 +1,106 @@
...
+
+print(carbon.model_dump())
+# {'mass': {'val': 12.011, 'unit': 'unified_atomic_mass_unit'}, 'charge': {'val': -0.1, 'unit': 'elementary_charge'}, 'position': {'val': [0, 0, 0], 'unit': 'nanometer'}}
+
+print(carbon.model_dump_json())
+# {"mass":{"val":12.011,"unit":"unified_atomic_mass_unit"},"charge":{"val":-0.1,"unit":"elementary_charge"},"position":{"val":[0,0,0],"unit":"nanometer"}}
+
+print(Atom.model_validate_json(carbon.model_dump_json()))
+# Atom(mass=<Quantity(12.011, 'unified_atomic_mass_unit')>, charge=<Quantity(-0.1, 'elementary_charge')>, position=<Quantity([0 0 0], 'nanometer')>)
+```
```
Yoshanuikabundi, line 59: You could consider a doctest here: 

```python
>>> print("hello world")
hello world
```

Might be a bit clearer

Yoshanuikabundi (approved):
Awesome love it. Few minor notes on the presentation of the docs and a little added context.

docs/annotations.md
mattwthompson: Would the runner run through Markdown files like so (provided it's pointed to them)?

PR #1110, opened 2024-11-15
Do not subtract tolerance from box size

### LINKED ISSUE #1106: Interchange subtracts tolerance when packing boxes, whereas Evaluator didn't.
**Description**

Not so much a bug report as an FYI for discussion -- there are still differences between Packmol behaviour between Evaluator and Interchange, and I believe one of them is down to the line below. Evaluator's input file does not seem to have subtracted the tolerance:

**Reproduction**

Please include a [minimally reproducing example](https://stackoverflow.com/help/minimal-reproducible-example) of this bug.

https://github.com/openforcefield/openff-interchange/blob/8d2cee156c8a6763fde0e1938809a96b688325da/openff/interchange/components/_packmol.py#L477

**Output**

Using a mixture of triethanolamine and water, modern Evaluator builds the following input file:

**Evaluator packmol_input.txt**
```
tolerance 2.000000
filetype pdb
output packmol_output.pdb

structure 0.pdb
  number 490
  inside box 0. 0. 0. 57.63472087723553 57.63472087723553 57.63472087723553
end structure

structure 1.pdb
  number 510
  inside box 0. 0. 0. 57.63472087723553 57.63472087723553 57.63472087723553
end structure
```

Interchange 0.4.0:

```
tolerance 2.000000
filetype pdb
output packmol_output.pdb

structure _PACKING_MOLECULE0.pdb
  number 490
  inside box 0. 0. 0. 55.63472087723553 55.63472087723553 55.63472087723553
end structure

structure _PACKING_MOLECULE1.pdb
  number 510
  inside box 0. 0. 0. 55.63472087723553 55.63472087723553 55.63472087723553
end structure
```

And the box size from density from both returns ~57.63472087723553:

Evaluator:
```
from openff.evaluator.utils.packmol import _approximate_box_size_by_density
from openff.toolkit.topology.molecule import Molecule, unit
from openff.evaluator import unit as eval_unit
from openff.evaluator.utils.openmm import openmm_quantity_to_pint
import openff.evaluator
import openff.toolkit

smiles = "OCCN(CCO)CCO"

comp0 = Molecule.from_smiles(smiles)
comp1 = Molecule.from_smiles("O")
copies = [490, 510]
mass_density = 0.95 * eval_unit.grams / eval_unit.milliliters
box_aspect_ratio = [1, 1, 1]

box = _approximate_box_size_by_density(
    [comp0, comp1],
    copies,
    mass_density,
    box_aspect_ratio,
)

# 	[57.63472087723553 57.63472087723553 57.63472087723553]

```

Interchange:

```
from openff.interchange.components._packmol import _box_from_density
from openff.toolkit.topology.molecule import Molecule, unit
import openff.interchange
import openff.toolkit
import numpy as np

box = _box_from_density(
    [comp0, comp1],
    copies,
    mass_density,
    np.identity(3),
)

# [[57.63472087723553 0.0 0.0]
# [0.0 57.63472087723553 0.0]
# [0.0 0.0 57.63472087723553]]
```

In a fun twist, I'm not getting Evaluator's input box to actually pack for me either in this separate workflow (it has worked fine in an actual run of Evaluator....) so there's also other stuff going on here, but thought I'd raise it.

**Software versions**

mattwthompson:
### Description

Resolves #1106

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/components/_packmol.py
+++ b/openff/interchange/components/_packmol.py
@@ -474,7 +474,7 @@ def _build_input_file(
         The path to the output file.
 
     """
-    box_size = (box_size - tolerance).m_as("angstrom")
+    box_size = box_size.m_as("angstrom")
     tolerance = tolerance.m_as("angstrom")
 
     # Add the global header options.


mattwthompson:
@lilyminium @j-wags do we want this?

Yoshanuikabundi:
When I wrote this code, I was concerned that packmol might not produce boxes that tile very well, and rather than introduce a second parameter I re-used the tolerance parameter to introduce a small buffer around the edge of the box. This only happens in the call to packmol - the actual returned box should be the requested size or density, just with that small empty buffer on one end of each dimension. In other words, packmol attempts to create a slightly overdense box, which is then placed in the corner of the returned box. After NVT equilibration, the box would be the requested uniform density. Part of my assumption was that PackMOL outputs are really only useful as starting points for equilibration.

lilyminium:
Josh and I just had a quick call where we talked over the above, and we both agreed we'd prefer the option that works generally across systems. Since all the tests I ran to troubleshoot the original packing issue did not implement this fix, that means I have a weak preference to *not* merge this PR since we know things work without it. I'm happy to re-run tests with this fix to validate if anyone has a strong opinion the other way of course! (And this behaviour in Evaluator seems to work ok)

mattwthompson:
Thanks all, my preferences here are weak so I'll not go with this PR but make an attempt to document this behavior (whether or not it's different than Evaluator's code)

mattwthompson:
I will say that it's a little odd that the density is lower because of the 10% scale-up in each linear dimension (33% scaleup in volume) and higher because of how the "tolerance" is subtracted from the edges of the box.

PR #1111, opened 2024-11-15
Discard other conformers when packing

mattwthompson:
### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/components/_packmol.py
+++ b/openff/interchange/components/_packmol.py
@@ -423,6 +423,10 @@ def _create_molecule_pdbs(molecules: list[Molecule]) -> list[str]:
         # Generate conformers if they're missing
         if molecule.n_conformers <= 0:
             molecule.generate_conformers(n_conformers=1)
+        else:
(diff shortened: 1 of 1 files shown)


openff/interchange/components/_packmol.py
```diff
@@ -423,6 +423,10 @@ def _create_molecule_pdbs(molecules: list[Molecule]) -> list[str]:
         # Generate conformers if they're missing
         if molecule.n_conformers <= 0:
             molecule.generate_conformers(n_conformers=1)
+        else:
+            # Possible issues writing multi-conformer PDBs with RDKit
+            # https://github.com/openforcefield/openff-interchange/issues/1107
+            molecule._conformers = [molecule.conformers[0]]
```
j-wags, line 429: This seems really simple and straightforward - I think it's unlikely to interact with an eventual fix to https://github.com/openforcefield/openff-toolkit/issues/1967

j-wags (approved):

PR #1118, opened 2024-12-03
Improve testing of GROMACS import

mattwthompson:
### Description

- [x] Add more tests importing GROMACS files
- [ ] Look into constraint estimation

### Checklist

- [x] Add tests
- [x] Lint
- [ ] Update docstrings


mattwthompson:
This could be ramped up a bit by running each of these tests as monolithic vs ITP structure

openff/interchange/_tests/_gromacs.py
```diff
@@ -0,0 +1,21 @@
...
+from openff.utilities import temporary_cd
+
+from openff.interchange import Interchange
+from openff.interchange.components.mdconfig import get_smirnoff_defaults
+from openff.interchange.drivers import get_gromacs_energies
+
+
+def gmx_roundtrip(state: Interchange, apply_smirnoff_defaults: bool = False):
+    with temporary_cd():
+        state.to_gromacs(prefix="state", decimal=8)
```
mattwthompson, line 12:
```suggestion
@pytest.mark.parametrize("monolithic", [False, True])
def gmx_roundtrip(state: Interchange, apply_smirnoff_defaults: bool = False, monolithic):
    with temporary_cd():
        state.to_gromacs(prefix="state", decimal=8, monolithic=monolithic)
```

Intended change in #1120 after this is merged in

openff/interchange/_tests/__init__.py
```diff
@@ -76,9 +76,9 @@ class MoleculeWithConformer(Molecule):
     """Thin wrapper around `Molecule` to produce an instance with a conformer in one call."""
 
     @classmethod
-    def from_smiles(self, smiles, name="", **kwargs):
+    def from_smiles(self, smiles, name="", *args, **kwargs):
         """Create from smiles and generate a single conformer."""
-        molecule = super().from_smiles(smiles, **kwargs)
+        molecule = super().from_smiles(smiles, *args, **kwargs)
```
j-wags, line 81: (not blocking) seems to be irrelevant to this changeset (not used anywhere else in the PR) but it makes sense as a standalone change.

j-wags (approved):
LGTM, one nonblocking suggestion on a slightly confusing test method but otherwise good to go!

openff/interchange/_tests/_gromacs.py
```diff
@@ -3,14 +3,22 @@
 from openff.utilities import temporary_cd
 
 from openff.interchange import Interchange
+from openff.interchange.components.mdconfig import get_smirnoff_defaults
 from openff.interchange.drivers import get_gromacs_energies
 
 
-def gmx_monolithic_vs_itp(state: Interchange):
+def gmx_roundtrip(state: Interchange, apply_smirnoff_defaults: bool = False):
```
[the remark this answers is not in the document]
mattwthompson: Good find, I think that was incidentally removed in a merge

openff/interchange/_tests/__init__.py
mattwthompson: Right, I think there was a reason in some tinkering that didn't make it to the end. I agree it stands on its own so I left it in

PR #1120, opened 2024-12-06
Read GROMACS topology files split into `.itp` files

mattwthompson:
### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/interop/gromacs/_import/combine.py
+++ b/openff/interchange/interop/gromacs/_import/combine.py
@@ -1,0 +1,48 @@
+from pathlib import Path
+
+
+def make_monolithic(topology_file: str, keep_file: bool = False) -> list[str]:
+    parent = Path(topology_file).parent
+    name = Path(topology_file).name
+
+    monolithic_path = parent / f"MONOLITHIC_{name}"
+
+    return_value = list()
+
+    with open(topology_file) as in_obj, open(monolithic_path, "w") as out_obj:
+        for line in in_obj:
+            if len(line) == 0:
+                continue
+
+            stripped = line.strip().split()
+
+            if len(stripped) == 0:
+                continue
+
+            if stripped[0].startswith(";"):
+                continue
+
+            if stripped[0].startswith("#include"):
+                itp_file = stripped[1].strip('"')
+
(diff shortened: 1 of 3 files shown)


mattwthompson:
Needs #1118 to really be tested

openff/interchange/interop/gromacs/_import/combine.py
```diff
@@ -0,0 +1,48 @@
...
+            if len(stripped) == 0:
+                continue
+
+            if stripped[0].startswith(";"):
+                continue
+
+            if stripped[0].startswith("#include"):
+                itp_file = stripped[1].strip('"')
+
+                with open(itp_file) as itp_obj:
```
Yoshanuikabundi, line 28: TOP files produced by GROMACS typically include ITP files distributed with GROMACS stored in `${GMXDATA}/top`. It might be worth attempting to load files from there if no local file can be found, eg something like:

```suggestion
                try:
                    included_file = open(itp_file)
                except FileNotFoundError as e:
                    included_file = open(f"{sys.env.get("GMXDATA", "")}/{itp_file}")
                with included_file as itp_obj:
```

I think in practice this won't actually allow many more TOP files to be loaded as they also mostly need `#IFDEF...#ENDIF` handling to deal with position restraints, but it'd be good future proofing.

Yoshanuikabundi (approved):
This looks good to me if the goal is to load non-monolithic TOP files produced by Interchange! I have one note but it's probably best left for the future.

To load typical TOP files produced by GROMACS, we probably need to also support `#IFDEF...#ELSE...#ENDIF`, `#DEFINE`, and use of macros defined by `#DEFINE`

openff/interchange/interop/gromacs/_import/combine.py
mattwthompson: I'd really like to avoid looking in `$GMXDATA` unless we're really sure we want to mix-and-match the force fields that are shipped there - this is worth considering for a future changeset but it's enough of a change in scope that I want to do it extremely cautiously and on an as-needed basis

mattwthompson:
Thank you @Yoshanuikabundi!

PR #1134, opened 2025-01-08
Create blank collections when importing only water

### LINKED ISSUE #1132: Interchange from_openmm for a water box does not work.
**Description**

`Interchange.from_openmm` does not work for water systems. This is due to the fact that `interchange.collections["Bonds"]` is not created and there for when water [bonds are filled](https://github.com/openforcefield/openff-interchange/blob/a70254f129c46e709c45e09f945901c3a7ed8e66/openff/interchange/interop/openmm/_import/_import.py#L125) the `KeyError` is raised.

**Reproduction**

A simple code for water box system can be used (I was using gromacs topology to create the system, but I believe that any other way to create systems will work as well):

```python
system_gro = GromacsGroFile(output.gro)
system_top = GromacsTopFile(
    output.top,
    periodicBoxVectors=system_gro.getPeriodicBoxVectors(),
)
system = system_top.createSystem(
    nonbondedMethod=PME,
    nonbondedCutoff=1*nanometer,
    constraints=HBonds
)
system_interchange = Interchange.from_openmm(
    system,
    system_top.topology,
    system_gro.positions,
    box_vectors=system_gro.getPeriodicBoxVectors(),
)
```

**Output**

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/pbuslaev/conda_env/astex_fep_noneq_test/lib/python3.11/site-packages/openff/interchange/components/interchange.py", line 922, in from_openmm
    return from_openmm(
           ^^^^^^^^^^^^
  File "/home/pbuslaev/conda_env/astex_fep_noneq_test/lib/python3.11/site-packages/openff/utilities/utilities.py", line 81, in wrapper
    return function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pbuslaev/conda_env/astex_fep_noneq_test/lib/python3.11/site-packages/openff/interchange/interop/openmm/_import/_import.py", line 125, in from_openmm
    if interchange.topology.n_bonds > len(interchange.collections["Bonds"].key_map):
                                          ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
KeyError: 'Bonds'

**Possible solution**
I think that the easiest way is to create the empty collection for `Bonds`, `Angles`, and `ProperTorsions` if they do not exist. So, after this [block](https://github.com/openforcefield/openff-interchange/blob/a70254f129c46e709c45e09f945901c3a7ed8e66/openff/interchange/interop/openmm/_import/_import.py#L79C9-L98) something like this can be added:
```python
if ["Bonds"] not in interchange.collections:
     interchange.collections["Bonds"] = BondCollection()
     ...
```

mattwthompson:
### Description

Fixes #1132

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/interop/openmm/_import/_import.py
+++ b/openff/interchange/interop/openmm/_import/_import.py
@@ -122,7 +122,12 @@ def from_openmm(
     # TODO: Does this run through the Interchange.box validator?
     interchange.box = _box_vectors
 
-    if interchange.topology.n_bonds > len(interchange.collections["Bonds"].key_map):
+    try:
+        num_physics_bonds = len(interchange["Bonds"].key_map)
+    except LookupError:
+        num_physics_bonds = 0
+
+    if interchange.topology.n_bonds > num_physics_bonds:
         # There are probably missing (physics) bonds from rigid waters. The topological
         # bonds are probably processed correctly.
         _fill_in_rigid_water_bonds(interchange)
@@ -347,6 +352,12 @@ def _fill_in_rigid_water_bonds(interchange: "Interchange"):
 
     from openff.interchange.components.potentials import Potential
     from openff.interchange.models import AngleKey, BondKey, PotentialKey
+
+    if "Bonds" not in interchange.collections:
+        interchange.collections.update({"Bonds": BondCollection()})
+
+    if "Angles" not in interchange.collections:
+        interchange.collections.update({"Angles": AngleCollection()})
 
     simple_water = _SimpleMolecule.from_molecule(Molecule.from_smiles("O"))
 
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -22,6 +22,7 @@ Please note that all releases prior to a version 1.0.0 are considered pre-releas
 
 * #1070 The `charge_from_molecules` argument must include only molecules that contain partial charges and are non-isomorphic with each other.
 * #1070 The `charge_from_molecules` argument as used by the OpenFF Toolkit is handled internally as `molecules_with_preset_charges`.
+* #1134 Fixes a bug in which importing OpenMM systems containing only rigid water would crash.
 
 ### Performance improvements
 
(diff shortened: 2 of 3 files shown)


openff/interchange/interop/openmm/_import/_import.py
```diff
@@ -348,6 +353,12 @@ def _fill_in_rigid_water_bonds(interchange: "Interchange"):
     from openff.interchange.components.potentials import Potential
     from openff.interchange.models import AngleKey, BondKey, PotentialKey
 
+    if "Bonds" not in interchange.collections:
+        interchange.collections.update({"Bonds": BondCollection()})
+
+    if "Angles" not in interchange.collections:
+        interchange.collections.update({"Angles": AngleCollection()})
+
```
pbuslaev, line 361: In my personal opinion it is a bit more logical to create those empty collections earlier, when all the collections are created. Those collections are anyway created, since `_fill_in_rigid_water_bonds` is anyway called, but now, `_fill_in_rigid_water_bonds` has sort of two tasks, which are not logically connected.

pbuslaev (commented):
I think the change is good, though my minor bit is about the place, where empty collections are created.

openff/interchange/interop/openmm/_import/_import.py
mattwthompson: I think of this as a sort of special case since it's only hit when the entire system lacks bonds and the collections are created elsewhere in dedicated private methods. I agree it shouldn't make a difference

The only problematic case I can think of is a system that constrained angles but not constrained bonds, which is weird and certainly outside what I think of as a normal treatment of rigid water

pbuslaev: Ok, I see you point. I think the PR is good. Thanks a lot for a prompt fix.

pbuslaev (approved):

PR #1137, opened 2025-01-10
Always use `openmm.LangevinMiddleIntegrator`

### LINKED ISSUE #1096: Update integrator used in OpenMM driver
A crucial detail I've run into while getting this ready; [OpenMM's VerletIntegrator](http://docs.openmm.org/7.2.0/api-python/generated/simtk.openmm.openmm.VerletIntegrator.html#simtk.openmm.openmm.VerletIntegrator) (and in general, any leapfrog algorithm integrator in OpenMM) **will report non-zero kinetic energies when evaluating State, _even_ if all velocities are zero**. Taken [straight from the horse's mouth](http://docs.openmm.org/7.2.0/api-python/generated/simtk.openmm.openmm.State.html#simtk.openmm.openmm.State.getKineticEnergy):``
"Note that [KE] may be different from simply mv/2 summed over all particles. For example, a leapfrog integrator will store velocities offset by half a step, so they must be adjusted before computing the kinetic energy." 
``

Some quick testing shows that for static structures exported from Interchange to OpenMM, VerletIntegrator and LangevinIntegrator both yield nonzero KEs, while Langevin<u>Middle</u>Integrator and BrownianIntegrator DO yield the expected 0 KE. I haven't yet checked the [full list of supported Integrators](http://docs.openmm.org/7.2.0/api-python/library.html#integrators) exhaustively.

This becomes relevant to Interchange when noting that the [drivers.openmm layer _does_ use VerletIntegrator](https://github.com/openforcefield/openff-interchange/blob/main/openff/interchange/drivers/openmm.py#L106) in its evaluation, and does not check 0 KE as far as I can tell. This has been introducing errors into my current data, and is likely doing the same for yours @mattwthompson

_Originally posted by @timbernat in https://github.com/openforcefield/openff-interchange/issues/1086#issuecomment-2461076326_

mattwthompson:
### Description

Resolves #1096 and expands it a bit

### Checklist

- [x] Use `openmm.LangevinMiddleIntegrator` in OpenMM driver
- [x] Use `openmm.LangevinMiddleIntegrator` in tests
- [x] Use `openmm.LangevinMiddleIntegrator` in examples
- [x] Lint
- [ ] Update release history

### DIFF
--- a/examples/ligand_in_water/ligand_in_water.ipynb
+++ b/examples/ligand_in_water/ligand_in_water.ipynb
@@ -154,7 +154,7 @@
     "    pdb_stride: int = 500,\n",
     "    trajectory_name: str = \"trajectory.pdb\",\n",
     ") -> openmm.app.Simulation:\n",
-    "    integrator = openmm.LangevinIntegrator(\n",
+    "    integrator = openmm.LangevinMiddleIntegrator(\n",
     "        300 * openmm.unit.kelvin,\n",
     "        1 / openmm.unit.picosecond,\n",
     "        1 * openmm.unit.femtoseconds,\n",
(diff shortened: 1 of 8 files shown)


mattwthompson:
Thanks @Yoshanuikabundi!

PR #1150, opened 2025-01-23
Update experimental tag on examples, tests

mattwthompson:
### Description

With the 0.4.0 release, some methods are no longer "experimental" but many tests were not updated and `INTERCHANGE_EXPERIMENTAL=1` was unnecessarily set.

This change also moves around an example that's no longer "experimental" into the root directory, which is needed for discovery in the docs automation: https://github.com/openforcefield/openff-docs/issues/71

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings


mattwthompson:
As this intersects with how `openff-docs` discovers examples, I think @Yoshanuikabundi is the best person to reason through these changes - particularly the example(s) moving around

j-wags (approved):
This looks good to me. I'm self-assigning and approving since this is such a big blocker, and @Yoshanuikabundi has a hot deadline with the PTM notebook and then BeSMARTS documentation immediately after while we still have Trevor available. If there's a chance that this borks the interface with docs, then let's add a ticket to the backlog to check that once @Yoshanuikabundi has availability again.

mattwthompson:
Thank you!

PR #1153, opened 2025-01-24
Error when non-bonded 1-4 scaling factors differ significantly

### LINKED ISSUE #380: Combination: Components with different 1-4 scaling factors

mattwthompson:
### Description

Closes #380

### Checklist

- [x] Add tests
- [x] Lint
- [ ] Update docstrings

### DIFF
--- a/docs/using/edges.md
+++ b/docs/using/edges.md
@@ -49,7 +49,7 @@ For example, `Interchange.topology.get_positions()` will never include positions
 
 ### Electrostatics 1-4 scaling factors may be slightly modified
 
-Amber family force fields historically use a 1-4 scaling factor of 1 / 1.2 (or 5/6). Some older OpenFF force field releases round this to 6 digits (0.833333) but more recent releases round this to 10 digits (0.8333333333). These are not stricly equal and arguably should not be combined, but (in this case only) for easier compatibility this difference is ignored and combination proceeds as if they were both 0.833333333 from the start. If value differ significantly, i.e. 0.5 vs 0.833333, an error is raised as this difference is non-trivial.
+Amber family force fields historically use a 1-4 scaling factor of 1 / 1.2 (or 5/6). Some older OpenFF force field releases round this to 6 digits (0.833333) but more recent releases round this to 10 digits (0.8333333333). These are not strictly equal and arguably should not be combined, but (in this case only) for easier compatibility this difference is ignored and combination proceeds as if they were both 0.833333333 from the start. If value differ significantly, i.e. 0.5 vs 0.833333, an error is raised as this difference is non-trivial.
 
 ## Quirks of `from_openmm`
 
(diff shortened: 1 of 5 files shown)


openff/interchange/operations/_combine.py
```diff
@@ -44,6 +44,31 @@ def _check_nonbonded_compatibility(
...
+        if sorted([interchange1["Electrostatics"].scale_14, interchange2["Electrostatics"].scale_14]) == [
+            0.833333,
+            0.8333333333,
+        ]:
+            warnings.warn(
+                "Found electrostatics 1-4 scaling factors of 5/6 with slightly different rounding "
+                "(0.833333 and 0.8333333333). This likely stems from OpenFF using more digits in rounding 1/1.2. "
+                "The value of 0.8333333333 will be used, which may or may not introduce small errors. ",
+                InterchangeCombinationWarning,
+            )
```
mattwthompson, line 62: Here's where I'd like to link to upstream https://github.com/openforcefield/openff-forcefields/issues/120

docs/using/edges.md
```diff
@@ -45,6 +45,12 @@ For example, `Interchange.topology.get_positions()` will never include positions
 
 `Molecule` and `Topology` objects can store partial charges, but these are ignored by default in methods like `Interchange.from_smirnoff`. This is because partial charges in SMIRNOFF force fields are defined by sections in the force field. To override this behavior, use the `charge_from_molecules` argument. Be aware that charges, and as a result most physics, will differ from what's perscribed by the contents of the force field.
 
+## Quirks of `Interchange.combine`
+
+### Electrostatics 1-4 scaling factors may be slightly modified
+
+Amber family force fields historically use a 1-4 scaling factor of 1 / 1.2 (or 5/6). Some older OpenFF force field releases round this to 6 digits (0.833333) but more recent releases round this to 10 digits (0.8333333333). These are not stricly equal and arguably should not be combined, but (in this case only) for easier compatibility this difference is ignored and combination proceeds as if they were both 0.833333333 from the start. If value differ significantly, i.e. 0.5 vs 0.833333, an error is raised as this difference is non-trivial.
```
lilyminium, line 52:
```suggestion
Amber family force fields historically use a 1-4 scaling factor of 1 / 1.2 (or 5/6). Some older OpenFF force field releases round this to 6 digits (0.833333) but more recent releases round this to 10 digits (0.8333333333). These are not strictly equal and arguably should not be combined, but (in this case only) for easier compatibility this difference is ignored and combination proceeds as if they were both 0.833333333 from the start. If value differ significantly, i.e. 0.5 vs 0.833333, an error is raised as this difference is non-trivial.
```

I might take some time to think about this but currently this seems ok to me

j-wags (approved):

mattwthompson:
Thanks!

Issue #1166, opened 2025-02-12
Adjust charges written to GROMACS topology


pbuslaev:
**Description**
I am using interchange to write topologies for lipids generated with openmm [`Modeller.addMembrane`](http://docs.openmm.org/latest/api-python/generated/openmm.app.modeller.Modeller.html?highlight=modeller#openmm.app.modeller.Modeller.addMembrane) method.

Overall, my code looks like that:

```python
from openmm.app import Modeller, PME, HBonds, ForceField, PDBFile
from openmm import NonbondedForce
from openmm.unit import nanometer
from openff.interchange import Interchange

for lipid, name in zip(["popc", "pope", "dppc"], ["POP", "POP", "DPP"]):
    print(lipid)
    pdb = PDBFile("memb_protein_trimer.pdb")

    m = Modeller(pdb.topology, pdb.positions)
    ff = ForceField("amber14-all.xml", "amber14/tip3pfb.xml")
    m.addHydrogens(ff)
    attempts_left = 5
    while attempts_left > 0:
        try:
            m.addMembrane(ff, lipidType=lipid.upper())
            break
        except Exception as e:
            print(e)
            attempts_left -= 1

    print(len(list(m.topology.residues())))
    names = []
    for i, r in enumerate(m.topology.residues()):
        if r.name not in names:
            names.append(r.name)
        if r.name == name:
            print(i)
            found = i
            break
    print(names)
    m.delete(list(m.topology.residues())[:found])
    m.delete(list(m.topology.residues())[1:])
    system = ff.createSystem(
        m.topology,
        nonbondedMethod=PME,
        nonbondedCutoff=1.0 * nanometer,
    )
    PDBFile.writeFile(
        m.topology, m.positions, open(f"test_memb_{lipid}.pdb", "w")
    )

    for force in system.getForces():
        if isinstance(force, NonbondedForce):
            n_p = force.getNumParticles()
            print(n_p)
            charges = [
                force.getParticleParameters(i)[0]._value for i in range(n_p)
            ]
            print(charges)
            print(sum(charges))

    i = Interchange.from_openmm(
        system, m.topology, m.positions, system.getDefaultPeriodicBoxVectors()
    )
    off_ch = [
        x.parameters["charge"].m
        for x in i.collections["Electrostatics"].potentials.values()
    ]
    print(off_ch)
    print(sum(off_ch))
    i.to_gromacs(lipid, _merge_atom_types=True)
    break
[...]

PR #1167, opened 2025-02-12
Added simple rounding error solution

pbuslaev:
### Description

This is a proposed solution for #1166 on the GROMACS writer side.

If this approach is approved I can add tests/parameters

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/interop/gromacs/export/_export.py
+++ b/openff/interchange/interop/gromacs/export/_export.py
@@ -222,7 +222,36 @@ class GROMACSWriter(_BaseModel):
         top.write("[ atoms ]\n")
         top.write(";index, atom type, resnum, resname, name, cgnr, charge, mass\n")
 
-        for atom in molecule_type.atoms:
+        def _adjust_charges(charges: numpy.array, tolerance = 8) -> numpy.array:
+            """
+            Adjust charges so that written charge for a molecule type is 0.
+            """
+            rounded_charges = numpy.round(charges, tolerance)
+            total_charge = numpy.round(numpy.sum(charges_to_write), 0)
+            def _rounding_error(_arr, _sum, _tolerance):
+                return numpy.round(numpy.sum(_arr) - _sum, _tolerance)
+
+            rounding_error = _rounding_error(rounded_charges, total_charge, tolerance)
+            
+            if rounding_error != 0:
+                rounded_charges += numpy.round(
+                    -rounding_error / len(charges_to_write),
+                    tolerance
+                )
+                diff = _rounding_error(rounded_charges, total_charge, tolerance)
+                while diff != 0:
+                    # Never adjust 0 charges
+                    indices = numpy.where(rounded_charges != 0)[0]
+                    rounded_charges[indices[0]] = numpy.round(
+                        rounded_charges[indices[0]] - diff, tolerance
+                    )
+                    diff = _rounding_error(rounded_charges, total_charge, tolerance)
+            return rounded_charges
+
+        charges_to_write = numpy.array([atom.charge.m for atom in molecule_type.atoms])
+        rounded_charges = _adjust_charges(charges_to_write)
+        
+        for atom, charge in zip(molecule_type.atoms, rounded_charges):
             if merge_atom_types:
                 top.write(
                     f"{atom.index:6d} "
@@ -231,7 +260,7 @@ class GROMACSWriter(_BaseModel):
                     f"{atom.residue_name:8s} "
                     f"{atom.name:6s}"
                     f"{atom.charge_group_number:6d}"
-                    f"{atom.charge.m:20.12f}"
+                    f"{charge:20.12f}"
                     f"{atom.mass.m:20.12f}\n",
                 )
             else:
@@ -242,7 +271,7 @@ class GROMACSWriter(_BaseModel):
                     f"{atom.residue_name:8s} "
                     f"{atom.name:6s}"
                     f"{atom.charge_group_number:6d}"
-                    f"{atom.charge.m:20.12f}"
+                    f"{charge:20.12f}"
                     f"{atom.mass.m:20.12f}\n",
                 )
 


openff/interchange/interop/gromacs/export/_export.py
```diff
@@ -222,7 +222,38 @@ def _write_atoms(
...
-        for atom in molecule_type.atoms:
+        def _adjust_charges(charges: numpy.array, tolerance=8) -> numpy.array:
+            """
+            Adjust charges so that written charge for a molecule type is 0.
+            """
+            rounded_charges = numpy.round(charges, tolerance)
+            total_charge = numpy.round(numpy.sum(charges_to_write), 0)
+
+            def _rounding_error(_arr, _sum, _tolerance):
+                return numpy.round(numpy.sum(_arr) - _sum, _tolerance)
```
j-wags, line 233: (not blocking) My external perspective here is that needing to come back to this definition repeatedly as this is called in multiple places decreases readability substantially, and that the line length is about the same if `numpy.round` were called directly on each line.

openff/interchange/interop/gromacs/export/_export.py
```diff
@@ -222,7 +222,38 @@ def _write_atoms(
...
+
+            rounding_error = _rounding_error(rounded_charges, total_charge, tolerance)
+
+            if rounding_error != 0:
+                rounded_charges += numpy.round(
+                    -rounding_error / len(charges_to_write),
+                    tolerance,
+                )
```
j-wags, line 241: (blocking) Wouldn't 0 charges be affected here, contradicting the "never adjust 0 charges" comment below?

openff/interchange/interop/gromacs/export/_export.py
```diff
@@ -222,7 +222,38 @@ def _write_atoms(
...
+                diff = _rounding_error(rounded_charges, total_charge, tolerance)
+                while diff != 0:
+                    # Never adjust 0 charges
+                    indices = numpy.where(rounded_charges != 0)[0]
+                    rounded_charges[indices[0]] = numpy.round(
```
j-wags, line 246:
```suggestion
                    # If there's a remainder of a charge after spreading everything out evenly, 
                    # dump the remainder on the first nonzero atom. 
                    rounded_charges[indices[0]] = numpy.round(
```

openff/interchange/interop/gromacs/export/_export.py
```diff
@@ -222,7 +222,38 @@ def _write_atoms(
...
+
+            rounding_error = _rounding_error(rounded_charges, total_charge, tolerance)
+
+            if rounding_error != 0:
+                rounded_charges += numpy.round(
+                    -rounding_error / len(charges_to_write),
+                    tolerance,
+                )
+                diff = _rounding_error(rounded_charges, total_charge, tolerance)
+                while diff != 0:
```
j-wags, line 243: (not blocking) Under what circumstances would this while loop iterate more than once? I can't think of any.

j-wags (changes requested):
Some questions about the implementation here. I think also it could be commented a bit better - it took me a while to grok why some intermediates were getting rounded and others not just by reading the code. This new behavior should also be described in the docs.

openff/interchange/interop/gromacs/export/_export.py
pbuslaev: You are right, should be resolved now.

openff/interchange/interop/gromacs/export/_export.py
pbuslaev: I agree, since tolerance is exactly the same, then one round should be enough.

openff/interchange/components/interchange.py
```diff
@@ -425,12 +439,19 @@ def to_top(
...
+            If False, charges are untouched.
 
         Notes
         -----
         Molecule names in written files are not guaranteed to match the `Moleclue.name` attribute of the
         molecules in the topology, especially if they are empty strings or not unique.
 
+        Charges written to the topology are normalized per molecule to ensure that each molecule has a charge which is
+        exactly integer. Charges that are exactly 0 are not touched.
+
```
mattwthompson, line 454: Since this is not the default behavior and the opt-in would be private here, I think this information should only live in the argument's description

openff/interchange/components/interchange.py
```diff
@@ -353,6 +354,14 @@ def to_gromacs(
...
             The flag to define behaviour of GROMACSWriter. If True, then similar atom types will be merged.
             If False, each atom will have its own atom type.
+        _normalize_charges: bool, default = False
+            If True charges written to the topology are normalized per molecule to ensure that each molecule has a
+            charge which is exactly integer. Charges that are exactly 0 are not touched.
+            If False, charges are untouched.
+        _normalize_charges: bool, default = False
+            If True charges written to the topology are normalized per molecule to ensure that each molecule has a
+            charge which is exactly integer. Charges that are exactly 0 are not touched.
+            If False, charges are untouched.
```
mattwthompson, line 364:
```suggestion
        _normalize_charges: bool, default = False
            If True charges written to the topology are normalized per molecule to ensure that each molecule has a
            charge which is exactly integer. Charges that are exactly 0 are not touched.
            If False, charges are untouched.
```

openff/interchange/interop/gromacs/export/_export.py
```diff
@@ -218,11 +226,51 @@ def _write_atoms(
...
+            """
+            Adjust charges so that written charge for a molecule type is integer.
+
+            We first get the initial charges and round them.
+            Note, that charges that are exactly equal to zero are never touched.
+            """
+            # placeholder for output charges
+            rounded_charges = numpy.round(charges, tolerance)
+            # integer total charge
+            total_charge = numpy.round(numpy.sum(charges), 0)
```
mattwthompson, line 244: I don't think we have formal charges of each atom at this point, right?

mattwthompson (commented):
I see what you're after here and think it's probably a net improvement. I still think most of the cases that would be fixed by this ought to be handled earlier in a workflow, but I see some value in safeguarding against small errors propagating through to GROMACS files on an opt-in basis.

mattwthompson:
@pbuslaev what do you think about trying to fix this problem earlier by "fixing" charges during `Interchange.from_openmm`?

PR #1186, opened 2025-03-11
Do not assume molecules are grouped

mattwthompson:
### Description

Resolves #1185 

### Checklist

- [x] Add tests
- [x] Lint
- [x] Update docstrings

PR #1187, opened 2025-03-11
Add example loading and comparing energy evaluations of prepared SAMPL6 system

mattwthompson:
### Description

Needs #1186 to get acceptable results

### DIFF
--- a/examples/load-SAMPL6/README.md
+++ b/examples/load-SAMPL6/README.md
@@ -1,0 +1,3 @@
+# Load a prepared SAMPL6 system
+
+[Source](https://github.com/samplchallenges/SAMPL6/tree/c661d3985af7fa0ba8c64a1774cfb2363cd31bda/host_guest/SAMPLing)
--- a/examples/load-SAMPL6/load-SAMPL6.ipynb
+++ b/examples/load-SAMPL6/load-SAMPL6.ipynb
@@ -1,0 +1,98 @@
+{
+ "cells": [
+  {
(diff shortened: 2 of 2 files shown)


Yoshanuikabundi (approved):
This is great, short and simple and works.

If you're looking for opportunities for improvement, I'd suggest:

- Introducing the notebook as "This notebook downloads a SAMPL6 challenge as an XML-serialized OpenMM `System`, creates an Interchange from it, and then computes the energy of the starting coordinates with GROMACS."
- Add a comment explaining why forces are removed from the System before import; something like

```
# Remove unsupported force objects that do not affect the energy
```

- Compute the energy for the deserialized system as well, especially if this can be done in one or two lines with the same tabular format, even if that means using a private function
- Maybe a comment on the lines setting the residue number metadata? It wasn't clear to me what that's for, and the notebook is so lean that it feels important.

I also found that attempting to compute the energy in Amber by removing the `_engines` argument or adding `"Amber"` to the list is taking a very long time - I gave up at 12 minutes. Is that expected? I don't think it's necessarily an issue but it is surprising.

mattwthompson:
Thanks @Yoshanuikabundi! This feedback is super useful. I can see with fresher eyes what is missing with some of the comments you've suggested.

I'm passing on evaluating the serialized representation of the system (I assume this is what you mean) because it's not trivially simple to do with OpenMM. If you meant with GROMACS that's pretty straightforward

Issue #544, opened 2022-10-21
Potential error in LAMMPS improper atom order


benjamindjensen:
**Description**

Interchange appears to be defining impropers with the central atom in the second position in LAMMPS output files.

The expected atom order for impropers in LAMMPS with `improper_style cvff` is I,J,K,L where the improper dihedral angle is between the plane of I,J,K and the plane of J,K,L (https://docs.lammps.org/improper_cvff.html). The central atom in the improper is the first atom I. See the note in the documentation: _Note that defining 4 atoms to interact in this way, does not mean that bonds necessarily exist between I-J, J-K, or K-L, as they would in a linear dihedral. Normally, the bonds I-J, I-K, I-L would exist for an improper to be defined between the 4 atoms._

**Reproduction**

```
molecule = Molecule.from_smiles('C1=CN=CN1')
sage = ForceField("openff_unconstrained-2.0.0.offxml")
molecule.generate_conformers()
cubic_box = (
    np.array(
        [
            [10.0, 0.0, 0.0],
            [0.0, 10.0, 0.0],
            [0.0, 0.0, 10.0],
        ]
    )
    * unit.angstrom
)

interchange = Interchange.from_smirnoff(topology=[molecule], force_field=sage, box=cubic_box)
interchange.to_lammps("imidazole.lmp")
```

**Output**

```
Impropers

1	3	5	4	8	3
2	2	1	5	4	9
3	1	6	1	2	5
6	1	3	2	7	1
8	1	1	2	3	7
10	3	3	4	5	8
12	2	9	5	1	4
15	2	4	5	9	1
16	1	2	1	5	6
20	1	7	2	1	3
23	1	5	1	6	2
24	3	8	4	3	5
```

![image](https://user-images.githubusercontent.com/64032677/197202566-187b3a64-77fe-4bb6-80dd-11295cc1a47e.png)

**Software versions**
From `conda list`:
`openff-interchange        0.2.1              pyhd8ed1ab_0    conda-forge`

mattwthompson:
[...]

PR #1195, opened 2025-04-04
LAMMPS improper atom order

benjamindjensen:
### Description

The order of atoms for impropers in LAMMPS is not correct. This wasn't showing up in the existing unit tests because the impropers were at their equilibrium configuration for the tested structures and were therefore zero energy regardless of the order. The new unit test performs a random perturbation of the atomic coordinates and compares the energies to the unperturbed molecule. An incorrect atom order for impropoers therefore fails the test. The bond/angle/torsion/etc. energy components are also compared between pristine and perturbed systems.

### Checklist

- [X] Add tests
- [X] Lint
- [ ] Update docstrings


j-wags:
Thanks for the high-quality PR, @benjamindjensen! @mattwthompson's the owner of this repo and he's off today, so he'll review this once he's back next week.

mattwthompson:
Beautiful, thanks @benjamindjensen! I have some notes below which I'm adding as notes for double-check my understanding. I think this is all correct, but improper torsion atom order is my achilles' heel (going back the better part of a decade now[^1]) so I want to be sure.

For this `improper_style cvff` (but not [all](https://matsci.org/t/improper-definition-in-lammps/33754/2), the central atom should be listed first:

> If the 4 atoms in an improper quadruplet (listed in the data file read by the [read_data](https://docs.lammps.org/read_data.html) command) are ordered I,J,K,L then the improper dihedral angle is between the plane of I,J,K and the plane of J,K,L.

https://docs.lammps.org/improper_cvff.html

I spun up a triazine to see how Interchange store this, since I can never remember:

```python
In [1]: from openff.toolkit import Molecule, ForceField

In [2]: molecule = Molecule.from_smiles("n1cncnc1")

In [3]: sage = ForceField("openff-2.1.1.offxml")
/Users/mattthompson/micromamba/envs/yammbs-dev/lib/python3.12/site-packages/smirnoff99frosst/smirnoff99frosst.py:11: UserWarning: Module openff was already imported from None, but /Users/mattthompson/software/openff-interchange is being added to sys.path
  from pkg_resources import resource_filename

In [4]: [*sage.create_interchange(molecule.to_topology())['ImproperTorsions'].key_map.keys()]
Out[4]:
[ImproperTorsionKey with atom indices (1, 0, 2, 6), mult 0,
 ImproperTorsionKey with atom indices (1, 2, 6, 0), mult 0,
 ImproperTorsionKey with atom indices (1, 6, 0, 2), mult 0,
 ImproperTorsionKey with atom indices (5, 0, 4, 8), mult 0,
 ImproperTorsionKey with atom indices (5, 4, 8, 0), mult 0,
 ImproperTorsionKey with atom indices (5, 8, 0, 4), mult 0,
 ImproperTorsionKey with atom indices (3, 2, 4, 7), mult 0,
 ImproperTorsionKey with atom indices (3, 4, 7, 2), mult 0,
 ImproperTorsionKey with atom indices (3, 7, 2, 4), mult 0]

```

This all seems good and well, except for [this inconsistency](https://github.com/openforcefield/openff-interchange/blob/v0.4.2/openff/interchange/models.py#L269-L271):
```python
In [5]: [key.get_central_atom_index() for key in sage.create_interchange(molecule.to_topology())
      ⋮ ['ImproperTorsions'].key_map.keys()]
Out[5]: [0, 2, 6, 0, 4, 8, 2, 4, 7]
```

So I think something's up elsewhere (which is well outside your responsibility, of course) which I'll need to dig into further. Your approach for testing this is great - impropers energies are quite easy to ignore - so I will likely be stealing it for testing other engines.

[^1]: You're already in the canon here, having fixed this issue once before (https://github.com/openforcefield/openff-interchange/issues/544). I'm not sure how this managed to regress, but the test you've added should safeguard it from happening again

benjamindjensen:
@mattwthompson Thank you! I was interested in validating the LAMMPS conversion and gaining some familiarity with how the unit tests were setup. The good news to me was that bond/angle/dihedral/vdw all seemed good to me. I'm not sure how critical impropers are but the various conventions for them across software packages mades them less straight forward. Let me know if there's anything I can do to help.

benjamindjensen:
Is this where the inconsistency is coming from?

https://github.com/openforcefield/openff-interchange/blob/ed5e38b3b1787a77c673bffca4a646037feb90eb/openff/interchange/smirnoff/_valence.py#L685-L698

In line 696 the central atom `key[1]` is assigned to the first atom index. 

This matches the unit test for formaldehyde where the central atom in the improper is listed first and not second:

https://github.com/openforcefield/openff-interchange/blob/ed5e38b3b1787a77c673bffca4a646037feb90eb/openff/interchange/_tests/unit_tests/smirnoff/test_valence.py#L100-L106

mattwthompson:
What tripped me up is that SMIRNOFF lists the central atom second: 

> The second atom in an improper (in the example above, the trivalent carbon) is the central atom in the trefoil.

https://openforcefield.github.io/standards/standards/smirnoff/

But for reasons that escape me, I decided that Interchange's internal representation stores the central atom first. Clearly I forgot this at some point, observed both with the LAMMPS code you found and `get_central_atom_index`.

Inconsistent between the two, yes, but I'm inclined to leave it in place since there is no standard across engines.

PR #1203, opened 2025-04-17
Remove hybrid styles from LAMMPS export

benjamindjensen:
### Description

One of the changes identified in #955 is removing hybrid styles from the LAMMPS export. Hybrid styles are not currently required for OpenFF potentials. Including hybrid LAMMPS potentials in a data file is inconvenient for users because LAMMPS will not write hybrid coefficients to restart/data files for subsequent simulations, requiring a user to manually implement the coefficients in the LAMMPS script instead of the data file. If OpenFF requires hybrid styles in the future (using both LJ and Buckingham potentials for example) I would recommend providing the force field coefficients in the LAMMPS script instead of the data file.

### Checklist

- [ ] Add tests
- [ X] Lint
- [ ] Update docstrings


mattwthompson:
Heh, I forget exactly why I went with "hybrid" potentials but at the time I thought it was the only way to get some detail correct. I'd be happy to be wrong - I never used them when I was running simulations myself.

Could you drop the `openeye: true` lines from the workflows and see if they run?

benjamindjensen:
I'm not particularly familiar with github actions. I modified the ci.yaml file and it seem to have passed most of the tests, although a few are failing from what looks like an openeye license error. Let me know if there's a more appropriate way to modify the workflow for this pull.

I've also not used hybrid potentials in lammps, but have been using this modification myself for a while now and thought it would be helpful for others to have in the official repository.

mattwthompson:
@benjamindjensen what version of LAMMPS are you using? What version(s) do you expect this to work with?

mattwthompson:
Nevermind, I was making an error elsewhere in my tooling.

I've grabbed these changes locally and everything looks good. I'm opening this in a separate PR for boring reasons relating to our automation, but I think it's good to go.

I assume that @mrshirts and @timbernat are okay with this change?

benjamindjensen:
> Nevermind, I was making an error elsewhere in my tooling.
> 
> I've grabbed these changes locally and everything looks good. I'm opening this in a separate PR for boring reasons relating to our automation, but I think it's good to go.
> 
> I assume that @mrshirts and @timbernat are okay with this change?

Great! Thank you for including this in Interchange.

PR #1219, opened 2025-05-13
Add simple charge array getter

### LINKED ISSUE #1188: User friendly way to get an array of charges
**Description**
Please describe the behavior you would like added to Interchange.

Would it be possible to add a user-friendly property or method to get an ordered array of charges? Right now as far as I'm aware the only way to do so is below, and it's a bit of a mouthful. I think the charges are usually ordered by particle but the existence of #970 makes me want to be sure.

```ipython
>>> charges = [
        q for k, q in sorted(
            interchange.collections["Electrostatics"].charges.items(),
            key=lambda x: x[0].atom_indices
        )
    ]
>>> charges
[<Quantity(-0.108680002, 'elementary_charge')>, <Quantity(0.0271700006, 'elementary_charge')>, <Quantity(0.0271700006, 'elementary_charge')>, <Quantity(0.0271700006, 'elementary_charge')>, <Quantity(0.0271700006, 'elementary_charge')>]
>>> charges_without_pint = [q.m for q in charges]
>>> charges_without_pint
[-0.10868000239133835, 0.027170000597834587, 0.027170000597834587, 0.027170000597834587, 0.027170000597834587]
```

(I know the ordering might be complicated for virtual sites.)

mattwthompson:
### Description

May resolve #1188 

I decided to, in a first pass
* Keep this method private
* Not support virtual sites
* Return an array, not a `Quantity`
* Test only the most common SMIRNOFF charge assignment methods and nothing non-SMIRNOFF

### Checklist

- [x] Add tests
- [x] Lint
- [ ] Update docstrings (won't do)

### DIFF
--- a/openff/interchange/_tests/unit_tests/smirnoff/test_nonbonded.py
+++ b/openff/interchange/_tests/unit_tests/smirnoff/test_nonbonded.py
@@ -225,6 +225,66 @@ class TestElectrostatics:
         compare_charges(original, get_charges_from_interchange(original))
         compare_charges(reordered, get_charges_from_interchange(reordered))
 
+    def test_get_charge_array(self, sage):
+        ammonia = Molecule.from_smiles("N")
+        ammonia.partial_charges = Quantity(numpy.array([-3, 1, 1, 1]) / 3, "elementary_charge")
+
+        # topology is water | ethanol         | ammonia
+        #              O H H C C O H H H H H H N H H H
+        topology = Topology.from_molecules(
+            [
+                Molecule.from_mapped_smiles("[H:2][O:1][H:3]"),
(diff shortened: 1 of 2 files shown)


mattwthompson:
I've updated this to

* return a `Quantity`
* be a public method

It's plausible that the simpler array representation is better for `.charges`, but that would be a sweeping change with broad consequences. Even though it's a little tedious to opt in to `.get_charge_array`, that provides a much safer ground for seeing if we want to make the behavior change at some point in the future while enabling the handy charge-getting that you originally were after

lilyminium (approved):
LGTM, thanks so much @mattwthompson!

PR #1224, opened 2025-05-20
Improve errors when parsing GROMACS files

mattwthompson:
### Description

This PR raises a more descriptive error in some failure modes of parsing GROMACS topology files. See #1212; where previously there was an `UnboundLocalError`.

Minimal demo:

```python
In [1]: from openff.interchange import Interchange

In [2]: import os

In [3]: os.environ["INTERCHANGE_EXPERIMENTAL"]="1"

In [4]: !echo "blah BLah blAh blAH" > x.top

In [5]: !touch x.gro

In [6]: Interchange.from_gromacs('x.top', 'x.gro')
---------------------------------------------------------------------------
GROMACSParseError                         Traceback (most recent call last)
Cell In[6], line 1
----> 1 Interchange.from_gromacs('x.top', 'x.gro')

File ~/software/openff-interchange/openff/interchange/_experimental.py:35, in experimental.<locals>.wrapper(*args, **kwargs)
     26 if os.environ.get("INTERCHANGE_EXPERIMENTAL", "0") != "1":
     27     raise ExperimentalFeatureException(
     28         f"\n\tFunction or method {func.__name__} is experimental. This feature is not "
     29         "complete, not yet reliable, and/or needs more testing to be considered suitable "
   (...)     32         "INTERCHANGE_EXPERIMENTAL=1.",
     33     )
---> 35 return func(*args, **kwargs)

File ~/software/openff-interchange/openff/interchange/components/interchange.py:900, in Interchange.from_gromacs(cls, topology_file, gro_file)
    896 from openff.interchange.interop.gromacs._import._import import from_files
    897 from openff.interchange.interop.gromacs._interchange import to_interchange
    899 return to_interchange(
--> 900     from_files(
    901         top_file=topology_file,
    902         gro_file=gro_file,
    903     ),
    904 )

File ~/software/openff-interchange/openff/interchange/_experimental.py:35, in experimental.<locals>.wrapper(*args, **kwargs)
     26 if os.environ.get("INTERCHANGE_EXPERIMENTAL", "0") != "1":
     27     raise ExperimentalFeatureException(
     28         f"\n\tFunction or method {func.__name__} is experimental. This feature is not "
     29         "complete, not yet reliable, and/or needs more testing to be considered suitable "
   (...)     32         "INTERCHANGE_EXPERIMENTAL=1.",
     33     )
---> 35 return func(*args, **kwargs)

[...]

PR #1225, opened 2025-05-20
Treat asterisks as comments in GROMACS topology files

### LINKED ISSUE #1223: Asterisks not processed as comments in GROMACS topology files
**Description**

I've always used (and seen) semicolons used to denote comments in GROMACS topology files. This is consistent with [what is documented](https://manual.gromacs.org/current/reference-manual/topologies/topology-file-formats.html#topology-file):

> Description of the file layout:
> Semicolon (;) and newline characters surround comments

It turns out, however, that asterisks can also be used. This causes `Interchange.from_gromacs` to fail because it doesn't know what to do with them.

See, for example: https://github.com/gromacs/gromacs/blob/cefa450b53c9e9506fa4b772c8932f8a6698373a/share/top/amber96.ff/forcefield.itp#L2

mattwthompson:
### Description

Resolves #1223 

Easiest to move forward if this follows #1224, but could be split out

### Checklist

- [x] Add tests
- [x] Lint
- [ ] Update docstrings


j-wags (approved):

j-wags (changes requested):
Oh, actually @lilyminium just nailed it with a possible edge case -  could you add a test of parsing [this file,](https://github.com/gromacs/gromacs/blob/cefa450b53c9e9506fa4b772c8932f8a6698373a/share/top/oplsaa.ff/ffbonded.itp#L8) where some atom names have asterisks in them? This will prevent someone from introducing a regression in the future.

mattwthompson:
That particular file uses a directive that we don't support (yet), but I've modified a existing file to make sure that atom types with trailing asterisks aren't discarded

j-wags (approved):
Thanks!

Issue #1268, opened 2025-07-22
`solvate_topology` does not neutralize


IAlibay:
**Description**

`solvate_topology` should neutralize a system with a negatively charged solute. However, it yields a positively charged system.

I manuallly checked and under the reproduction conditions below, `solvate_topology` attempts to add:
  - 1 negatively charged acetic acid  
  - 4 Na+
  - 2 Cl-
  - 1718 waters

**Reproduction**

```python
from openff.units import unit
from openff.toolkit import Molecule, Topology, ForceField
from openff.interchange import Interchange
from openff.interchange.components._packmol import solvate_topology

solute = Molecule.from_smiles("CC(=O)[O-]")
solute.generate_conformers(n_conformers=1)
solute.assign_partial_charges(partial_charge_method='gasteiger')
print("solute charge: ", sum(solute.partial_charges))

top = solvate_topology(Topology.from_molecules([solute]), nacl_conc=0.1 * unit.molar, padding=2 * unit.nm)
ff = ForceField("openff-2.0.0.offxml")
inter = Interchange.from_smirnoff(ff, top)

print("system charge:", sum(inter['Electrostatics'].get_charge_array()))
```

**Output**

```
solute charge:  -0.9999999999999999 elementary_charge
system charge: 1.0 elementary_charge
```

**Software versions**

Interchange v0.4.3 and openff-toolkit 0.16.9

mattwthompson:
Good find, I think this is a quirk of rounding which is un-checked?

```python
ipdb> solute_charge = Quantity("-1 elementary_charge")
ipdb> nacl_to_add = 3
ipdb> nacl_to_add - solute_charge.m / 2.0, nacl_to_add + solute_charge.m / 2.0
(3.5, 2.5)
ipdb>  numpy.ceil(nacl_to_add - solute_charge.m / 2.0), numpy.floor(nacl_to_add + solute_charge.m / 2.0)
(4.0, 2.0)
```

floor-ing or ceil-ing both would produce sensible solutions, I think

```python
ipdb>  numpy.floor(nacl_to_add - solute_charge.m / 2.0), numpy.floor(nacl_to_add + solute_charge.m / 2.0)
(3.0, 2.0)
ipdb>  numpy.ceil(nacl_to_add - solute_charge.m / 2.0), numpy.ceil(nacl_to_add + solute_charge.m / 2.0)
(4.0, 3.0)

mattwthompson:
[...]

PR #1269, opened 2025-07-22
Ensure result of `solvate_topology` is neutral

mattwthompson:
### Description

This starts fixing #1268

### Checklist

- [x] Raise error when topology is (probably?) non-neutral
- [ ] More consistently round numbers of ions?
- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/components/_packmol.py
+++ b/openff/interchange/components/_packmol.py
@@ -909,7 +909,12 @@ def solvate_topology(
     na_to_add = numpy.ceil(nacl_to_add - solute_charge.m / 2.0)
     cl_to_add = numpy.floor(nacl_to_add + solute_charge.m / 2.0)
 
-    # Pack the box
+    if abs(solute_charge.m + na_to_add - cl_to_add) > 1e-6:
+        raise PACKMOLValueError(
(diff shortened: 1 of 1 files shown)


openff/interchange/components/_packmol.py
```diff
@@ -900,16 +900,29 @@ def solvate_topology(
...
+
     # Compute the number of water molecules to add to make up the remaining mass
     water_mass_to_add = solvent_mass - nacl_mass
     water_to_add = (water_mass_to_add / water_mass).m_as("dimensionless").round()
 
-    # Neutralise the system by adding and removing salt
+    # Neutralise the system by **adding** the counter-ion, which we think OpenMM does
     solute_charge = sum([molecule.total_charge for molecule in topology.molecules])
-    na_to_add = numpy.ceil(nacl_to_add - solute_charge.m / 2.0)
-    cl_to_add = numpy.floor(nacl_to_add + solute_charge.m / 2.0)
```
mattwthompson, line 910: Another solution might be to simply make these _both_ round up or down, compared to what it currently does which is round to even numbers

PR #1270, opened 2025-07-22
Always round up number of ions to add when neutralizing topology

mattwthompson:
### Description

Another possible fix for #1268, this possibly being more elegant

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/components/_packmol.py
+++ b/openff/interchange/components/_packmol.py
@@ -912,7 +912,7 @@ def solvate_topology(
     if abs(solute_charge.m + na_to_add - cl_to_add) > 1e-6:
         raise PACKMOLValueError(
             f"Failed to neutralise solute with charge {solute_charge.m}; meant to add {nacl_to_add} NaCl.\n"
-            f"Tried adding {na_to_add} Na+ and {cl_to_add} Cl- ions.",
+            f"Tried adding {na_to_add} Na+ and {cl_to_add} Cl- ions. This should not happen, please raise an issue!",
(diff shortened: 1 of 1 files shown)


openff/interchange/components/_packmol.py
```diff
@@ -907,9 +907,14 @@ def solvate_topology(
     # Neutralise the system by adding and removing salt
     solute_charge = sum([molecule.total_charge for molecule in topology.molecules])
     na_to_add = numpy.ceil(nacl_to_add - solute_charge.m / 2.0)
-    cl_to_add = numpy.floor(nacl_to_add + solute_charge.m / 2.0)
+    cl_to_add = numpy.ceil(nacl_to_add + solute_charge.m / 2.0)
```
IAlibay, line 910: This method falls apart when `nacl_to_add` is low and the solute charge is high.

Take the case of a heparan sulfate dissacharide which has a net charge of -4 (it's a small system so the rest of what I saw is likely). At low `nacl_conc` you can imagine a case where `nacl_to_add` == 1.

In that case:

`na_to_add = np.ceil(1 - -4 /2) == 3`
`cl_to_add = np.ceil(1 + -4/2) == -1`

You can't have negative ions, so the whole thing breaks down.

Now this example is a bit extreme, but it actually happens quite a lot when `nacl_conc == 0 * unit.molar`, which is a common simulation case, i.e. "I just want to neutralize my system".

Issue #1289, opened 2025-08-10
`padding` description of `solvate_topology` and `solvate_topology_nonwater` seems incorrect.


IAlibay:
`padding` is described as: " Ignored if the topology already has box vectors."

However, there doesn't seem to be anything in the code that actually does this and padding is used directly when calculating `image_distance` and `box_bectors`, e.g.: https://github.com/openforcefield/openff-interchange/blob/main/openff/interchange/components/_packmol.py#L1024

Yoshanuikabundi:
I agree, I think this is incorrect in both functions. I think the documented behavior might be better, as it lets the user specify the exact box vectors and still take advantage of the other conveniences of this method, but it could also be confusing if the user doesn't realise their topology has box vectors.

mattwthompson:
What's the path forward here? Update the docstring, implement the described behavior, or something else?

j-wags:
Since we are actively debugging the new Pontibus protocol that uses/used this method, I think we should avoid changing behavior. You have the final call on this, but in the stand-up we discussed having Josh open a PR with doc fixes.

IAlibay:
> new Pontibus protocol that uses/used this method,

Just FYI - pontibus no longer uses these methods. We use pack_box directly, so changes here won't affect things.

PR #1299, opened 2025-08-14
Implement offending sentence from padding parameter docstring (and test)

Yoshanuikabundi:
### Description

Possible fix for #1289, alternative to #1298

### Checklist

- [x] Add tests
- [x] Lint
- [x] Update docstrings


mattwthompson:
Thanks for your work on this - it's high-quality as always - but as we discussed, we found a different solution to go with

PR #1302, opened 2025-08-18
Make `padding` argument optional, avoid over-specifying inputs

### LINKED ISSUE #1289: `padding` description of `solvate_topology` and `solvate_topology_nonwater` seems incorrect.

mattwthompson:
### Description

Resolves #1289

### Checklist

- [x] Make `padding` argument optional
- [x] Throw error when box vectors and padding distance are specified
- [x] Add tests
- [x] Lint
- [x] Update docstrings
- [ ] Update release history

### DIFF
--- a/openff/interchange/_tests/unit_tests/components/test_packmol.py
+++ b/openff/interchange/_tests/unit_tests/components/test_packmol.py
@@ -509,49 +509,25 @@ class TestPackmolWrapper:
 
         assert topology.n_molecules == 11
 
-    @pytest.mark.slow
-    @pytest.mark.parametrize("solvation_function", ["solvate_topology", "solvate_topology_nonwater"])
-    @pytest.mark.parametrize("n_monomers", [1, 3, 5])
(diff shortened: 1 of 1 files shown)


Yoshanuikabundi (approved):
Looks great! The test appears correct, the implementation appears correct, and my independent tests give the desired behavior. I might suggest (non-blocking) adding an example of how to remove the box from the topology in the docstrings, just in case someone loads a PDB file with the crystallographic box and wants to solvate it.

mattwthompson:
Thanks!

I tried [minimally documenting](https://github.com/openforcefield/openff-interchange/pull/1302/files#diff-096240c43cd443b7a0f7aebcf6c7bb1a17929eb474a70e29ee89ab9f78fee1e3R1071) how to do that, but if it comes up with users we can expand on that

PR #1308, opened 2025-08-21
Fix environments

mattwthompson:
### Description

There's some weird things going on in this step: https://github.com/openforcefield/openff-interchange/blob/main/.github/workflows/ci.yaml#L70-L73


j-wags:
In case it helps, the one big recent packaging change that comes to mind was including openff-nagl as a dependency of openff-toolkit. I poked around and can't tell what the root issue/smoke was that you observed in CI, but if you can post it here I'll take a look and see if I have ideas.

mattwthompson:
So far I'm unable to find a smoking gun - fresh environments get me some charge differences, but almost all < 0.001 - so I'm currently chalking it up to a ghost in the machine. There's cruft in these environments that's a hassle to hunt down, so I'm trying to nudge things towards recent versions.

NAGL's effect on the dependency tree, as far as I can tell now, is not much cause for concern

Issue #1310, opened 2025-08-22
DoubleExponentialVirtualSites are Incorrectly Handled


fjclark:
**Description**

I would like to use the latest double exponential force field from https://github.com/jthorton/de-forcefields/tree/main. However, `interchange > 0.4` does not work with `DoubleExponentialVirtualSites`, as shown by [the failing test in `smirnoff-plugins`](https://github.com/openforcefield/smirnoff-plugins/blob/8b88910b2794737565d209da1d6fbe87533dd067/smirnoff_plugins/_tests/handlers/test_nonbonded.py#L957) and the failing CI on https://github.com/jthorton/de-forcefields/actions?page=2.

**Reproduction**

Run the [the failing test in `smirnoff-plugins`](https://github.com/openforcefield/smirnoff-plugins/blob/8b88910b2794737565d209da1d6fbe87533dd067/smirnoff_plugins/_tests/handlers/test_nonbonded.py#L957).

**Output**

```
../../../miniforge3/envs/deforce/lib/python3.12/site-packages/openff/interchange/components/interchange.py:160: in from_smirnoff
    return _create_interchange(
../../../miniforge3/envs/deforce/lib/python3.12/site-packages/openff/interchange/smirnoff/_create.py:196: in _create_interchange
    _plugins(interchange, force_field, interchange.topology)
../../../miniforge3/envs/deforce/lib/python3.12/site-packages/openff/interchange/smirnoff/_create.py:464: in _plugins
    vdw_collection=interchange[tagnames[0]],
                   ^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = Interchange with 3 collections, periodic topology with 6 atoms., item = 'DoubleExponentialVirtualSites'

    def __getitem__(self, item: str):
        """Syntax sugar for looking up collections or other components."""
        if type(item) is not str:
            raise LookupError(
                f"Only str arguments can be currently be used for lookups.\nFound item {item} of type {type(item)}",
            )
        if item == "positions":
            return self.positions
[...]

PR #1312, opened 2025-08-25
Fix handling of DoubleExponentialVirtualSites

fjclark:
### Description

This is a first pass which attempts to close https://github.com/openforcefield/openff-interchange/issues/1310, and which only works in combination with a PR of the same name to be opened in https://github.com/openforcefield/smirnoff-plugins/tree/main.

I've ensured that the "Fix Me!" test [here](https://github.com/openforcefield/smirnoff-plugins/blob/8b88910b2794737565d209da1d6fbe87533dd067/smirnoff_plugins/_tests/handlers/test_nonbonded.py#L957) now passes, as do the tests in https://github.com/jthorton/de-forcefields/tree/main.

Replying to https://github.com/openforcefield/openff-interchange/issues/1310#issuecomment-3214721469

> This seems fine to me. My only note is that I'd rather not add a new class method, but this is just on a plugin with a small user base so it's fine.

This is what I've gone for, but I'm very happy to change this if you have any other suggestions.

> I'd be happy to hear any suggestions you have about how to make this process easier! It's more cumbersome than it should be

Thanks, but I'm afraid I don't have any great ideas. I'll keep thinking!

### Checklist

- [ ] Add tests
- [x] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/smirnoff/_create.py
+++ b/openff/interchange/smirnoff/_create.py
@@ -440,13 +440,9 @@ def _build_collection(
 
     kwargs["topology"] = topology
 
-    try:
-        from smirnoff_plugins.collections.vsites import _VsitePlugin  # type: ignore[import-not-found]
-    except ImportError:
-        _VsitePlugin = None
-
-    # VirtualSite collections need vdW + Electrostatics
-    if _VsitePlugin is not None and issubclass(collection_class, _VsitePlugin):
+    # VirtualSite collections need vdW + Electrostatics collections, and we can identify
+    # them by the allowed_vdw_parameter_handlers class method.
+    if hasattr(collection_class, "allowed_vdw_parameter_handlers"):
         vdw_tagnames = [x._TAGNAME for x in collection_class.allowed_vdw_parameter_handlers()]  # type: ignore[attr-defined]
 
         if (n_vdw_tagnames := len(vdw_tagnames)) != 1:


openff/interchange/smirnoff/_create.py
```diff
@@ -423,6 +423,43 @@ def _virtual_sites(
...
+        kwargs["parameter_handler"] = force_field[handler_classes[0]._TAGNAME]
+    else:
+        kwargs["parameter_handler"] = [force_field[cls._TAGNAME] for cls in handler_classes]
+
+    kwargs["topology"] = topology
+
+    try:
+        from smirnoff_plugins.collections.vsites import _VsitePlugin  # type: ignore[import-not-found]
+    except ImportError:
+        _VsitePlugin = None
```
mattwthompson, line 446: Is there any way to avoid a dependency loop here?

mattwthompson: Maybe `hasattr(collection, "allowed_vdw_parameter_handlers")`?

mattwthompson: Either seems to work okay in terms of code, I think the attribute checking is a little cleaner but I can be talked out of it.

It's just that A <-> B packaging dependencies are something I try to avoid wherever possible, even though in terms of pure code the import-checking is a fine solution

mattwthompson (approved):

Issue #1300, opened 2025-08-14
Improve `Interchange.combine` tests


mattwthompson:
**Description**
Given separate force fields with separate (intended-for) ligand, protein, and water parameters, parametrize separately and combine,
- [ ] Check that each component is given parameters from the expected source
- [ ] Check atom ordering in the topology
- [ ] Look into performance, with low standards

PR #1327, opened 2025-09-16
Test charge cache after `Interchange.combine`

mattwthompson:
### Description

This was discovered in #1325 and touches on #1300

There is a subtle behavior change in that the input `Interchange`s' charge caches are not invalidated, but I don't see a reason to do that.

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/_tests/unit_tests/operations/test_combine.py
+++ b/openff/interchange/_tests/unit_tests/operations/test_combine.py
@@ -258,6 +258,7 @@ class TestCombine:
 
         assert int3["Electrostatics"]._charges_cached is False
 
+        # Check that the input caches are intact
         for interchange_ in (int1, int2):
             assert len(interchange_["Electrostatics"]._get_charges()) == interchange_.topology.n_atoms
 


openff/interchange/_tests/unit_tests/operations/test_combine.py
```diff
@@ -219,3 +219,21 @@ def test_constraint_key_collision(self, parsley, sage, ethanol, flip_order, hand
...
+
+        for interchange_ in (int1, int2):
+            assert len(interchange_["Electrostatics"]._get_charges()) == interchange_.topology.n_atoms
+
+        if flip_order:
+            int3 = int2.combine(int1)
+        else:
+            int3 = int1.combine(int2)
+
+        assert int3["Electrostatics"]._charges_cached is False
```
Yoshanuikabundi, line 236: Huh, cool! Didn't know `foo is False` was correct.

openff/interchange/_tests/unit_tests/operations/test_combine.py
```diff
@@ -219,3 +219,21 @@ def test_constraint_key_collision(self, parsley, sage, ethanol, flip_order, hand
...
+
+        for interchange_ in (int1, int2):
+            assert len(interchange_["Electrostatics"]._get_charges()) == interchange_.topology.n_atoms
```
Yoshanuikabundi, line 239: NON-BLOCKING - Feel free not to include this, but if I'm understanding correctly I think a short comment to clarify why `int3` is not included in this check could be helpful for future readers. 

```suggestion 
        # Check that the input caches are intact
        for interchange_ in (int1, int2):
            assert len(interchange_["Electrostatics"]._get_charges()) == interchange_.topology.n_atoms
```

Yoshanuikabundi (approved):
Looks great! I appreciate the change towards not modifying the input interchanges. Also didn't know the booleans could be used correctly with `is`! I have a non-blocking suggested code comment that you should feel free to treat as nothing more than a test of my understanding.

mattwthompson:
Thanks again @Yoshanuikabundi!

PR #1328, opened 2025-09-16
Skip nonintegral charge logic when not needed

mattwthompson:
### Description

There are some wasteful $O(N_{atoms})$ loops done only to check `allow_nonintegral_charges`, which is typically turned on but does not necessarily need to be.

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/smirnoff/_nonbonded.py
+++ b/openff/interchange/smirnoff/_nonbonded.py
@@ -503,6 +503,7 @@ class SMIRNOFFElectrostaticsCollection(ElectrostaticsCollection, SMIRNOFFCollect
             molecules_with_preset_charges=molecules_with_preset_charges,
             allow_nonintegral_charges=allow_nonintegral_charges,
         )
+
         handler._charges = dict()
 
         return handler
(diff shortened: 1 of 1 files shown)


openff/interchange/smirnoff/_nonbonded.py
```diff
@@ -1014,41 +1015,35 @@ def store_matches(
...
                         f"Molecule {molecule.to_smiles(explicit_hydrogens=False)} has "
                         f"a net charge of {charge_sum} compared to a total formal charge of "
                         f"{formal_sum}.",
                     )
 
-            molecule.partial_charges = Quantity(
-                molecule_charges,
-                unit.elementary_charge,
-            )
-
```
mattwthompson, line 1051: I don't understand what this assignment is doing. I suspect it's wasteful and simply lingering from some sweeping change

Yoshanuikabundi: Perhaps it should go in `Interchange.to_topology()`!

openff/interchange/smirnoff/_nonbonded.py
```diff
@@ -1014,41 +1015,35 @@ def store_matches(
...
-            formal_sum = molecule.total_charge.m
+                charge_sum = sum(molecule_charges)
+                formal_sum = molecule.total_charge.m
 
-            if abs(charge_sum - formal_sum) > 0.01:
-                if allow_nonintegral_charges:
-                    # TODO: Is it worth communicating this as a warning, or would it simply be bloat?
-                    pass
-                else:
+                if abs(charge_sum - formal_sum) > 0.01:
```
Yoshanuikabundi, line 1040: Do we need a defensive check that `formal_sum` is an integer?

mattwthompson: The toolkit seems to provide it as a float, even though it should typically have nothing after the decimal place.


```python
In [1]: from openff.toolkit import Molecule

In [2]: molecule = Molecule.from_smiles("CCO")

In [3]: molecule.total_charge
Out[3]: <Quantity(0.0, 'elementary_charge')>

In [4]: molecule.total_charge.m
Out[4]: 0.0
```

I suppose we could add a check that it's int or int-like prior to this check, but I think that we can trust the toolkit here[^1]. If we got a report that the total charge on a molecule was reported as 0.5, I'd call that a bug in the toolkit.

[^1]: It's not strictly documented as the sum _of formal charges_ but it is [implemented as such](https://github.com/openforcefield/openff-toolkit/blob/5b4941c791cd49afbbdce040cefeb23da298ada2/openff/toolkit/topology/molecule.py#L3479-L3482)

mattwthompson:
Thanks @Yoshanuikabundi!

PR #1329, opened 2025-09-17
Exclude data files from conda/Python packages

mattwthompson:
### Description

This is more or less copied from https://github.com/openforcefield/openff-nagl/pull/216

Something like 80% of the size of this package (when ... packaged) is data files that are only needed for testing:

image

### DIFF
--- a/.github/workflows/examples.yaml
+++ b/.github/workflows/examples.yaml
@@ -57,7 +57,7 @@ jobs:
 
     - name: Run example notebooks
       run: |
-        python -m pytest -c pyproject.toml --nbval-lax --dist loadscope -n logical --durations=20 examples/ \
+        python -m pytest -x -c pyproject.toml --nbval-lax --dist loadscope -n logical --durations=20 examples/ \
           --ignore=examples/deprecated/ \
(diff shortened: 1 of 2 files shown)


mattwthompson:
Aiming to fix upstream CI #1331

mattwthompson:
I have no idea why the examples are failing in CI

.github/workflows/examples.yaml
```diff
@@ -57,7 +57,7 @@ jobs:
 
     - name: Run example notebooks
       run: |
-        python -m pytest -c pyproject.toml --nbval-lax --dist loadscope -n logical --durations=20 examples/ \
+        python -m pytest -x -c pyproject.toml --nbval-lax --dist loadscope -n logical --durations=20 examples/ \
```
mattwthompson: This was originally included for debugging but I think it's worth leaving in

Upsides:
* When there is a critical failure, like high-level imports breaking, we get ~100 lines back instead of ~100 bazillion
* Quicker failure reporting when something goes wrong

Downsides:
* Prone to flakiness
* Misses other failures if a change causes multiple

PR #1332, opened 2025-09-18
Update Packmol defaults

### LINKED ISSUE #1275: Update Packmol defauls
For historical reasons, the Packmol wrappers by default trim the box and then expand it by 10%, which is a bit funky. Before making these functions part of the public API (#1170), I would like to propose reverting both details of this behavior such that the `box_vectors` argument is used for the box to pack molecules into and the requested density `target_density` is _not_ modified.

* For relatively simple use cases, the packing result will be strictly different (different box and positions) but functionally similar (should equilibrate to the same similar thermodynamic state)
* These functions are currently in the private API and have no stability guarantees
* Currently, [there is no guarantee](https://github.com/openforcefield/openff-interchange/blob/312197cdacbb48c8177e3a2e55575a8c858a091f/openff/interchange/components/_packmol.py#L635-L645) that the returned box is at the requested density
* The existing behavior could be maintained by changing argument values, passing a different-sized box and/or a lower target density
* Users are expected to be responsible for equilibrating their systems before production runs

Josh goes into more detail [here](https://github.com/openforcefield/openff-interchange/pull/1252#discussion_r2200055900); this issue is mostly paraphrasing their suggestion

Pinging some stakeholders who would be affected - @IAlibay @atravitz @lilyminium @j-wags @hannaomi @orionarcher (or another Atomate2 contact, maybe @JaGeo?) @JMorado @JHoeflich1 @jaketanderson

I am strongly in favor of this change since prior to locking in a 1.0 API, but am open to reasons why this behavior would be worse and not worth the change.

mattwthompson:
### Description

Closes #1275

### Checklist

- [ ] Add tests
- [x] Lint
- [x] Update docstrings

### DIFF
--- a/openff/interchange/components/_packmol.py
+++ b/openff/interchange/components/_packmol.py
@@ -372,9 +372,6 @@ def _scale_box(box: NDArray, volume: Quantity, box_scaleup_factor: float=1.0) ->
         3x3 matrix in angstroms.
 
     """
-    if box_scaleup_factor == 1.0:
-        return box
-
     final_volume = volume.m_as("angstrom ** 3")
 
     initial_volume = numpy.abs(numpy.linalg.det(box))
--- a/openff/interchange/components/_packmol.py
+++ b/openff/interchange/components/_packmol.py
@@ -348,7 +348,7 @@ def _box_from_density(
     return _scale_box(box_shape, volume)
 
 
-def _scale_box(box: NDArray, volume: Quantity, box_scaleup_factor=1.1) -> Quantity:
+def _scale_box(box: NDArray, volume: Quantity, box_scaleup_factor: float=1.0) -> Quantity:
     """
     Scale the parallelepiped spanned by ``box`` to the given volume.
 
@@ -372,6 +372,9 @@ def _scale_box(box: NDArray, volume: Quantity, box_scaleup_factor=1.1) -> Quanti
         3x3 matrix in angstroms.
 
     """
+    if box_scaleup_factor == 1.0:
+        return box
+
     final_volume = volume.m_as("angstrom ** 3")
 
     initial_volume = numpy.abs(numpy.linalg.det(box))
@@ -462,8 +465,7 @@ def _build_input_file(
         The path to the structure to solvate.
     box_size
         The lengths of each side of the box we want to fill. This is the box
-        size of the rectangular brick representation of the simulation box; the
-        packmol box will be shrunk by the tolerance.
+        size of the rectangular brick representation of the simulation box
     tolerance
         The packmol convergence tolerance.
 
@@ -475,7 +477,7 @@ def _build_input_file(
         The path to the output file.
 
     """
-    box_size = (box_size - tolerance).m_as("angstrom")
+    box_size = box_size.m_as("angstrom")
     tolerance = tolerance.m_as("angstrom")
 
     # Add the global header options.
@@ -633,14 +635,10 @@ def pack_box(
 
     Notes
     -----
-    Returned topologies may have smaller or larger box vectors than what would be defined by the
-    target density if the box vectors are determined by `target_density`. When calling Packmol, each
-    linear dimension of the box is scaled up by 10%.  However, Packmol by default adds a small
-    buffer (defined by the `tolerance` argument which otherwise defines the minimum distance,
-    default 2 Angstrom) at the end of the packed box, which causes small voids when tiling copies of
-    each periodic image. This void is removed in hopes of faster equilibration times but the box
-    density is slightly increased as a result. These changes may cancel each other out or result in
-    larger or smaller densities than the target density, depending on argument values.
+    Packmol places molecules at random positions in the box, satisfying
+    geometric constraints but no thermodynamic considerations. The resulting
+    configurations need to be subject to energy minimization and equilibration
+    routines before being used in production simulations.
 
     """
     # Make sure packmol can be found.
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -33,6 +33,9 @@ jobs:
           - false
         rdkit:
           - true
+        ambertools:
+          - true
+          - false
```
Yoshanuikabundi: 🍾

.github/workflows/ci.yaml
```diff
@@ -41,6 +44,8 @@ jobs:
         exclude:
           - python-version: "3.13"
             openeye: true
+          - ambertools: false
+            openeye: false
```
Yoshanuikabundi: Baby steps I guess

openff/interchange/components/_packmol.py
```diff
@@ -475,7 +479,34 @@ def _build_input_file(
...
+        except subprocess.CalledProcessError as expected_error:
+            # STDOUT from Packmol is a lot of text, but in this middle is something like
+            # ... Version 21.1.1 ...
+            # and they tend to follow something semver-like
+            found_version = Version(expected_error.stdout.decode("utf-8").split("Version ")[1].split(" ")[0])
+
+        except Exception as error:
+            raise PACKMOLRuntimeError(f"Unexpected error ({type(error)}) while running Packmol.") from error
+
+    PACKMOL_USE_PBC = found_version >= Version("20.15.0")
```
Yoshanuikabundi: BLOCKING: We also need a check that the original box (not the box currently passed to this function) is orthorhombic. That'll require some code reorganization, or at least an `apply_orthorhombic_pbcs` argument to `_build_input_file` It doesn't make a whole lot of sense to apply orthorhombic PBCs to the brick representation of a triclinic box! Packmol seems to currently only support orthorhombic PBCs.

openff/interchange/components/_packmol.py
```diff
@@ -633,15 +667,16 @@ def pack_box(
...
+    Packmol places molecules at random positions in the box, satisfying
+    geometric constraints but no thermodynamic considerations. The resulting
+    configurations need to be subjected to energy minimization and equilibration
+    routines before being used in production simulations.
+
+    Periodic boundary conditions are accounted for during packing if using Packmol version 20.15.0 or newer.
 
+    If Packmol version 20.15.0 or less is used, periodic boundary conditions may not be handled correctly.
+
+    Periodic boundary conditions are accounted for during packing.
```
Yoshanuikabundi: These lines should be covered by the above suggestion but GitHub wouldn't let me 

```suggestion
```

docs/releasehistory.md
```diff
@@ -13,6 +13,16 @@ Please note that all releases prior to a version 1.0.0 are considered pre-releas
 
 ## 0.4.8 - 2025-10-09
 
+### Behavior changes
+
+* #1332 Introduces the following behavior changes to the private Packmol wrappers:
+  * Packmol version 20.15.0 or newer is recommended.
+  * Periodic boundary conditions are accounted for when placing molecules if Packmol version 20.15.0 or newer is installed; this is the minimum version supporting this feature.
```
Yoshanuikabundi: BLOCKING

```suggestion
  * Periodic boundary conditions are accounted for when placing molecules if the box is orthorhombic and Packmol version 20.15.0 or newer is installed; this is the minimum version supporting this feature.
```

mattwthompson:
Thanks for the feedback - I suspect it's worth using PBCs where possible and would like to move forward with a similar approach that bakes the orthorhombic condition in. But I'd like to demonstrate that this is worth it before making the change, and unfortunately I'm better off making that change along with changing the default values.

openff/interchange/components/_packmol.py
```diff
@@ -475,7 +479,34 @@ def _build_input_file(
...
+            found_version = Version(result.decode("utf-8").split("Version ")[1].split(" ")[0])
+
+        except subprocess.CalledProcessError as expected_error:
+            # STDOUT from Packmol is a lot of text, but in this middle is something like
+            # ... Version 21.1.1 ...
+            # and they tend to follow something semver-like
+            found_version = Version(expected_error.stdout.decode("utf-8").split("Version ")[1].split(" ")[0])
+
+        except Exception as error:
+            raise PACKMOLRuntimeError(f"Unexpected error ({type(error)}) while running Packmol.") from error
```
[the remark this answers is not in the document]
mattwthompson: Yeah, good idea

mattwthompson:
This is doing everything I want it to do and I believe I've incorporated the limitations found in the first review, so I'm re-opening it. The path has been long but the diff is not so huge

Some details I would like the reviewer to keep in mind
* Are there **new** tests to be added? Nothing came up, since the existing tests cover all changed behavior
* Code coverage dropped by a tiny bit only because fewer lines are in the denominator - the number of missed lines (numerator) is unchanged
* Docstrings are not rendered in API documentation, so the formatting may be off but that can be fixed later

openff/interchange/components/_packmol.py
```diff
@@ -1040,8 +1095,20 @@ def solvate_topology_nonwater(
...
+    configurations need to be subjected to energy minimization and equilibration
+    routines before being used in production simulations.
+
+    Orthorhombic periodic boundary conditions are accounted for during packing
+    if using Packmol version 20.15.0 or newer.
+
+    Non-orthorhombic triclinic boxes, and all boxes in Packmol versions older
+    than 20.15.0, are accounted for by packing the
+    `"brick" representation <https://manual.gromacs.org/current/reference-manual/algorithms/periodic-boundary-conditions.html#fig-pbc>`__
+    after shrinking each dimension by the `tolerance`. This introduces a small
```
Yoshanuikabundi:
```suggestion
    after shrinking each dimension by the ``tolerance``. This introduces a small
```

PR #1342, opened 2025-10-01
Fix CI

mattwthompson:
https://github.com/openforcefield/openff-interchange/actions/runs/18151952620


openff/interchange/_tests/unit_tests/interop/lammps/export/test_lammps.py
```diff
@@ -156,7 +156,7 @@ def test_to_lammps_single_mols_relative_to_perturbed_coordinates(
 
         tolerances = {
             "Bond": 1e-11 * kj_mol,
-            "Angle": 1e-12 * kj_mol,
+            "Angle": 1e-10 * kj_mol,
```
mattwthompson, line 159: Not sure what changed but this is still a minuscule tolerance

PR #1352, opened 2025-10-16
Improve unit conversion and OpenMM export runtime

mattwthompson:
There are two changes in this PR, which I can split up if that would make it easier.

1. Use quicker unit conversions when dealing with Pint. There are some minor per-call speedups using `Quantity(1.2, "nanometer")` and `value.m_as("nanometer")` vs. `1.2 * unit.nanometer` and `value.m_as(unit.nanometer)`
2. Cache valence parameter lookups based on `PotentialKey` values when exporting to OpenMM. (This could be done for other engines, too.)

### Description

- [ ] Add tests
- [x] Lint
- [x] Update docstrings

### DIFF
--- a/devtools/conda-envs/test_not_py313.yaml
+++ b/devtools/conda-envs/test_not_py313.yaml
@@ -5,7 +5,7 @@ dependencies:
   # Core
   - python
   - versioningit
-  - numpy =2.2
+  - numpy <2.3
   - pydantic >=2,<2.12
   # OpenFF stack
   - openff-toolkit-base =0.17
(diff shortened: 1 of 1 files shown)


openff/interchange/drivers/gromacs.py
```diff
@@ -218,15 +217,6 @@ def _parse_gmx_energy(edr_path: str) -> dict[str, Quantity]:
...
 
-    #   for key in energies:
-    #       energies[key] *= kj_mol
-
-    #   # TODO: Better way of filling in missing fields
-    #   # GROMACS may not populate all keys
-    #   for required_key in ["Bond", "Angle", "Proper Dih."]:
-    #       if required_key not in energies:
-    #           energies[required_key] = 0.0 * kj_mol
-
```
j-wags: (not blocking) Did you mean to remove this?

j-wags (changes requested):
I think the `__hash__` change requires corresponding changes to `__eq__`, see comment.

j-wags (approved):
Awesome, excited for this and the toolkit isomorphism optimization to hit production!

PR #1364, opened 2025-11-05
Fix some Markdown formatting

### LINKED ISSUE #1363: Section header links broken
**Description**

Please describe what you were trying to do, what you expected to happen, and what happened instead.

This link

Image

is means to go here: https://docs.openforcefield.org/projects/interchange/en/stable/using/intro.html#units-in-interchange

I also think the 🔗 button here:

Image

should end in `interchange_units` but it's not

mattwthompson:
### Description

Closes #1363

This renders fine locally, for what I looked at. When I click "see Units in Interchange" I'm brought to where I expect


mattwthompson:
It looks like that's pulled down because of the Sphinx down-pin: https://myst-parser.readthedocs.io/en/latest/develop/_changelog.html#id2

PR #1376, opened 2025-11-24
Make `Interchange.topology` not store positions

### LINKED ISSUE #1283: Make `Interchange.positions` the sole source of truth for positions/coordinates
i.e. make `Interchange.topology.{get|set|clear}_positions` do nothing

Relates to #1282 

I would like to know what @Yoshanuikabundi thinks about this idea

P.S. maybe a `Interchange.coordinates` alias would be useful?

mattwthompson:
### Description

Resolves #1283 

- [x] Make a new class for `Interchange.topology` which does not store positions
- [x] Add tests
  - [x] Ensure basic behavior mirrors `openff.toolkit.Topology`
  - [x] Test raised errors
  - [x] Other tests pass
- [x] Mypy still passes
- [ ] Add `Interchange.to_topology` which returns `openff.toolkit.Topology` with set positions? Decided against this as it's confusing - a little repetitive and goes upstream of the recommended information flow
- [x] Update release history

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -1,0 +1,10 @@
+MD013: false
+
+MD033: false
+
+MD024: false
+
+MD046: false
+
+MD060:
+  style: "compact"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,7 +12,7 @@ repos:
   hooks:
   - id: add-trailing-comma
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.14.5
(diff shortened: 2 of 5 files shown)


Yoshanuikabundi (approved):
LGTM! This is quite different to my suggestion but it is nice and simple. I've checked that it works through a couple of Interchange creation pathways and everything looks good. Tests look good, and I've looked over the other examples by hand and I think you've covered everything. We might need to still be wary of people getting positions by iterating over the first conformers of each molecule but there's only so much you can do.

Issue #1393, opened 2025-12-16
Add `ElectrostaticsCollection.get_charge_array`


mattwthompson:
**Description**
Right now this is on `SMIRNOFFElectrostaticsCollection` and therefore not accessible as `ElectrostaticsCollection`, which is used in a few corner cases.

PR #1396, opened 2025-12-17
Fix charge ordering in Amber files

### LINKED ISSUE #1395: Possible charge ordering failure in `Interchange.to_prmtop`
**Describe the bug**

Amber files, in some cases, can be mangled due to charges being ordered incorrectly.

**To Reproduce**

(I may replace this with a more targeted reproduction later)

Using the files from [this comment](https://github.com/openforcefield/openff-toolkit/issues/2136#issuecomment-3662349667),

```python
In [1]: from openff.interchange import Interchange

In [2]: interchange = Interchange.model_validate_json(open("../interchange/interchange.json").read())

In [3]: [*interchange['Electrostatics'].key_map.items()][196:200]  # 196 atoms in the protein, then waters
Out[3]:
[(LibraryChargeTopologyKey(this_atom_index=196),
  PotentialKey associated with handler 'LibraryCharges' with id '[#1]-[#8X2H2+0:1]-[#1]', mult 0),
 (LibraryChargeTopologyKey(this_atom_index=199),
  PotentialKey associated with handler 'LibraryCharges' with id '[#1]-[#8X2H2+0:1]-[#1]', mult 0),
 (LibraryChargeTopologyKey(this_atom_index=202),
  PotentialKey associated with handler 'LibraryCharges' with id '[#1]-[#8X2H2+0:1]-[#1]', mult 0),
 (LibraryChargeTopologyKey(this_atom_index=205),
  PotentialKey associated with handler 'LibraryCharges' with id '[#1]-[#8X2H2+0:1]-[#1]', mult 0)]

In [4]: [*interchange['Electrostatics'].key_map.items()][-13:-10]  # 10 or so ions, with last water atoms before
Out[4]:
[(LibraryChargeTopologyKey(this_atom_index=6204),
  PotentialKey associated with handler 'LibraryCharges' with id '[#1:1]-[#8X2H2+0]-[#1]', mult 0),
 (LibraryChargeTopologyKey(this_atom_index=6207),
  PotentialKey associated with handler 'LibraryCharges' with id '[#1:1]-[#8X2H2+0]-[#1]', mult 0),
 (LibraryChargeTopologyKey(this_atom_index=6210),
  PotentialKey associated with handler 'LibraryCharges' with id '[#1:1]-[#8X2H2+0]-[#1]', mult 0)]

In [5]: interchange.to_amber('x')
/Users/mattthompson/mamba/envs/toolkit-2136/lib/python3.13/site-packages/openff/interchange/components/mdconfig.py:503: UserWarning: Ambiguous failure while processing constraints. Constraining h-bonds as a stopgap.
  warnings.warn(
/Users/mattthompson/mamba/envs/toolkit-2136/lib/python3.13/site-packages/openff/interchange/components/mdconfig.py:431: SwitchingFunctionNotImplementedWarning: A switching distance 8.0 angstrom was specified by the force field, but Amber does not implement a switching function. Using a hard cut-off instead. Non-bonded interactions will be affected.
  warnings.warn(
```

The issue probably stems from an assumption that ordering of dictionary attributes in `SMIRNOFFElectrostaticsHandler` which doesn't hold. Look closely at the atom indices in the `LibraryChargeTopologyKey`s above, which are not ordered sequentially.

**Output**

Also at the `%FLAG CHARGE` section in `x.prmtop` to see that the water atoms are out of order. Every oxygen is listed first, then every hydrogen. (It should be O H H O H H ..., matching the topology.) The other sections in Amber files (`%FLAG ATOM_NAME`, `%FLAG ATOMIC_NUMBER`) are ordered correctly as are the atoms in the OpenMM system.

**Additional context**

I'm confused how this could come to be ... a strong set of test cases would involve shuffling the order of these dictionaries.

mattwthompson:
### Description

Resolves #1395

### Checklist

- [x] Add tests
- [x] Lint
- [x] Update docstrings

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,7 +12,7 @@ repos:
   hooks:
   - id: add-trailing-comma
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.14.8
+  rev: v0.14.9
   hooks:
     - id: ruff-check
       args: ["--fix"]
@@ -23,7 +23,7 @@ repos:
   - id: blacken-docs
     files: ^docs/
 - repo: https://github.com/igorshubovych/markdownlint-cli
-  rev: v0.46.0
+  rev: v0.47.0
   hooks:
   - id: markdownlint
     exclude: .github


Yoshanuikabundi (approved):
Looks good! I've checked that:

- [x] New test passes
- [x] New test fails at first assert with old `_export.py`
- [x] New test fails at second assert with old `_export.py` and first assert commented out
- [x] New test fails at third assert with old `_export.py` and first two asserts commented out
- [x] New test fails at energy comparison with old `_export.py` and all three asserts commented out

So the new test certainly captures the issue comprehensively, and the fix in `_export.py` certainly fixes it. The fix and test also look correct on inspection, as the `get_charge_array()` method is certainly guaranteed to be in order, and any unintended changes in system energies should be caught by other tests. `get_charge_array` raises when virtual sites are present, but the docs say that we don't support virtual site export to Amber so this is not an issue. I agree that #1393 would be an improvement, but as you say this is an urgent bug fix.

Also congrats on the super fast turnaround, I can imagine this one being a pain to diagnose!

PR #1408, opened 2026-01-07
Make 0.5.0 release notes

### LINKED ISSUE #1384: Refresh README
It's been a few years and I think a refresh, if not a full re-write, is justified. Josh has one suggestion below

---

Pretty much irrelevant here but can't help but think that "parametrized chemical system" might be more precise than "parametrized data"?

_Originally posted by @Yoshanuikabundi in https://github.com/openforcefield/openff-interchange/pull/1371#discussion_r2583876083_

### LINKED ISSUE #1402: New interchange `target_density` scale factor behaviour does not have a direct 1.1 relationship with previous behaviour
**Description**

The current news item for the scale factor change states that "previous behavior can be restored by passing scaling the `target_density` argument down by a factor of 1.1" (note there's a typo in the news item!).

However, this is incorrect - it's actually closer to 1.33. I've not been able to nail down the exact number because of the `** (1/3)`, but 1.33 is close enough. (Edit: actually, I think it is just 1.1**3 - someone double check my maths though)

**Reproduction**

Assume a `total_mass` of 100, and an `initial_volume` of 50.

#### Scale of 1.1 and target density fo 0.95

`volume` (as passed to `_scale_box`) = 100 / 0.95 = 105.2632
`volume_scale_factor` = 2.1053
`linear_scale_factor` = 1.4089

#### Scale of 1.0 and target density of 0.8636 (~1.1 scaled down)

`volume` (as passed to `_scale_box`) = 100 / 0.8636 = 115.7943
`volume_scale_factor` = 2.3159
`linear_scale_factor` = 1.3230

#### Scale of 1.0 and target density of 0.715 (~1.33 scaled down)

`volume` (as passed to `_scale_box`) = 100 / 0.715 = 139.8601
`volume_scale_factor` = 2.7972
`linear_scale_factor` = 1.4089

**Software versions**

Current develop version of interchange.

mattwthompson:
### Description

Tag #1257
Closes #1402
Closes #1384

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,30 +6,29 @@
 | **Latest release** | ![GitHub release (latest by date including pre-releases)](https://img.shields.io/github/v/release/openforcefield/openff-interchange?include_prereleases) | |
 | **User support** | [![Documentation Status](https://readthedocs.org/projects/openff-interchange/badge/?version=latest)](https://openff-interchange.readthedocs.io/en/latest/?badge=latest) | [![Discussions](https://img.shields.io/badge/Discussions-GitHub-blue?logo=github)](https://github.com/openforcefield/discussions/discussions) |
 
-A project (and object) for storing, manipulating, and converting molecular mechanics data.
+A project (and object) for storing, manipulating, and converting parametrized chemical systems.
 
 ## Documentation
 
-Documentation for Interchange, including examples, a brief user guide, release history, and API docs, is available [on the OpenFF website](https://docs.openforcefield.org/projects/interchange/en/stable). Example notebooks are rendered online among examples from other projects in the [OpenFF ecosysytem docs](https://docs.openforcefield.org/en/latest/examples.html)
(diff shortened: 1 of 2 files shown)


Yoshanuikabundi (approved):
Perfect! We previously scaled the entire box matrix by 1.1. This had the effect of scaling each dimension of the box by 1.1, so a total scaling of 1.1\*\*3 = 1.331. I can also see this by seeing that it scales each row of the matrix by 1.1, which scales the determinant of the box matrix by 1.1\*\*3 = 1.331, and the determinant is the volume. Plus, ~1.33 is what Irfan came up with numerically, which is very consistent with 1.331. So I feel confident that this is correct.

Other changes also look like improvements to me!

mattwthompson:
Thanks @Yoshanuikabundi!

PR #1431, opened 2026-02-17
Create OpenMM particles even with no non-bonded handlers

### LINKED ISSUE #1429: Could OpenMM systems created without a NonbondedForce still add particles?
**Description**

Related to [this issue and comment in Evaluator](https://github.com/openforcefield/openff-evaluator/issues/727#issuecomment-3900401601), if an OpenMM system is created without a NonbondedForce, no particles are added. I think that's because it happens inside the nonbondedforce creation part of the conversion:

https://github.com/openforcefield/openff-interchange/blob/7a0b56394aa4070031365ace6fa20a4aaf4b6f63/openff/interchange/interop/openmm/_nonbonded.py#L176-L181

However, as shown above, this part seems to rely only on the input topologies rather than require NonbondedForce quantities. (It does then check the system index against a particle map, which I'm much hazier about.) Would it be possible to move this outside the nonbondedforce handling and just always add particles?

mattwthompson:
### Description

Create an `openmm.System` from a force field with no non-bonded handlers and check that it contains particles

This ended up being simpler to implement than I originally thought - 

Subtle behavior(s) resulting here:
- [x] If no non-bonded handlers are present, no `openmm.NonbondedForce`s are created (even though they would be empty)
- [x] If a virtual site handler is present and no vdW handler (or plugin) is present, error

Adjacent behaviors (still) unimplemented (are any relevant now or soon?):
- [ ] If a virtual site handler is present without electrostatics handler(s), unspecified/unhandled behavior
- [ ] If no vdW handler (or plugin) is present but an electrostatics handler(s) is present, don't know what would happen
- [ ] Any Electrostatics "plugins"

Closes #1429

91c55819b2edcb6d1d7a868fab4e00893f9efc37 is a bit of a diversion and I will split it off if this grows in scope any more

### Checklist

- [x] Add tests
  - [x] [Reproduce original issue](https://github.com/openforcefield/openff-interchange/actions/runs/22104013351/job/63881357796?pr=1431)
- [x] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/interop/openmm/_nonbonded.py
+++ b/openff/interchange/interop/openmm/_nonbonded.py
@@ -67,19 +67,7 @@ def _process_nonbonded_forces(
     `combine_nonbondoed_forces=False`.
 
     """
-    from openff.interchange.common._nonbonded import _NonbondedCollection
-
-    for collection in interchange.collections.values():
-        if isinstance(collection, _NonbondedCollection):
-            break
-    else:
-        # If there are no non-bonded collections, assume here that there can be no virtual sites,
-        # so just return an i-i mapping between OpenFF and OpenMM indices
-        return {i: i for i in range(interchange.topology.n_atoms)}
-
-    has_virtual_sites = "VirtualSites" in interchange.collections
-
-    if has_virtual_sites:
+    if "VirtualSites" in interchange.collections:
         from openff.interchange.interop._virtual_sites import (
             _virtual_site_parent_molecule_mapping,
         )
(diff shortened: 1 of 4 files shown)


openff/interchange/interop/openmm/_nonbonded.py
```diff
@@ -126,26 +114,19 @@ def _process_nonbonded_forces(
...
+            "important to you, please raise an issue describing the functionality you wish to "
+            "see.",
+        )
 
     _data = _prepare_input_data(interchange)
 
-    if combine_nonbonded_forces:
+    if _data.vdw_collection is None and _data.electrostatics_collection is None:
+        # Will not make any non-bonded forces, but we have made particles, so just return early
+        return openff_openmm_particle_map
```
mattwthompson: It's not at all obvious in the diff, but returning here instead of at the top of the function results in `openff_openmm_particle_map = _add_particles_to_system(...)` being called when it wasn't before

openff/interchange/interop/openmm/_nonbonded.py
```diff
@@ -67,19 +67,7 @@ def _process_nonbonded_forces(
     `combine_nonbondoed_forces=False`.
```
lilyminium:
```suggestion
    `combine_nonbonded_forces=False`.
```

lilyminium (approved):
LGTM, thanks Matt! Took me a second to get my head around the diff, thanks for the explanatory comment.

I did find a typo that's not actually in the diff here, so feel free to ignore.

PR #1437, opened 2026-02-22
Use custom 1-4 scaling factors on interactions involving virtual sites

### LINKED ISSUE #1436: Non-default 1-4 scaling is not reflected in virtual site exceptions
**Description**

After finding some discrepancies in openmmforcefields' test suite between OpenFF's and OpenMM's handling of virtual sites, I think I found a case where Interchange doesn't apply scaling to nonbonded exceptions involving virtual sites correctly.

The test molecule here is H-O-O-H, with virtual sites attached to the H atoms.  By default, a 1-4 scaling of 0.833333(3333) is being applied.  When I change scale14 to another value, like 0.777777 below, the correct scaling gets applied between the H atoms, but the site-site and site-H scalings still use the default value.  (I'm updating the object in the script but this also happens when I specify it in the XML.)

**Reproduction**

```python

offxml = """<SMIRNOFF version="0.3" aromaticity_model="OEAroModel_MDL">
    <VirtualSites version="0.3">
        <VirtualSite type="BondCharge" match="all_permutations" smirks="[H:1]-[O:2]" distance="1 * angstrom" charge_increment1="-1 * elementary_charge" />
    </VirtualSites>
    <LibraryCharges version="0.3">
        <LibraryCharge smirks="[#1:1]" charge1="2 * elementary_charge" />
        <LibraryCharge smirks="[#8:1]" charge1="-2 * elementary_charge" />
    </LibraryCharges>
    <Electrostatics version="0.3" />
    <vdW version="0.3">
        <Atom smirks="[*:1]" sigma="0 * angstrom" epsilon="0 * kilojoule_per_mole" />
    </vdW>
</SMIRNOFF>"""

molecule = Molecule.from_mapped_smiles("[H:1][O:2][O:3][H:4]")
forcefield = ForceField(offxml)

def dump_exceptions():
    print(f"scale14 in forcefield: {forcefield["Electrostatics"].scale14}")
    system = forcefield.create_interchange(molecule.to_topology()).to_openmm_system()
    nonbonded, = (force for force in system.getForces() if isinstance(force, NonbondedForce))
    for index in range(nonbonded.getNumExceptions()):
        i, j, qq = nonbonded.getExceptionParameters(index)[:3]
        print(f"  {i}-{j}: {qq}")

dump_exceptions()
forcefield["Electrostatics"].scale14 = 0.7777777
dump_exceptions()
```

**Output**

```
scale14 in forcefield: 0.833333
  0-1: 0.0 e**2
  0-2: 0.0 e**2
  1-2: 0.0 e**2
  0-3: 0.833333 e**2
  1-3: 0.0 e**2
  2-3: 0.0 e**2
  4-1: 0.0 e**2
  4-2: 0.0 e**2
  4-3: 0.8333333333 e**2
  5-0: 0.8333333333 e**2
  4-5: 0.8333333333 e**2
  5-1: 0.0 e**2
  5-2: 0.0 e**2
  0-4: 0.0 e**2
  3-5: 0.0 e**2
scale14 in forcefield: 0.7777777
  0-1: 0.0 e**2
  0-2: 0.0 e**2
  1-2: 0.0 e**2
  0-3: 0.7777777 e**2
  1-3: 0.0 e**2
  2-3: 0.0 e**2
  4-1: 0.0 e**2
  4-2: 0.0 e**2
  4-3: 0.8333333333 e**2
  5-0: 0.8333333333 e**2
  4-5: 0.8333333333 e**2
  5-1: 0.0 e**2
  5-2: 0.0 e**2
  0-4: 0.0 e**2
  3-5: 0.0 e**2
```

Given the unit charges, I expect to see `0.7777777 e**2` for all exceptions.  It does seem though like the default that is getting output instead is coming from:

https://github.com/openforcefield/openff-interchange/blob/8e7da426fe891027532931f06ebef8bdf52251e6/openff/interchange/interop/openmm/_nonbonded.py#L1103-L1110

(note the distinct rounding from the non-vsite exception) and suspiciously, when these methods are called elsewhere in this file, the `charge_scaling` and `vdw_scaling` are never set to the values of `coul_14` and `vdw_14` loaded:

https://github.com/openforcefield/openff-interchange/blob/8e7da426fe891027532931f06ebef8bdf52251e6/openff/interchange/interop/openmm/_nonbonded.py#L479-L532

But this is just a hunch on my part as to where this problem lies; apologies if I'm simply doing something wrong with OpenFF here.

**Software versions**

macOS 15.7.4, conda 25.11.1, Python 3.13.12, openff-interchange 0.5.0, environment [(conda_list.txt)](https://github.com/user-attachments/files/25453979/conda_list.txt) created with  `conda create -n off openff-toolkit`

mattwthompson:
### Description

Closes #1436

The short of it is that custom 1-4 scaling factors (i.e. not the default 5/6 and 1/2) were applied to exceptions involving pairs of atoms, but **_not_** to any interactions involving virtual sites.

A small assumption I'm making is that the same 1-4 factor should be applied to all particle pairs (atom-atom, virtual site-virtual site, atom-virtual site. I don't see anything in [the spec](https://openforcefield.github.io/standards/standards/smirnoff/#virtualsites-virtual-sites-for-off-atom-charges) that makes me second-guess myself:

> virtual site particles should exclude non-bonded interactions with, or scale their interactions with, the same particles that the main 'parent atom' that they are attached to does. 

### Checklist

- [x] Ensure virtual site-atom exclusions get get custom 1-4 scaling factors
- [x] Ensure virtual site-virtual site exclusions get custom 1-4 scaling factors
- [x] Add tests
  - [x] Set 1-4 charge scaling factor to non-standard value, ensure virtual site-virtual site exclusions and virtual sites-atom exclusions use this value when inherited from 1-4 parents 
  - [x] Do the same for 1-4 vdW scaling factor (on virtual sites with vdW interactions)
- [x] Lint
- [ ] Update docstrings

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,6 +20,7 @@ jobs:
     name: Test on ${{ matrix.os }}, ${{ matrix.environment }}
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         os:
           - ubuntu-latest
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,7 +20,6 @@ jobs:
     name: Test on ${{ matrix.os }}, ${{ matrix.environment }}
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: false
       matrix:
         os:
           - ubuntu-latest
--- a/openff/interchange/_tests/unit_tests/interop/openmm/test_virtual_sites.py
+++ b/openff/interchange/_tests/unit_tests/interop/openmm/test_virtual_sites.py
@@ -1,3 +1,5 @@
+import random
+
 import numpy
 import pytest
 from openff.toolkit import ForceField, Molecule, Quantity, Topology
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -20,6 +20,7 @@ jobs:
     name: Test on ${{ matrix.os }}, ${{ matrix.environment }}
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
```
mattwthompson: I aim to revert this on merge - this was just to make the CI failure from the reproducing test less ambiguous

openff/interchange/_tests/unit_tests/interop/openmm/test_virtual_sites.py
```diff
@@ -669,3 +671,107 @@ def test_oxygen_interactions_shifted_to_virtual_site(
...
+        # Graph of molecule after parametrization (using zero-indexed OpenMM particle indices):
+        #
+        # H(0) - O(1) - O(2) - H(3)
+        # |                    |
+        # VS(4)                VS(5)
+        #
+        # Using exclusion rule "parents," H-H are a 1-4 pair and so VS-VS are also a 1-4 pair
+        # also H(0)-VS(5) and H(3)-VS(4) are 1-4 pairs, everything else is 1-2, 1-3, so non-interacting
+
+        scale_14 = random.uniform(0.5, 1.0)
```
j-wags: (not blocking) I don't think the benefits of using a random value outweigh the costs (irreproducibility), and would suggest using a fixed list of values if broad coverage is desired.

j-wags (approved):
Looks good. Thanks for the quick turnaround!

PR #1454, opened 2026-03-02
Fix 1453

### LINKED ISSUE #1453: `_clear_caches` can raise an error depending on state of python session
**Description**

When a dead weakref.proxy exists in gc.get_objects(), calling Interchange.from_smirnoff() raises:

    ReferenceError: weakly-referenced object no longer exists

in _clear_caches() at interchange/components/interchange.py:1118.

See: https://github.com/OpenFreeEnergy/ExampleNotebooks/actions/runs/22472512331/job/65092363261

**To Reproduce**

```python

import gc
import weakref

class _Dummy:
    pass

# Create a dead weakref proxy that remains tracked by gc.
# This simulates what happens in long-running processes where libraries
# (OpenMM, gufe, openfe, etc.) create and abandon weakref proxies.
obj = _Dummy()
proxy = weakref.proxy(obj)
container = [proxy]  # prevent the proxy itself from being collected
del obj
gc.collect()

# Trigger the bug via Interchange's public API
from openff.toolkit import ForceField, Molecule

mol = Molecule.from_smiles("C")
mol.generate_conformers(n_conformers=1)
ff = ForceField("openff-2.2.0.offxml")

interchange = Interchange.from_smirnoff(force_field=ff, topology=[mol])
print("Success — no ReferenceError raised")
```

**Output**

```pytb
Traceback (most recent call last):
  File "/Users/jeffreywagner/projects/OpenForceField/ExampleNotebooks/showcase/reproduce_clear_caches_bug.py", line 38, in <module>
    interchange = Interchange.from_smirnoff(force_field=ff, topology=[mol])
  File "/Users/jeffreywagner/projects/OpenForceField/openff-interchange/openff/interchange/components/interchange.py", line 161, in from_smirnoff
    _clear_caches()
    ~~~~~~~~~~~~~^^
  File "/Users/jeffreywagner/projects/OpenForceField/openff-interchange/openff/interchange/components/interchange.py", line 1118, in _clear_caches
    if isinstance(obj, functools._lru_cache_wrapper) and obj.__module__.startswith("openff.interchange"):
       ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ReferenceError: weakly-referenced object no longer exists
```

**Software versions**

- Interchange 0.5.1

j-wags:
- [x] Fix #1453
- [x] Add tests
- [x] Lint

### DIFF
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -15,6 +15,7 @@ Please note that all releases prior to a version 1.0.0 are considered pre-releas
 
 ### Bug fixes
 
+* #1454 Fixes a bug that causes cache operations to error under some circumstances.
 * #1435 Fixes a bug in which `Colletion.set_force_field_parameters` did not update force field parameters upon export to simulation engines.
 
 ## 0.5.1 - 2025-02-25
--- a/openff/interchange/_tests/test_issues.py
+++ b/openff/interchange/_tests/test_issues.py
@@ -302,7 +302,7 @@ def test_issue_1234_openmm(sage, valence_handler):
 def test_clear_caches_with_dead_weakref_proxy():
     """Reproduce a bug where _clear_caches raises ReferenceError when dead
     weakref proxies exist in gc.get_objects(). See
-    https://github.com/OpenFreeEnergy/ExampleNotebooks/actions/runs/22472512331/job/65092363261
+    https://github.com/openforcefield/openff-interchange/issues/1453
     """
     import gc
     import weakref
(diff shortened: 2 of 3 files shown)


mattwthompson (approved):
There are some possible cases I don't understand (quickly noted below the line), but this fixes the minimal reproduction and also the [upstream case](https://github.com/mattwthompson/ExampleNotebooks/pull/1#issuecomment-3986860877), so approved.

---

I don't fully understand what's causing the linked issue, so it's possible this doesn't actually clear an Interchange cache if hit. But for now, since this only happens with OpenFF + (somewhat ambiguous external packages) I'm just assuming that only the first conditional will actually trigger the `ReferenceError`. If that isn't the case, an attempt at clearing a cache would be silently skipped, therefore not fixing the issue altogether. I think this unlikely, however, based on the context of the reported issue.

PR #1435, opened 2026-02-20
Clear caches when re-creating from new force fields

mattwthompson:
### Description

Fix #1234, at least the case of the force field itself being modified

### Checklist

- [x] Add tests
- [x] Lint
- [ ] Update docstrings

Issue #1234, opened 2025-06-02
Updated forcefield parameters are being ignored when creating an `Interchange` due to caching


tom-potter-cresset:
**Description**

In openff-interchange 0.4.1+, parameters are cached when creating a new `Interchange`. When the `ForceField` is updated with new parameters, these are ignored when creating another `Interchange` and the cached parameters are used instead. This is causing issues with Forcebalance, which creates many `Interchange` objects with slightly modified parameters. 

It would be helpful if there was a way to turn off parameter caching in openff-interchange so that these issues could be avoided. 

**Reproduction**

Running any forcebalance torsion-drive calculation with a SMIRNOFF forcefield fails because identical potential energies are being calculated for all modified forcefields. A minimal example is:

```
from forcebalance import smirnoffio
from openff.toolkit import ForceField, Molecule
from openff.interchange import Interchange

ff = ForceField('openff_unconstrained-2.2.1.offxml')
mol = Molecule.from_smiles('CC')
interchange = Interchange.from_smirnoff(ff, mol.to_topology())
print(interchange["ProperTorsions"].potentials)

smirnoffio.assign_openff_parameter(ff, 0.25, "ProperTorsions/Proper/k1/[#1:1]-[#6X4:2]-[#6X4:3]-[#1:4]")
interchange = Interchange.from_smirnoff(ff, mol.to_topology())
print(interchange["ProperTorsions"].potentials)
```

**Output**

With interchange v0.4.2, _k_ is not updated:
```
{PotentialKey associated with handler 'ProperTorsions' with id '[#1:1]-[#6X4:2]-[#6X4:3]-[#1:4]', mult 0: Potential(parameters={'k': <Quantity(0.239039384, 'kilocalorie / mole')>, 'periodicity': <Quantity(3, 'dimensionless')>, 'phase': <Quantity(0.0, 'degree')>, 'idivf': <Quantity(1.0, 'dimensionless')>}, map_key=None)}
{PotentialKey associated with handler 'ProperTorsions' with id '[#1:1]-[#6X4:2]-[#6X4:3]-[#1:4]', mult 0: Potential(parameters={'k': <Quantity(0.239039384, 'kilocalorie / mole')>, 'periodicity': <Quantity(3, 'dimensionless')>, 'phase': <Quantity(0.0, 'degree')>, 'idivf': <Quantity(1.0, 'dimensionless')>}, map_key=None)}
```

With interchange v0.4.0, _k_ is updated correctly:
```
[...]

PR #1456, opened 2026-03-02
Clear internal caches when re-creating from same SMIRNOFF objects

mattwthompson:
### Description

Another possible fix for #1453, including the test @j-wags added

A problem with the `_clear_cache` function introduced in #1435 is that it went through [_all objects visible to the garbage collector_](https://docs.python.org/3/library/gc.html#gc.get_objects) (!) and only interacted with objects that are themselves internal to the `openff.interchange` namespace. This is casting a net that is awfully wide and unnecessary since we already know what caches we want to clear. Since the crux of the bad behavior can be traced to two internal functions (see [here](https://github.com/openforcefield/openff-interchange/issues/1234#issuecomment-3948368333) for a good example) we don't need to iterate so widely.

Put another way: the problematic cached functions[^1] are only two in number and inside of Interchange. We can just clear those specific function caches directly when we need to.

### Checklist

- [x] Add tests
- [x] Lint
- [ ] Update docstrings

[^1]: and they do nothing but look up parameters inside of `ParameterHandler`s, see https://github.com/openforcefield/openff-toolkit/issues/2065


openff/interchange/smirnoff/_create.py
```diff
@@ -134,6 +134,24 @@ def _preprocess_preset_charges(
     return molecules_with_preset_charges
 
 
+def _clear_smirnoff_creation_caches():
+    """
+    Due to bugs in the toolkit, these caches can be invalid of a force field is modified and then
```
j-wags: (not blocking)
```suggestion
    Due to bugs in the toolkit, these caches can be invalid if a force field is modified and then
```

j-wags (approved):

PR #1462, opened 2026-03-09
Fix behavior of `charge_from_molecules` being empty

### LINKED ISSUE #1461: Spurious warning if using vsite force field and charge_from_molecules=[]
**Description**

There's a nice warning printed explaining a possibly surprising behavior that can occur when users use both vsites in a force field and specify `charge_from_molecules`. However this warning is spuriously raised if `charge_from_molecules` is an empty list:

```python
mol = Molecule.from_smiles("CCO")
ff = ForceField("openff-2.3.0.offxml", "tip5p.offxml")
ic = Interchange.from_smirnoff(ff, mol.to_topology(), charge_from_molecules=[])
```

> [/Users/jeffreywagner/projects/OpenForceField/openff-interchange/openff/interchange/smirnoff/_create.py:149](http://localhost:8889/lab/tree/openff/interchange/smirnoff/_create.py#line=148): PresetChargesAndVirtualSitesWarning: Preset charges were provided (via `charge_from_molecules`) alongside a force field that includes virtual site parameters. Note that virtual sites will be applied charges from the force field and cannot be given preset charges. Virtual sites may also affect the charges of their orientation atoms, even if those atoms are given preset charges.

`charge_from_molecules` defaults to `None`, and so the check [here](https://github.com/openforcefield/openff-interchange/blob/04a1712d41058c070977a2c1cda301049cb14f87/openff/interchange/smirnoff/_create.py#L150) doesn't include checking for empty lists.

While this is a small nit here, we noticed that this warning always triggers when using the vsite updates to OpenMMForceFields, since that always passes a list for `charge_from_molecules`, even if it's empty. I think it makes more sense to check for the condition in Interchange than to have OMMFFs convert `[]` to `None` (since the Interchange warning is always spurious if the list is empty).

mattwthompson:
### Description

Fixes #1461

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,7 +30,7 @@ jobs:
           - py311openeye
           - py311examples
           - py311etc
-          - py311amber
+          - py312amber
           - py312openeye
           - py312examples
           - py312etc
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,7 +20,7 @@ jobs:
     name: Test on ${{ matrix.os }}, ${{ matrix.environment }}
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
         os:
           - ubuntu-latest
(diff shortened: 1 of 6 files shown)


j-wags (approved):
Looks great, thanks for the quick turnaround @mattwthompson!

PR #1481, opened 2026-04-10
Ensure box vectors are stored with nanometers

### LINKED ISSUE #1480: Box vectors are are not consistently converted after being set
**Description**
The [docs](https://docs.openforcefield.org/projects/interchange/en/latest/using/construction.html#positions-and-velocities) say that box vectors are stored internally with nanometers:

> Positions [...] are represented in nanometers [...] internally and when reported.

but the setter will accept (and not convert) units like Angstroms.

Note that this is just a problem of being stored in nanometers vs. compatible units - this is not a problem at export time

**Reproduction**

When using a setter (with compatible units), the internal state isn't converted to use nanometers:

```python
>>> from openff.toolkit import *
>>> out = ForceField("2.3.0.offxml").create_interchange(Molecule.from_smiles("CCO").to_topology())
>>> out.box = Quantity([1, 2, 3], 'picometer')
>>> out.box
<Quantity([[1. 0. 0.]
 [0. 2. 0.]
 [0. 0. 3.]], 'picometer')>
```

This isn't a huge deal with nanometers vs. Angstroms since we take care to ensure the correct units at export time. But consistent internal state is desirable.

I think this comes from a missed case [here](https://github.com/openforcefield/openff-interchange/blob/edaec3dcaf9c68f36e34cee9c779159470dee203/openff/interchange/_annotations.py#L203-L209):

```python
>>> from openff.interchange._annotations import _duck_to_nanometer
>>> _duck_to_nanometer([1, 2])
<Quantity([1 2], 'nanometer')>
>>> _duck_to_nanometer(Quantity([0.5, 1.5], "angstrom"))
<Quantity([0.5 1.5], 'angstrom')>
```

**Software versions**

- Which operating system and version are you using? macOS 26.3.1
- How did you install Interchange? Pixi, latest development branch

mattwthompson:
### Description

Fixes #1480

### Checklist

- [ ] Add tests
- [ ] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/_annotations.py
+++ b/openff/interchange/_annotations.py
@@ -204,6 +204,8 @@ def _duck_to_nanometer(value: Any):
     """Cast list or ndarray without units to Quantity[ndarray] of nanometer."""
     if isinstance(value, list | numpy.ndarray):
         return Quantity(value, "nanometer")
+    elif isinstance(value, Quantity):
+        return value.to("nanometer")
     else:
         return value
 
--- a/openff/interchange/_tests/test_issues.py
+++ b/openff/interchange/_tests/test_issues.py
@@ -437,3 +437,13 @@ def test_issue_1480(sage, ethanol, attribute):
             )
 
     assert str(getattr(out, attribute).units) == "nanometer"
+
+    if attribute == "box":
+        box = out.box
+
+        assert box[0][0] == box[1][1] == box[2][2]
+
+        assert str(box[0][0].units) == "nanometer"
+        assert box[0][0].m == pytest.approx(3.5)
+        assert str(box[1][0].units) == "nanometer"
+        assert box[1][0].m == pytest.approx(0.0)
--- a/openff/interchange/_tests/test_issues.py
+++ b/openff/interchange/_tests/test_issues.py
@@ -10,7 +10,7 @@ from openff.toolkit import ForceField, Molecule, Quantity, Topology
 from openff.utilities import get_data_file_path, has_executable, skip_if_missing
 
 from openff.interchange import Interchange
-from openff.interchange._tests import MoleculeWithConformer, needs_gmx, shuffle_topology
+from openff.interchange._tests import MoleculeWithConformer, _rng, needs_gmx, shuffle_topology
 from openff.interchange.drivers import get_amber_energies, get_gromacs_energies, get_openmm_energies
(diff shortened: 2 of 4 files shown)


openff/interchange/_tests/test_issues.py
```diff
@@ -418,3 +418,22 @@ def test_issue_1461(sage, opc, ethanol):
...
+
+
+@pytest.mark.parametrize("attribute", ["box", "positions"])
+def test_issue_1480(sage, ethanol, attribute):
+    """Test that positions and box vectors are stored in nm after being set with non-nm length units."""
+    out = sage.create_interchange(ethanol.to_topology())
+
+    match attribute:
+        case "box":
+            out.box = Quantity(numpy.eye(3) * 35, "angstrom")
```
chrisiacovella, line 430: Philosophical question on your approach to testing; do you think it is worth checking the magnitude of one of the components (since it is incredibly trivial), or are you of the mind that such testing is already covered in the pint/openff-units frameworks and thus not necessary?

openff/interchange/_annotations.py
```diff
@@ -204,6 +204,8 @@ def _duck_to_nanometer(value: Any):
     """Cast list or ndarray without units to Quantity[ndarray] of nanometer."""
     if isinstance(value, list | numpy.ndarray):
         return Quantity(value, "nanometer")
```
chrisiacovella, line 206: I just looked at the testing for the box quantity and there does not appear to be a check the case where an np array is passed to the function (only a list is tested in the function below).  Seems like it might be worth it, while you are at it, to explicitly check just passing an np array.

https://github.com/openforcefield/openff-interchange/blob/edaec3dcaf9c68f36e34cee9c779159470dee203/openff/interchange/_tests/unit_tests/test_annotations.py#L40-L51

chrisiacovella (approved):

openff/interchange/_tests/test_issues.py
mattwthompson: Yes to both - likely within the radius of trust of upstream libraries but also worth testing. However, it takes some creativity to think of details like this so I often forget

openff/interchange/_annotations.py
mattwthompson: Nice find, thanks

mattwthompson:
Thanks for the review @chrisiacovella!

PR #1488, opened 2026-04-17
Log charge provenance once per unique molecule

### LINKED ISSUE #1483: Improve charge provenance logging with copies of molecules
Since - and I'm only about 90% sure this is true (@j-wags?) - chemically identical molecules will **always** get the same partial charges applied from the same partial charge assignment method, it's a little redundant to log the per-atom charge assignment provenance for the 2nd and greater copies of a given molecule.

I'm not yet sure if it's worth logging _that_ copies of a molecule got the same charges as the original. The upside is that it's more useful to log this provenance than have the user figure out, for a given molecule, what the corresponding original molecule is. The downside is that it would be extremely redundant for solvent molecules, i.e. 5,000 lines of logging that water molecules got OPC3 charges.

### LINKED ISSUE #1484: Improve charge provenance logging with large molecule(s)
As @IAlibay and others have mentioned, charge provenance logging is more or less useless with large molecules like proteins. The original design logs each atom on a new line, so i.e. a protein given AshGC charges will log several thousand approximately-identical lines.

Irfan also suggests a solution: simply log the entire molecule in one line.

mattwthompson:
Closes #1484 
Closes #1483 

(I found this was easier to make these two changes in one go, but I can refactor it into separate changes if desired)

### Description

With these changes:

* The charge method used to apply partial charges to a given molecule is only logged once
  * Previously this was per-atom, and therefore repetitive. Now it's just once per molecule
* The charge method used to apply partial charges to a duplicates of a unique molecule is only logged once
  * Previously, this was duplicated for each duplicate molecule, so e.g. 1000 waters would be 3000 lines in the log
* InChI is used to identify the moleule
  * SMILES for all molecules won't work, because of proteins
  *  Another solution could be to use SMILES for small molecules like water and ions but InChI for molecules above some threshold size 
  * `Molecule.to_inchi` failures are handled with a wide net
* Logging logic is not entered unless it's turned to `logging.INFO`

Here is an example of it in use, adopting from others' code
```python
In [1]: import logging
   ...: logging.basicConfig(level=logging.INFO)
   ...:
   ...: from openff.toolkit import ForceField, Molecule
   ...: from openff.interchange import Interchange
   ...: from openff.interchange.components._packmol import pack_box
   ...: from openff.units import unit
   ...:
   ...: mols = [Molecule.from_smiles(smiles) for smiles in ['CCO', 'O']]
   ...: for mol in mols:
   ...:     mol.generate_conformers(n_conformers=1)
   ...:
   ...: topology = pack_box(
   ...:     molecules=mols,
   ...:     number_of_copies=[3, 2],
   ...:     target_density=0.1 * unit.gram / unit.milliliter,
   ...: )
   ...:
   ...: ff = ForceField("openff-2.2.1.offxml")
   ...: interchange = Interchange.from_smirnoff(force_field=ff, topology=topology)
WARNING:pint.util:Redefining '[electric_potential]' (<class 'pint.delegates.txt_defparser.plain.DerivedDimensionDefinition'>)
/Users/mattthompson/software/openff-interchange/openff/interchange/components/interchange.py:1119: FutureWarning: `torch.distributed.reduce_op` is deprecated, please use `torch.distributed.ReduceOp` instead
  if isinstance(obj, functools._lru_cache_wrapper) and obj.__module__.startswith("openff.interchange"):
INFO:openff.interchange.smirnoff._nonbonded:Charge section ToolkitAM1BCC, using charge method am1bccelf10, applied to molecule with InChI InChI=1S/C2H6O/c1-2-3/h3H,2H2,1H3
INFO:openff.interchange.smirnoff._nonbonded:Charge section LibraryCharges applied to molecule with InChI InChI=1S/H2O/h1H2
```

I have not done it yet, but, in principle, a large solvated protein-ligand system may have its entire charge provenance logged with as few as 5 lines i the log

### Checklist

- [x] Update tests
- [x] Lint
- [ ] Update docstrings

### DIFF
--- a/openff/interchange/_tests/unit_tests/smirnoff/test_charge_assignment_logging.py
+++ b/openff/interchange/_tests/unit_tests/smirnoff/test_charge_assignment_logging.py
@@ -574,3 +574,13 @@ def test_case14(caplog, sage, ligand):
         # Should use preset charges, not NAGL
         assert info["preset"] == ["InChI=1S/C2H6O/c1-2-3/h3H,2H2,1H3"]
         assert "NAGL" not in info
+
+
+def test_inchi_fallback(caplog, sage):
+    """Test that molecules that fail InChI generation are still logged in some way."""
+
+    # TODO: Might be a toolkit bug that needs to be worked around
+    with caplog.at_level(logging.INFO):
+        sage.create_interchange(
+            Molecule.from_smiles(342 * "C").to_topology(),
+        )
--- a/openff/interchange/_tests/unit_tests/smirnoff/test_charge_assignment_logging.py
+++ b/openff/interchange/_tests/unit_tests/smirnoff/test_charge_assignment_logging.py
@@ -115,7 +115,7 @@ AM1BCC_KEY = "am1bccelf10" if OPENEYE_AVAILABLE else "am1bcc"
 NAGL_KEY = "openff-gnn-am1bcc-1.0.0.pt"
 
 
-def map_methods_to_atom_indices(caplog: pytest.LogCaptureFixture) -> dict[str, list[int]]:
+def map_methods_to_inchi(caplog: pytest.LogCaptureFixture) -> dict[str, list[int]]:
     """
     Map partial charge assignment methods to (sorted) atom indices.
     """
@@ -131,13 +131,21 @@ def map_methods_to_atom_indices(caplog: pytest.LogCaptureFixture) -> dict[str, l
         message = record.msg
 
         if message.startswith("Charge section LibraryCharges"):
-            info["library"].append(int(message.split("atom index ")[-1]))
+            _, molecule = message.split(", ")
+
+            info["library"].append(molecule.split("InChI ")[-1])
 
         elif message.startswith("Charge section ToolkitAM1BCC"):
-            info[message.split(", ")[1].split(" ")[-1]].append(int(message.split("atom index ")[-1]))
+            _, method, molecule = message.split(", ")
+
+            assert method.split()[-1] == AM1BCC_KEY, f"Expected method {AM1BCC_KEY} but got {method.split()[-1]}"
+
+            info[method.split()[-1]].append(molecule.split("InChI ")[-1])
 
         elif message.startswith("Charge section NAGLCharges"):
-            info["NAGLChargesHandler"].append(int(message.split("atom index ")[-1]))
+            _, _, molecule = message.split(", ")
+
+            info["NAGL"].append(molecule.split("InChI ")[-1])
 
         # without also pulling the virtual site - particle mapping (which is different for each engine)
         # it's hard to store more information than the orientation atoms that are affected by each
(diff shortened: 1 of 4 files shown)


openff/interchange/smirnoff/_nonbonded.py
```diff
@@ -1025,6 +988,63 @@ def store_matches(
...
+                "Charge section ChargeIncrementModel, using charge method "
+                f"{key.partial_charge_method}, "
+                f"applied to molecule with InChI {inchi}",
+            )
+
+        elif type(key) is ChargeIncrementTopologyKey:
+            # here is where the individual increments could be logged at creation time, but they're
+            # also logged in _get_charges
+            pass
+
```
mattwthompson, line 1034: This is a subtle behavior change.

Previously, there was logging for the charge method applied to the _whole_ molecule (`partial_charge_method` in the SMIRNOFF data and API) and then also each individual charge increment. Each of these are done on a "per-atom" view, if repetitive, same as how other charge methods were logged before these changes.

Taking a "per-molecule" view works nicely for other methods (no need to log $N_{atoms}$ lines for every atom in molecule that's given AM1-BCC charges) and works fine for `ChargeIncrementModelHandler.partial_charge_method` but doesn't work for each individual charge increment (i.e. BCC). There's probably a way to refactor all of this to cater to that detail, but I don't think it's trivial. Given that `ChargeIncrementModel` is not actually used in production, and there aren't plans to do so, I'm inclined to not put that effort into this change.

openff/interchange/smirnoff/_nonbonded.py
```diff
@@ -952,50 +952,13 @@ def store_matches(
...
-
-                            else:
-                                raise ValueError(f"Unhandled key type {type(new_key)}")
 
                             self.key_map[new_key] = matches[key]
 
+            # don't bother going through logging machinery if it's not opted-in to
+            if logger.isEnabledFor(logging.INFO):
+                self._log_charge_provenance(unique_molecule, new_key)
+
```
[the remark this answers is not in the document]
mattwthompson: This is intentional - the last key in a molecule contains all of the information we need. And if that loop fails, we have bigger problems.

openff/interchange/_tests/unit_tests/smirnoff/test_charge_assignment_logging.py
```diff
@@ -542,8 +569,18 @@ def test_case14(caplog, sage, ligand):
...
+
+
+def test_inchi_fallback(caplog, sage):
+    """Test that molecules that fail InChI generation are still logged in some way."""
+
+    # TODO: Might be a toolkit bug that needs to be worked around
+    with caplog.at_level(logging.INFO):
+        sage.create_interchange(
+            Molecule.from_smiles(342 * "C").to_topology(),
+        )
```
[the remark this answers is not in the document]
mattwthompson: This diff won't work but the idea is valuable

openff/interchange/smirnoff/_nonbonded.py
```diff
@@ -1025,6 +988,64 @@ def store_matches(
                         f"{formal_sum}.",
                     )
 
+    @staticmethod
+    def _log_charge_provenance(
+        unique_molecule: Molecule,
+        key: ChargeModelTopologyKey | SingleAtomChargeTopologyKey | LibraryChargeTopologyKey,
+    ):
+        try:
+            inchi = unique_molecule.to_inchi()
```
j-wags: (blocking) inchi will scale with the size of the molecule, and so it will have the same problems as SMILES for proteins. inchikey is constant length and I think is the better thing to use here. 
```suggestion
            inchikey = unique_molecule.to_inchikey()
```

openff/interchange/smirnoff/_nonbonded.py
```diff
@@ -1025,6 +988,64 @@ def store_matches(
...
+            logger.info(
+                f"Could not generate InChI for molecule with Hill formula {unique_molecule.to_hill_formula()}. "
+                "This molecule's charge provenance will be logged without a valid InChI. "
+                f"Error was {error}",
+            )
+            inchi = "UNKNOWN_INCHI"
+
+        if type(key) is LibraryChargeTopologyKey:
+            logger.info(
+                f"Charge section LibraryCharges, applied to molecule with InChI {inchi}",
```
j-wags: (not blocking) I think it'd be good for each message to also include the Hill formula for the molecule, since users will be able to read what they need from that 90% of the time, whereas inchi/inchikey are harder to interpret quickly.

j-wags (changes requested):

PR #1537, opened 2026-09-15
Assorted docs fixes from Claude

mattwthompson:
### Description

I asked Claude

```
Here are docs for one of my projects: https://docs.openforcefield.org/projects/interchange/en/stable/

The source code is at: https://github.com/openforcefield/openff-interchange

Find any errors in the docs, or suggest improvements therein
```


mattwthompson:
Claude (whatever model I'm using on the cheap) 🤝 Copilot (whatever model I'm using on the cheap

PR #1539, opened 2026-09-19
Look up Buckingham virtual sites by smirks, name, and match

### LINKED ISSUE #1457: Buckingham VSite handler tries to get parameters by SMIRKS
**Description**

Probably not a common scenario, but smee tests have uncovered that the Buckingham VSite handler still tries to get parameters by SMIRKS only, which is no longer supported.

https://github.com/openforcefield/openff-interchange/blob/7efc15d4bc91d8b7def7e0b75dedab0425436d5c/plugins/nonbonded_plugins/virtual_sites.py#L78-L80

**Reproduction**

Please include a [minimally reproducing example](https://stackoverflow.com/help/minimal-reproducible-example) of this bug.

See CI here: https://github.com/openforcefield/smee/actions/runs/22615205684/job/65526240736?pr=140

**Output**

Please include the output, including full tracebacks of any errors, resulting from the reproduction.

**Software versions**

shaneraphel:
`store_potentials` in the Buckingham plugin splits `potential_key.id` and then looks up `parameter_handler.parameters[smirks]`, dropping name and match. `SMIRNOFFVirtualSiteCollection` already uses `_lookup_virtual_site_parameter(smirks, name, match)` because VirtualSiteType SMIRKS are not unique (openff-toolkit #1847).

This uses the same lookup on the Buckingham arm so a name/match pair is not collapsed onto the first matching SMIRKS.

Fixes #1457


mattwthompson:
@shanyu did you author this?

shaneraphel:
This pr is coding by a formal language system and not generation by llm.

shaneraphel:
Yes,i am the author,I will fix it.

